PR #4, opened 2025-01-06
Make standalone framework and add features

davidwalter2:
With this PR the combinetf2 is outscoped from the narf framework (https://github.com/bendavid/narf). Files and directories are renamed and files are split into multiple smaller files to have a clearer separation.

New features:
- Writing out (nuisance) parameter pulls and constraints
- Computing and writing out global and traditional impacts on parameters and groups of parameters
- Implement and writing out global impacts on observables (the non-profile case is wrong)
- Write out additional information such as covariance on observable e.g. for unfolded cross sections
- Factorize functionality to save and write histograms to better toggle to other formats (in the future) using a "workspace" object.
- Implement sparse mode (but very slow, not sure if it can be improved)
- Add option to skip chi2 calculation
- Adding a "TensorWriter" class to provide an interface to construct the input needed for fitting. It is closely inspired by the HDF5Writer from WRemnants (https://github.com/WMass/WRemnants/blob/main/wremnants/HDF5Writer.py).
  - Symmetrization options are also implemented 
  - The functionalities are tested in tests/make_tensor.py which can also be used as a starting point how to use it
- Minimal functions to read the fitresults are added in combinetf2/io_tools.py
- Minimal scripts are added to print parameter pulls & constraints or impacts. 
- Added Pre-commit hooks
- Added github CI workflow


bendavid:
So does this PR fully contain the functionality of https://github.com/bendavid/narf/pull/27 and https://github.com/bendavid/narf/pull/28 ?

scripts/pullsAndImpacts.py
```diff
@@ -0,0 +1,859 @@
...
+            nspace = nval.apply(
+                lambda x, n=nval.apply(len).max(): " " * (n - len(x) + 1)
+            )
+            if include_ref:
+                frmt_ref = (
+                    "{:0"
+                    + str(
+                        int(
+                            np.log10(max(df[f"absimpact_ref"]))
+                            if max(df[f"absimpact_ref"]) > 0
```
kdlong, line 182: I think the thing that is making this fail is a logic error here: in the case where the maximum value is much smaller than 1 but non-zero, the log gives a negative number. So I think it would be good enough to make this > 1, for example, but then I'm not 100% sure I follow the logic of adding 2.

scripts/pullsAndImpacts.py
```diff
@@ -0,0 +1,859 @@
...
+                )
+                + ".2f}"
+            )
+            nval = df["absimpact"].apply(
+                lambda x, frmt=frmt: frmt.format(x)
+            )  # .astype(str)
+            nspace = nval.apply(
+                lambda x, n=nval.apply(len).max(): " " * (n - len(x) + 1)
+            )
+            if include_ref:
```
kdlong, line 176: Actually looking a bit closer, isn't this logic quite odd? if include_ref is nested inside if include_ref?

davidwalter2: I think it's to make the spacing right in case there are single and double digit numbers, which is more relevant in case reference fitresult is given to have the numbers all aligned in columns.

PR #6, opened 2025-02-14
Style formatting of plots

davidwalter2:


README.md
```diff
@@ -0,0 +1,94 @@
+![Framework Logo](data/logo/logo.svg)
+
+Perform complex profile binned maximum likelihood fits by exploiting state-of-the-art differential programming. 
+Computations are based on the tensorflow 2 library and scipy minimizers with multithreading support on CPU (FIXME: and GPU).
+Implemted approximations in the limit of large sample size to simplify intensive computations.
```
bendavid, line 5: typo

README.md
```diff
@@ -0,0 +1,94 @@
...
+First, make a python version, e.g. in the combinetf2 base directory (On some machines you have to use `python3`):
+```bash
+python -m venv env
+```
+The activate it and install the necessary packages
+```bash
+source env/bin/activate
+pip install wums tensorflow numpy h5py hist scipy matplotlib mplhep seaborn pandas plotly kaleido
+```
+The packages `matplotlib`, `mplhep`, `seaborn`, `pandas`, `plotly`, and `kaleido` are only needed for the plotting scripts.
```
bendavid, line 23: can add a ```pip install combinetf2[plotting]``` option to automatically get the right dependencies for this as well?

README.md
```diff
@@ -0,0 +1,94 @@
...
+```bash
+python -m venv env
+```
+The activate it and install the necessary packages
+```bash
+source env/bin/activate
+pip install wums tensorflow numpy h5py hist scipy matplotlib mplhep seaborn pandas plotly kaleido
+```
+The packages `matplotlib`, `mplhep`, `seaborn`, `pandas`, `plotly`, and `kaleido` are only needed for the plotting scripts. 
+In case you want to contribute to the development, please also install the linters `isort`, `flake8`, `autoflake`, `black`, and `pylint` used in the pre-commit hooks and the github CI
```
bendavid, line 24: and maybe for here as well combinetf2[devel] or similar?

README.md
```diff
@@ -0,0 +1,94 @@
...
+```bash
+source setup.sh
+```
+
+## Making the input tensor
+
+An example can be found in ```tests/make_tensor.py -o test_tensor.hdf5```. 
+
+### Symmetrization
+By default, systematic variations are asymmetric.
```
bendavid, line 62: for "safe" usage from wremnants might still prefer to keep "average" as the default (or possibly allow forcing symmetric-only variations and throwing an error if a user tries to add an asymmetric one)

combinetf2/inputdata.py
```diff
@@ -0,0 +1,212 @@
...
+
+            if self.sparse:
+                self.norm_sparse = makesparsetensor(hnorm_sparse)
+                self.logk_sparse = makesparsetensor(hlogk_sparse)
+            else:
+                self.norm = maketensor(hnorm)
+                self.logk = maketensor(hlogk)
+
+            self.normalize = normalize
+            if self.normalize:
```
bendavid, line 180: this would need to be adapted to support the symmetric_tensor case

suggest adding a NotImplementedError for this case for the moment at least

combinetf2/io_tools.py
```diff
@@ -0,0 +1,118 @@
...
+        key = f"{key}_asimov"
+    h5results = ioutils.pickle_load_h5py(h5file[key])
+    if meta:
+        meta = ioutils.pickle_load_h5py(h5file["meta"])
+        return h5results, meta
+    return h5results
+
+
+def get_poi_names(fitresult):
+    if "impacts" in fitresult.keys():
```
bendavid, line 25: this seems like a very indirect way to get the poi names.  Should relevant metadata simply be stored directly in the fitresults?

combinetf2/fitter.py
```diff
@@ -0,0 +1,1410 @@
...
+        self.beta0 = tf.Variable(
+            tf.ones_like(self.indata.data_obs), trainable=False, name="beta0"
+        )
+
+        nexpfullcentral = self.expected_yield()
+        self.nexpnom = tf.Variable(nexpfullcentral, trainable=False, name="nexpnom")
+
+        # parameter covariance matrix
+        self.cov = tf.Variable(self.prefit_covariance(), trainable=False, name="cov")
+        self.hess = tf.zeros(
```
bendavid, line 100: why is this initialized with zeros?

also, looks like only the top-left corner of this is ever explicitly used (and only after the fit)

combinetf2/fitter.py
```diff
@@ -0,0 +1,1410 @@
...
+    def defaultassign(self):
+        self.cov.assign(self.prefit_covariance())
+        self.theta0defaultassign()
+        self.beta0defaultassign()
+        self.xdefaultassign()
+
+    def bayesassign(self):
+        if self.npoi > 0:
+            raise NotImplementedError(
+                "Assignment for Bayesian toys is not currently supported in the presence of explicit POIs"
```
bendavid, line 149: why not?  just set them to the nominal?

combinetf2/fitter.py
```diff
@@ -0,0 +1,1410 @@
...
+
+    @tf.function
+    def impacts_parms(self):
+        # impact for poi at index i in covariance matrix from nuisance with index j is C_ij/sqrt(C_jj) = <deltax deltatheta>/sqrt(<deltatheta^2>)
+        cov_poi = self.cov[: self.npoi]
+        cov_noi = tf.gather(self.cov[self.npoi :], self.indata.noigroupidxs)
+        v = tf.concat([cov_poi, cov_noi], axis=0)
+        impacts = v / tf.reshape(tf.sqrt(tf.linalg.diag_part(self.cov)), [1, -1])
+
+        # impact data stat # FIXME: This might not be correct in case noi != systsnoconstraint
```
bendavid, line 224: I don't think the FIXME comment is relevant, noi's are irrelevant for this purpose

combinetf2/fitter.py
```diff
@@ -0,0 +1,1410 @@
...
+        inv_hess_stat = tf.linalg.solve(hess_stat, identity)  # Solves H * X = I
+
+        if self.binByBinStat:
+            # impact bin-by-bin stat
+            val_no_bbb, grad_no_bbb, hess_no_bbb = self.loss_val_grad_hess(
+                profile_grad=False
+            )
+
+            hess_stat_no_bbb = hess_no_bbb[:nstat, :nstat]
+            inv_hess_stat_no_bbb = tf.linalg.solve(hess_stat_no_bbb, identity)
```
bendavid, line 237: same as above

combinetf2/fitter.py
```diff
@@ -0,0 +1,1410 @@
...
+        expcov_noBBB = expcov
+        if self.binByBinStat:
+            varbeta0 = tf.math.reciprocal(self.indata.kstat)
+            exp_cov_BBB = pdexpdbeta0 @ (varbeta0[:, None] * tf.transpose(pdexpdbeta0))
+            expcov += exp_cov_BBB
+
+        if compute_global_impacts:
+            # FIXME This is not correct
+            print(
+                "WARNING: Global impacts on observables without profiling is under development and probably wrong!"
```
bendavid, line 488: better to error out here for now?

combinetf2/fitter.py
```diff
@@ -0,0 +1,1410 @@
...
+        dxdtheta0_squared = tf.square(dxdtheta0)
+
+        # global impact data stat
+        dxdnobs_poi = dxdnobs[: self.npoi]
+        dxdnobs_noi = tf.gather(dxdnobs[self.npoi :], self.indata.noigroupidxs)
+        dxdnobs = tf.concat([dxdnobs_poi, dxdnobs_noi], axis=0)
+
+        if self.externalCovariance:
+            data_cov = tf.linalg.inv(self.data_cov_inv)
+            data_stat = tf.einsum("ij,jk,ik->i", dxdnobs, data_cov, dxdnobs)
```
bendavid, line 286: Can you at least add a comment with the corresponding matrix multiplication formula?
(it's dxdnobs @ data_cov @ dxdnobs.T I guess?)
Is the einsum really better/faster?

combinetf2/fitter.py
```diff
@@ -0,0 +1,1410 @@
...
+                tf.ragged.constant(self.indata.systgroupidxs, dtype=tf.int32),
+                fn_output_signature=tf.TensorSpec(
+                    shape=(dxdtheta0_squared.shape[0],), dtype=tf.float64
+                ),
+            )
+            impacts_grouped_syst = tf.transpose(impacts_grouped_syst)
+            impacts_grouped = tf.concat([impacts_grouped_syst, impacts_grouped], axis=1)
+
+        # global impacts of unconstrained parameters are always 0, only store impacts of constrained ones
+        impacts = dxdtheta0[:, self.indata.nsystnoconstraint :]
```
bendavid, line 321: Even if they're always zero doesn't it make sense to store it for the sake of keeping all lists of parameters consistent?

scripts/fit.py
```diff
@@ -0,0 +1,564 @@
...
+        common_elements, idxs, idxs_ext = np.intersect1d(
+            parms, parms_ext, assume_unique=True, return_indices=True
+        )
+        xvals[idxs] = x_ext[idxs_ext]
+        covval[np.ix_(idxs, idxs)] = cov_ext[np.ix_(idxs_ext, idxs_ext)]
+
+        fitter.x.assign(xvals)
+        fitter.cov.assign(tf.constant(covval))
+    else:
+        fitter.profile = True
```
bendavid, line 390: Doesn't this need to be reset to False somewhere in case of multiple toys?  (e.g. in defaultassign() ?)

scripts/fit.py
```diff
@@ -0,0 +1,564 @@
...
+
+        if dofit:
+            fitter.minimize()
+
+        val, grad, hess = fitter.loss_val_grad_hess()
+        fitter.hess = hess
```
bendavid, line 396: What's the motivation for having different treatment/logic of hess and cov here? (assign by python reference vs assign by tensorflow value)

In general for things which are reused within a function, using tensor.assign should be better for avoiding recompilation of the function  (but I'm not sure which is better/worse for memory usage in case of large matrices)

combinetf2/fitter.py
davidwalter2: The equivalent would be ```tf.linalg.diag_part(dxdnobs @ data_cov @ tf.transpose(dxdnobs))```. The einsum should be faster since it avoids computing the full matrix, this is also what I see when I do a quick benchmarking.

scripts/fit.py
davidwalter2: Removed the hess attribute from the class

README.md
```diff
@@ -0,0 +1,94 @@
...
+### In a python virtual environment
+The simplest is to make a python virtual environment. It depends on the python version you are working with (tested with 3.9.18).
+First, make a python version, e.g. in the combinetf2 base directory (On some machines you have to use `python3`):
+```bash
+python -m venv env
+```
+The activate it and install the necessary packages
+```bash
+source env/bin/activate
+pip install wums tensorflow numpy h5py hist scipy matplotlib mplhep seaborn pandas plotly kaleido
```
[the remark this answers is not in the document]
davidwalter2: This is the description of setting up an environment if you clone the repo. I'm making it a pip installable as well and add a description

README.md
davidwalter2: OK, I can try to add optional dependencies in the pip install

README.md
davidwalter2: Developers should clone the repo and set up their environment

README.md
davidwalter2: In wremnants we have a wrapper "datagroups.addSystematic" that has the default set the "average" to be passed to tensorwriter.add_systematic. But for this standalone framework I would keep the default as asymmetric as it's more direct to what the user provides, and as the user is optimizing things, and if it's appropriate for his analysis he can set it to "average"

combinetf2/fitter.py
davidwalter2: Removed this. I pass the hessian to the impacts calculation directly now, without storing it permanently

combinetf2/fitter.py
```diff
@@ -0,0 +1,1410 @@
...
+        cov_poi = self.cov[: self.npoi]
+        cov_noi = tf.gather(self.cov[self.npoi :], self.indata.noigroupidxs)
+        v = tf.concat([cov_poi, cov_noi], axis=0)
+        impacts = v / tf.reshape(tf.sqrt(tf.linalg.diag_part(self.cov)), [1, -1])
+
+        # impact data stat # FIXME: This might not be correct in case noi != systsnoconstraint
+        nstat = self.npoi + self.indata.nsystnoconstraint
+        hess_stat = self.hess[:nstat, :nstat]
+        identity = tf.eye(nstat, dtype=hess_stat.dtype)
+        inv_hess_stat = tf.linalg.solve(hess_stat, identity)  # Solves H * X = I
```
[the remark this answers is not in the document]
davidwalter2: Changed to use tf.linalg.inv now

combinetf2/fitter.py
davidwalter2: Also changed it here

combinetf2/workspace.py
```diff
@@ -0,0 +1,493 @@
...
+        self.npoi = fitter.npoi
+        self.noigroupidxs = fitter.indata.noigroupidxs
+
+        self.extension = "hdf5"
+        self.file_path = self.get_file_path(outdir, outname, postfix)
+
+    def __enter__(self):
+        """Open the file when entering the context."""
+        print(f"Write results in file {self.file_path}")
+        self.fout = h5py.File(self.file_path, "w")
```
bendavid, line 85: I don't think this is the correct way of "wrapping" the context manager functionality (trying to find an appropriate example of the right way to do it though)

bendavid: so to do this consistently I think the self.fout =  statement should be moved to the end of __init__ and the __exit__ function should return False in order to properly re-raise any exceptions

combinetf2/fitter.py
```diff
@@ -0,0 +1,1417 @@
...
+        # scan positive side and negative side independently to profit from previous step
+        for sign in [-1, 1]:
+            param_scan_values = xval[idx].numpy() + sign * param_offsets
+            for i, ixval in enumerate(param_scan_values):
+                if i == 0:
+                    continue
+
+                self.x.assign(tf.tensor_scatter_nd_update(self.x, [[idx]], [ixval]))
+
+                def scipy_loss(xval):
```
bendavid, line 1108: For the future: worth testing modifying the loss/grad/hess functions to imply 1 for the corresponding hessian element instead of 0, since this might allow the minimizer to converge more efficiently

combinetf2/tensorwriter.py
```diff
@@ -0,0 +1,814 @@
...
+
+        for p, u in zip(process, uncertainty):
+            norm = self.dict_norm[channel][p]
+            syst = norm * u
+            logk = self.get_logk(syst, norm)
+            self.book_logk_avg(logk, channel, p, name)
+
+        self.book_systematic(name, groups=groups, profile=profile)
+
+    def add_systematic(
```
bendavid, line 194: Strongly motivated to allow passing multiple systematics in one histogram (e.g. by specifying syst_axes) but this can be added later.

bendavid:
Modulo addressing remaining comments this looks good.

For the future we can consider refactoring of the TensorWriter to build the tensor on the fly such that one could build the tensor and run the fit without writing out to disk if desired.

combinetf2/workspace.py
davidwalter2: OK, moved the file opening to the __init__ function. I think where we open the file only matters in practice if the Workspace object is used not as context manager. I.e. as ```ws = Workspace()```. 
From https://book.pythontips.com/en/latest/context_managers.html I read that if nothing is returned in the __exit__ function the exception is raised, so it doesn't matter if we return False or nothing.

combinetf2/tensorwriter.py
davidwalter2: There is already an open issue on this: https://github.com/WMass/combinetf2/issues/5

combinetf2/fitter.py
```diff
@@ -0,0 +1,1417 @@
...
+            impacts_bbb = tf.sqrt(tf.nn.relu(impacts_bbb_sq))  # max(0,x)
+            impacts_bbb = tf.reshape(impacts_bbb, (-1, 1))
+            impacts_grouped = tf.concat([impacts_data_stat, impacts_bbb], axis=1)
+        else:
+            impacts_data_stat = tf.sqrt(tf.linalg.diag_part(inv_hess_stat))
+            impacts_data_stat = tf.reshape(impacts_data_stat, (-1, 1))
+            impacts_grouped = impacts_data_stat
+
+        if len(self.indata.systgroupidxs):
+            impacts_grouped_syst = tf.map_fn(
```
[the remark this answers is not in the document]
davidwalter2: From some tests it's not totally, running mw impacts I get 3.6s for the python loop vs. 3.8s for the current implementation. But I think when using a pyhthon loop tensorflow can't optimize the code in compilation anymore, so it probably would require some re-structuring. For this reason I leave this for the future.

combinetf2/workspace.py
bendavid: Ok yes this is fine then.

bendavid:
Ok this PR is good to be merged then.  (But I don't currently have permission to do it)

(Please don't merge the narf or wremnants PR's yet though)

PR #9, opened 2025-03-11
Masked channels

davidwalter2:
- Add masked channels similar as it was done in combinetf1. 
- Masked channels are always added trailing while not masked channels are added leading
- To avoid calculating the yields in the masked channel during likelihood minimization the masked channels can be clipped in ```_compute_yields```
- Bin by bin uncertainties in the masked channel are treated in the same way with the exception that they are never profiled and don't contribute to the likelihood. 
- Previously the fitter had an attribute 'profile' as global variable to steer if the systematics are profiled or not. However, this is captured in the tf graph and would not be able to change one the graph is traced. But since his is expected to change between prefit and postfit, it was changed to use function parameter instead.

### DIFF
--- a/combinetf2/h5pyutils.py
+++ b/combinetf2/h5pyutils.py
@@ -82,27 +82,3 @@ def writeSparse(indices, values, dense_shape, h5group, outname, maxChunkBytes=10
     outgroup.attrs["dense_shape"] = np.array(dense_shape, dtype="int64")
 
     return nbytes
-
-
-def simple_sparse_slice0end(in_sparse, end):
-    """
-    Slice a tf.sparse.SparseTensor along axis 0 starting 0 to the 'end'.
-    """
-
-    # Convert dense_shape, indices, and values to tensors if they aren't already
-    dense_shape = in_sparse.dense_shape
-    indices = in_sparse.indices
-    values = in_sparse.values
-
-    # Compute output dense shape after slicing
-    out_shape = tf.concat([[end], dense_shape[1:]], axis=0)
-
-    # Filter rows: select entries where indices[:, 0] < end
-    mask = indices[:, 0] < end
-    selected_indices = tf.boolean_mask(indices, mask)
-    selected_values = tf.boolean_mask(values, mask)
-
-    # Return the sliced sparse tensor
-    return tf.sparse.SparseTensor(
-        indices=selected_indices, values=selected_values, dense_shape=out_shape
-    )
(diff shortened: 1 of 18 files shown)


bendavid:
Please add something about the change of the "profile" behaviour to the PR description.

davidwalter2:
> Please add something about the change of the "profile" behaviour to the PR description.

Done

combinetf2/inputdata.py
```diff
@@ -80,7 +103,18 @@ def __init__(self, filename, pseudodata=None, normalize=False):
                                 name="obs",
                             )
                         ]
-                    }
+                    },
+                    "ch1": {
```
bendavid, line 107: Should this channel be added in case there are no masked channels?

bendavid: Also probably "masked" should appear in the channel name.

combinetf2/tensorwriter.py
```diff
@@ -96,8 +96,8 @@ def add_pseudodata(self, h, name=None, channel="ch0"):
             )
         self.dict_pseudodata[channel][name] = self.get_flat_values(h)
 
-    def add_process(self, h, name, channel="ch0", signal=False):
-        self._check_hist_and_channel(h, channel)
+    def add_process(self, h, name, channel="ch0", signal=False, masked=False):
```
bendavid, line 99: General question:  Why not just require the user to always call add_channel before calling add_process? (in this case the "masked" parameter would not be needed here, where it could be confusing/inconsistent)

bendavid:
One general comment which can be addressed in the future is that it could make sense to provide observed data for masked channels in some cases (but it would have to be optional clearly)

combinetf2/h5pyutils.py
```diff
@@ -82,3 +82,27 @@ def writeSparse(indices, values, dense_shape, h5group, outname, maxChunkBytes=10
     outgroup.attrs["dense_shape"] = np.array(dense_shape, dtype="int64")
 
     return nbytes
+
+
+def simple_sparse_slice0end(in_sparse, end):
```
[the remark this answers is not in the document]
davidwalter2: I tested and using
```
snormnorm_sparse = tf.sparse.slice(snormnorm_sparse, [0,0], [self.indata.nbins, self.indata.nproc])
```
Gives the same result. 
But running the full fit I get the error 'LookupError: gradient registry has no entry for: SparseSliceGrad' so we probably need to keep this helper function. I move it in a separate file "tfhelpers.py" if there is no objection.

combinetf2/tensorwriter.py
davidwalter2: Yes we could be more restrictive and require the user to always call add_channel first. Probably less error prone.

davidwalter2:
> One general comment which can be addressed in the future is that it could make sense to provide observed data for masked channels in some cases (but it would have to be optional clearly)

OK, I addressed all the points above and added an issue for this one: https://github.com/WMass/combinetf2/issues/18
Depending on how fast you review this PR and on my availability, I can either implement it on top of this PR or make a new one.

PR #10, opened 2025-03-12
Project inclusive yield

davidwalter2:
- By using the option "--project channel" without axes the projection of the total yield is performed. An Integer axis "yield" with a single bin is created to make the histogram and store it as any other histogram.
- Support print and plot impacts on observables (hisograms) e.g. for masked channels

### DIFF
--- a/bin/combinetf2_plot_hists_uncertainties.py
+++ b/bin/combinetf2_plot_hists_uncertainties.py
@@ -316,6 +316,12 @@ def make_plot(
             # unrolled ND histograms
             h_impact = hh.unrolledHist(h_impact, obs=[a.name for a in axes])
 
+        if len(h_impacts.axes) == 1:
+            h_impact = hist.Hist(
+                hist.axis.Integer(0, 1, name="yield", overflow=False, underflow=False),
+                data=h_impact.value,
+            )
+
         hep.histplot(
             h_impact,
             xerr=False,
@@ -339,6 +345,7 @@ def make_plot(
             yerr=False,
             histtype="step",
             color="black",
+            linestyle="dashed",
             label=translate_label.get("Total", "Total"),
             density=False,
             ax=ax1,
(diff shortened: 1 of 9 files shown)


bin/combinetf2_plot_hists_uncertainties.py
```diff
@@ -316,6 +316,12 @@ def make_plot(
             # unrolled ND histograms
             h_impact = hh.unrolledHist(h_impact, obs=[a.name for a in axes])
 
+        if len(h_impacts.axes) == 1:
```
bendavid, line 319: Would this cause problems in case there was a channel which really just had only one axis?

bin/combinetf2_plot_hists_uncertainties.py
```diff
@@ -421,17 +428,19 @@ def make_plots(
 
     hist_total = result["hist_postfit_inclusive"].get()
 
+    if hist_impacts is not None:
+        uncertainties = np.array(hist_impacts.axes["impacts"], dtype=str)
```
bendavid, line 432: I realize the code was like this before, but why does this actually need to be an np.array?

bendavid:
This needs to be rebased now.

bin/combinetf2_plot_hists_uncertainties.py
davidwalter2: Yes because the one axis is the impact axis, later a bin in this impact axis is selected which gives an object that only has "value" and "variance" instead of "values()" and "variances()" and the code would crash

davidwalter2: This happens for example for a histogram with only charge axis and then split plots by charge

bin/combinetf2_plot_hists_uncertainties.py
davidwalter2: Yes this is to get an array of the bin labels for a StrCategory axis. The hist_impacts.axes["impacts"] gives a StrCategory object which doesn't support things like slicing, e.g. "[:]" that is used later.

bendavid: right but regular python strings should support slicing operations as well

davidwalter2: You mean list? It should be a list instead of a np.array?

davidwalter2:
> This needs to be rebased now.

Done

bin/combinetf2_plot_hists_uncertainties.py
bendavid: Ah yes right, it should be a list in this case

```list(hist_impacts.axes["impacts"])``` should work

bendavid: I'll merge as-is in any case and this can be adjusted later.

PR #16, opened 2025-03-17
Remove python conditionals from tf.function that are known at init

davidwalter2:
A separate branch in the tf graph is build for each python conditional. But many python conditionals are known at init and the branches will never be evaluated. To minimize the graph building time and graph size, these conditionals are moved to 'Fitter.__init__' to define relevant function blocks there. The time saved is minimal at the moment but it allows to support more options in the future w/o degrading the performance. This also cleans up the code.

### DIFF
--- a/bin/combinetf2_fit.py
+++ b/bin/combinetf2_fit.py
@@ -27,6 +27,12 @@ def make_parser():
     parser.add_argument(
         "--noColorLogger", action="store_true", help="Do not use logging with colors"
     )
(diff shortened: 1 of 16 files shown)


bendavid:
Can this one be closed given the previous discussion and other changes included in https://github.com/WMass/combinetf2/pull/9 ?

davidwalter2:
Yes, some of the factorizing might be good for better code modularity, but can be done later/differently

PR #19, opened 2025-03-22
Physics model

davidwalter2:
Support for (custom) physics models. 
- A new functionality is introduced to compute transformations of observables and parameter in `combinetf2/physicsmodels/` . 
  - The `Basemodel` is simply taking the observables as they come.
  - This `Basemodel` can be inherited to implement custom physics models. 
  - One custom physics model is `Project` that replaces the previous `--Project` option. 
  - The physics model must provide a few things that are used in the fitter and workspace:
    - a differentiable `compute` function where the predictions are calculated, given the flat vector of observables and/or parameters,
    - some attributes to define the behavior of the fit and the histograms in the fitresult file,
    - a `parse_args` function to parse the command line argument for the physics model (optional)
    - the physics model must be defined in a separate file under `combinetf2/physicsmodels/` and the class name must be the same as the file name just that the file name must start with a small character and the class name with a capital one. The file name also must end with ".py".
- Custom physics models can be called in `combinetf2_fit.py` using `-m <model name> args` e.g. `-m Project ch0 pt`
  - Multiple custom physics models can be specified.
- The normalization of inputdata has been removed. Normalized distributions can now be produced via the physics model functionality. E.g.  `-m Normalize ch0 pt`
- Another physics model that is added is `Ratio` to compute ratios between channels, processes, or bins
- Another physics model for parameters is `Param` to multiply a scale and add an offset to the parameter is in `tests/param_model.py`

I've changed again the structure of the result dict, the result of each physics model is stored under
`fitresult["physics_model"][model.key]` to keep each physics model separate. Also channels are under this directory since physics models can define multiple channels and add new channels in the results.


combinetf2/physicsmodels.py
```diff
@@ -0,0 +1,164 @@
...
+    ----------
+    """
+
+    def __init__(self, indata):
+        self.name = "basemodel"  # Used for the output dictionary
+        # channel info of the resulting histogram, needed for the writing
+        self.channel_info = indata.channel_info
+        # if data histograms are stored or not, and if chi2 is calculated
+        self.has_data = True
+        self.normalize = False
```
bendavid, line 18: Why does normalize have to be a member of the base class like this?  ie does it trigger any "special" behaviour beyond just being a detail of the particular derived class that implements it?

davidwalter2: It's needed to calculate ndf but I've removed it here and assume it as False if not a member of the class

combinetf2/physicsmodels.py
```diff
@@ -0,0 +1,399 @@
...
+        self.has_data = True
+        # a list of strings, the result will be written under these keys
+        self.identifiers = []
+
+    def make_fun(self, fun_flat, inclusive=True):
+        return fun_flat
+
+    def get_data(self, data):
+        fun = self.make_fun(lambda: data)
+        return fun(), None
```
bendavid, line 25: Couldn't this just directly be ```return data, None``` ?

combinetf2/physicsmodels.py
```diff
@@ -0,0 +1,399 @@
...
+                i for i, n in enumerate(channel_axes_names) if n in selections.keys()
+            ]
+        else:
+            self.selections = None
+            self.selection_idxs = None
+
+        self.channel_axes = channel_axes
+
+
+class Ratio(PhysicsModel):
```
bendavid, line 212: This class doesn't provide a get_data method, is it optional in the API?  (and what does the code do if it's missing?)

combinetf2/physicsmodels.py
```diff
@@ -0,0 +1,399 @@
...
+                exp /= norm
+            exp = tf.reshape(exp, out_shape)
+            exp = tf.reduce_sum(exp, axis=self.proj_idxs)
+            exp = tf.transpose(exp, perm=transpose_idxs)
+
+            return exp
+
+        return fun
+
+    def get_data(self, data, cov=None):
```
bendavid, line 119: Should the cov parameter be part of the api in general?  (ie should the base class have this as well?)

combinetf2/physicsmodels.py
```diff
@@ -0,0 +1,399 @@
...
+            cov = tf.transpose(
+                cov, perm=perm_idxs + [i + len(perm_idxs) for i in perm_idxs]
+            )
+
+            if self.normalize:
+                norm = tf.reduce_sum(val)
+                jac = tf.eye(tf.shape(val)[0]) / norm - tf.expand_dims(val, 1) / norm**2
+                cov = tf.matmul(jac, tf.matmul(cov, tf.transpose(jac)))
+
+            var = tf.linalg.diag_part(cov)
```
bendavid, line 141: In case the data has a covariance matrix, in general the projected data will also have a covariance matrix with off diagonal terms.  This could be computed and propagated here.  (API would need to be extended I guess?)

bendavid:
Can a user actually define a custom physics model in the current setup?  (the use of the "models" dictionary to map magic strings to class names would seem to make this more complicated)

bin/combinetf2_plot_hists_uncertainties.py
```diff
@@ -576,34 +577,35 @@ def main():
...
 
-        make_plots(
-            outdir,
-            result,
-            axes=info["axes"],
-            channel=channel,
-            lumi=info.get("lumi", None),
-            **opts,
-        )
+            if model == "basemodel":
```
bendavid, line 590: Ideally there would be enough metadata to avoid needing to hardcode behaviour based on which model was used?

bin/combinetf2_fit.py
```diff
@@ -212,19 +214,15 @@ def make_parser():
...
-        action="store_true",
-        help="Normalize prediction and systematic uncertainties to the overall event yield in data",
-    )
-    parser.add_argument(
-        "--project",
+        "-m",
+        "--physicsModel",
         nargs="+",
         action="append",
         default=[],
```
bendavid, line 221: Does it make more sense to just make the default argument ["basemodel"] rather than appending this to the beginning of the list later on in the code?

bin/combinetf2_fit.py
```diff
@@ -573,11 +508,19 @@ def main():
...
     logger = logging.setup_logger(__file__, args.verbose, args.noColorLogger)
 
-    indata = inputdata.FitInputData(
-        args.filename, args.pseudoData, normalize=args.normalize
-    )
+    indata = inputdata.FitInputData(args.filename, args.pseudoData)
     ifitter = fitter.Fitter(indata, args)
 
+    # models for observables
+    models = [pm.models["basemodel"](indata)]
```
bendavid, line 515: See earlier comment about accomplishing this via default arguments.

bendavid:
The residual function used to compute the chi^2 here https://github.com/WMass/combinetf2/blob/f706f692fb59387e60e3b7df67a76f5a366c5ad2/combinetf2/fitter.py#L929-L938 needs to take into account the PhysicsModel rather than just using compute_yields

But actually there's a more general issue here.  In the original version of expected_events_projection the data statistical uncertainty was effectively automatically propagated to the projection by just running the corresponding transformation on the residual.

I wonder if this can be sufficiently generalized such that
a) an explicit get_data() function is not needed for the physics models
b) resulting post-transformation variance/covariance for the observed data doesn't need to be computed by hand

combinetf2/physicsmodels.py
davidwalter2: In this case yes but if a more complex PhsicsModel is implemented then this should gives the right values for data without the need to override the function.

combinetf2/physicsmodels.py
davidwalter2: The get_data is inherited from PhysicsModel which calls self.make_fun, in this way the ratio for data should be correct in case it's done inclusively (w/o dividing processes where data makes no more sense). Of course the variance is not correct, so eventually a get_data would need to be implemented. But I'm thinking if the get_data can be derived from the make_fun also for the variance so I didn't do more work on this for now

combinetf2/physicsmodels.py
davidwalter2: You mean a member of the PhysicsModel? I think it should be treated in the same way as the data, i.e. a member of the fitter or indata class that get's passed in the get_data function.

bendavid: I meant why doesn't the base class version of get_data also have a "cov" parameter?

bendavid:
Thinking a bit more I think the slightly awkward ```make_fun``` semantics can be avoided in the derived classes and just be implemented in the base class.

e.g. in the base class something like

```
# passthrough version for base class
def get_expected(self, expected_flat):
    return expected_flat

# generic version which should not need to be overridden
def make_fun(self, fun_flat):
    def fun():
        exp = self.get_expected(fun_flat())
        return exp

    return fun
```

Then the derived classes would only have to override get_expected (and the logic is much simpler also if users want to implement custom ones)

davidwalter2:
> Thinking a bit more I think the slightly awkward `make_fun` semantics can be avoided in the derived classes and just be implemented in the base class.
> 
> e.g. in the base class something like
> 
> ```python
> # passthrough version for base class
> def get_expected(self, expected_flat):
>     return expected_flat
> 
> # generic version which should not need to be overridden
> def make_fun(self, fun_flat):
>     def fun():
>         exp = self.get_expected(fun_flat())
>         return exp
> 
>     return fun
> ```
> 
> Then the derived classes would only have to override get_expected (and the logic is much simpler also if users want to implement custom ones)

Sounds good! I also think the get_data is not needed for derived classes, can just derive the `get_expected(data)` and get the data uncertainty from there. Have to think about a better name, maybe `compute`

combinetf2/physicsmodels.py
davidwalter2: Done

combinetf2/physicsmodels.py
davidwalter2: Yes, done

bin/combinetf2_fit.py
davidwalter2: Done

bin/combinetf2_fit.py
davidwalter2: Done

davidwalter2:
> Can a user actually define a custom physics model in the current setup? (the use of the "models" dictionary to map magic strings to class names would seem to make this more complicated)

I don't see a better way right now but open for suggestions, considering
- some models can be templated, e.g. using `normalize` switch, these get specific in the `models` dict
- the `models` dict keeps track of all defined models such that the user can find out what models exist from the command line

bin/combinetf2_plot_hists_uncertainties.py
davidwalter2: This is needed only for the basemodel since the basemodel is unlike other models stored on a higher level in the results dictionary, directly under the channel key. The other models are stored under [<model name>][<identifier>] to avoid naming conflicts between different models. We could put basemodel some levels lower under ["basemodel"][<identifier>] for the sake of consistency with other models, then this check would not be needed. But given that the basemodel is a special case I think it's fine to keep it as it is.

davidwalter2:
> > Can a user actually define a custom physics model in the current setup? (the use of the "models" dictionary to map magic strings to class names would seem to make this more complicated)
> 
> I don't see a better way right now but open for suggestions, considering
> 
> * some models can be templated, e.g. using `normalize` switch, these get specific in the `models` dict
> * the `models` dict keeps track of all defined models such that the user can find out what models exist from the command line

OK this has now been updated, I've edited the description of the PR accordingly

combinetf2/physicsmodels/basemodel.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class Basemodel:
+    """
+    A class to output histograms without any transformation, can be used as base class to inherit custom physics models from.
+    """
+
+    name = "basemodel"
+    need_observables = True  # if observables should be provided to the compute function
+    need_params = (
```
bendavid, line 11: Is there really any advantage to doing this?  The code is arguably simpler if all models take both the observables and the parameters and then just don't use one or the other if not needed.

combinetf2/physicsmodels/helpers.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+            if name != "helpers"
+        ]
+        for mod in available_modules:
+            print(f"- {mod}")
+        return None
+
+
+def instance_from_module(module_name, *args, **kwargs):
+    class_name = module_name.capitalize()
```
bendavid, line 33: I would just make the user specify the class name directly with capitalization as-is.

ie provide directly "Project" on the command line as appropriate

combinetf2/fitter.py
```diff
@@ -980,167 +972,44 @@ def fun():
...
+                compute_cov=True,
             )
 
-        extra_axes = []
-        if not inclusive:
-            exp_axes.append(self.indata.axis_procs)
-            hist_axes.append(self.indata.axis_procs)
-            extra_axes.append(self.indata.axis_procs)
+            chi2val = self.chi2(pred - data, pred_cov + data_cov).numpy()
+            ndf = tf.size(pred).numpy() - getattr(model, "normalize", False)
```
bendavid, line 1009: Right, rather than using the normalize attribute here, the model should just have an attribute specifying how many degrees of freedom to subtract.

bendavid:
Possible radical suggestion to consider:

Since the splitting into channels and reshaping is relatively generic, could/should this happen in BaseModel such that "compute" gets passed not a flat tensor but a dictionary of tensors already split by channel and reshaped.  (and then the functions can return the same rather than needing to flatten the output)

Possible drawback is that it will make operations that need the flattened expectation less efficient (the chi^2 computation for example) since it will need to be concatenated.

Maybe the model can optionally implement compute_flat to do it more efficiently in case the splitting by channel is not needed/convenient. (ie the passthrough case for BaseModel)

combinetf2/physicsmodels/basemodel.py
davidwalter2: This is to avoid computing the observables if they are not needed.
I checked, it's running both the tracing and the computation through the flat_fun() also if observables are not used further.
I will simplify that params are always provided since they don't have to be computed.

combinetf2/physicsmodels/helpers.py
davidwalter2: OK!

bin/combinetf2_plot_hists_cov.py
```diff
@@ -231,92 +233,106 @@ def main():
...
-                outdir, h_cov, args, channel, projection_axes, config=config, meta=meta
+                outdir,
+                h_cov,
+                args,
+                config=config,
+                meta=meta,
+                suffix=model + "_" + instance_name,
             )
 
+    # for channel, info in channel_info.items():
```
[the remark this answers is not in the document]
davidwalter2: Repaired this finctionality

bin/combinetf2_plot_hists_uncertainties.py
davidwalter2: No, not anymore

davidwalter2:
> Possible radical suggestion to consider:
> 
> Since the splitting into channels and reshaping is relatively generic, could/should this happen in BaseModel such that "compute" gets passed not a flat tensor but a dictionary of tensors already split by channel and reshaped. (and then the functions can return the same rather than needing to flatten the output)
> 
> Possible drawback is that it will make operations that need the flattened expectation less efficient (the chi^2 computation for example) since it will need to be concatenated.
> 
> Maybe the model can optionally implement compute_flat to do it more efficiently in case the splitting by channel is not needed/convenient. (ie the passthrough case for BaseModel)

I've added a class "PhysicsModelChannel" that can be inherited for single channel physics models. I thought about making it generic in "PhysicsModelChannels" but this would require much more work since the computation of data uncertainty and the histogram writing downstream would also need to be be adapted. Maybe it's better to keep it as it is to avoid code duplication.

combinetf2/physicsmodels/physicsmodel.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def compute(self, params, observables):
+        return observables
+
+    def compute_per_process(self, params, observables):
+        return self.compute(params, observables)
+
+    def compute_flat(self, params, observables):
+        exp = tf.reshape(observables[self.start : self.stop], tuple(self.channel_shape))
+        return self.compute(params, exp)
```
bendavid, line 102: This function now returns the data carrying the shape of the channel, not necessarily flat.  Is this intended or should it be flattened back out here to behave the same as for other models?

davidwalter2: As long as it's only one channel it's no problem if it's not flat

combinetf2/physicsmodels/helpers.py
```diff
@@ -0,0 +1,136 @@
...
+                f"Model {class_name} not found, available baseline models are {baseline_models.keys()}"
+            )
+        module_name = f"combinetf2.physicsmodels.{baseline_models[class_name]}"
+
+    # Try to import the module
+    module = importlib.import_module(module_name)
+
+    try:
+        cls = getattr(module, class_name)
+    except AttributeError:
```
bendavid, line 35: Probably better to just crash here rather than skip the missing model silently

bendavid:
PR description needs to be updated at least with "project" -> "Project" etc

combinetf2/physicsmodels/physicsmodel.py
bendavid: But what happens downstream then?  Does it get flattened out again somewhere?

davidwalter2: No it's not flattened but if it is flat it gets reshaped into the histogram shape:
https://github.com/WMass/combinetf2/blob/main/combinetf2/workspace.py#L147-L149

bendavid: I think it's better to flatten it to avoid a heterogeneous situation downstream.

This code is already getting a bit complicated for my taste
https://github.com/WMass/combinetf2/blob/c70b43730d788ffff299f8a3b9f5981c82c3ee82/combinetf2/workspace.py#L149-L154

davidwalter2: OK

bendavid:
Is this ready to merge now?

davidwalter2:
Yes, I would say so

bendavid:
Hmm it looks like the CI web pages are down so let's wait to have a final look at that first.

PR #20, opened 2025-03-23
Additional fit options and fixes/improvements to impacts and toys

bendavid:
Add support for Gaussian systematic uncertainties (configured through the TensorWriter) and binByBin statistical uncertainties (configurable at run time)

The Gaussian binByBinStat case can now also be used with an externalCovarianceMatrix

In case the likelihood is purely quadratic it is now minimized by a matrix solve operations (via Cholesky decomposition) rather than an iterative fit.

Fit convergence criteria now tries to minimize the likelihood as much as possible within numerical precision.

Both bayesian and frequentist toys are now properly supported for binByBinStat.
Note that to accomplish this the "profile" options have been restructured again.  Since beta is now an explicit variable, a profiled central value with non-profiled gradients can be achieved with profile=False (since the correct values of beta are now set as part of the minimization procedure).  So the profile_grad option is no longer needed.

### DIFF
--- a/bin/combinetf2_fit.py
+++ b/bin/combinetf2_fit.py
@@ -402,6 +402,8 @@ def save_hists(args, fitter, ws, prefit=True, profile=False):
 
 def fit(args, fitter, ws, dofit=True):
 
+    edmval = None
+
     if args.externalPostfit is not None:
         # load results from external fit and set postfit value and covariance elements for common parameters
         with h5py.File(args.externalPostfit, "r") as fext:
@@ -436,8 +438,22 @@ def fit(args, fitter, ws, dofit=True):
         if dofit:
             fitter.minimize()
 
+        # compute the covariance matrix and estimated distance to minimum
+
         val, grad, hess = fitter.loss_val_grad_hess()
-        fitter.cov.assign(tf.linalg.inv(hess))
+
+
+        # use a Cholesky decomposition to easily detect the non-positive-definite case
+        chol = tf.linalg.cholesky(hess)
+        del hess
+
+        gradv = grad[..., None]
+        edmval = 0.5 * tf.linalg.matmul(gradv, tf.linalg.cholesky_solve(chol, gradv), transpose_a=True)
+        print(edmval)
+
+        fitter.cov.assign(tf.linalg.cholesky_solve(chol, tf.eye(chol.shape[0], dtype=chol.dtype)))
+
+        del chol
 
         if args.doImpacts:
             ws.add_impacts_hists(*fitter.impacts_parms(hess))
@@ -454,6 +470,7 @@ def fit(args, fitter, ws, dofit=True):
     ws.results.update(
         {
             "nllvalfull": nllvalfull,
+            "edmval": edmval,
             "satnllvalfull": satnllvalfull,
             "ndfsat": ndfsat,
             "postfit_profile": args.externalPostfit is None,
--- a/bin/combinetf2_fit.py
+++ b/bin/combinetf2_fit.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import tensorflow as tf
+
 tf.config.experimental.enable_op_determinism()
 
 import argparse
@@ -8,7 +9,6 @@ import time
 
 import h5py
 import numpy as np
-
 
 from combinetf2 import fitter, inputdata, io_tools, workspace
 
@@ -190,7 +190,7 @@ def make_parser():
     parser.add_argument(
         "--binByBinStat_type",
         default="gamma",
-        choices = ["gamma", "normal"],
+        choices=["gamma", "normal"],
         help="probability density for bin-by-bin statistical uncertainties",
     )
     parser.add_argument(
@@ -363,7 +363,7 @@ def save_hists(args, fitter, ws, prefit=True, profile=False):
             inclusive=True,
             compute_variance=False,
             compute_variations=True,
-            profile=False
+            profile=False,
         )
 
         ws.add_expected_hists(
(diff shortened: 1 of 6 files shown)


bendavid:
There are quite some changes in the CI, will need to think about whether they make sense given the changes here.

combinetf2/inputdata.py
```diff
@@ -118,13 +118,17 @@ def __init__(self, filename, pseudodata=None, normalize=False):
                     }
 
             self.symmetric_tensor = self.metadata.get("symmetric_tensor", False)
+
+            # TODO this should be deprecated in favor of systematic_type normal
```
davidwalter2, line 122: Can you remove the exponential_transform functionality in this PR?

bin/combinetf2_fit.py
```diff
@@ -1,11 +1,14 @@
 #!/usr/bin/env python3
 
+import tensorflow as tf
+
+tf.config.experimental.enable_op_determinism()
```
davidwalter2, line 5: What is this doing?

bin/combinetf2_fit.py
```diff
@@ -184,6 +187,12 @@ def make_parser():
         action="store_true",
         help="add bin-by-bin statistical uncertainties on templates (adding sumW2 on variance)",
     )
+    parser.add_argument(
+        "--binByBinStat_type",
```
davidwalter2, line 191: We use CamelCase for arguements

combinetf2/fitter.py
```diff
@@ -654,9 +748,16 @@ def _compute_yields_noBBB(self, compute_norm=False, full=True):
...
             logsnorm = tf.reshape(logsnorm, [nbins, self.indata.nproc])
 
-            snorm = tf.exp(logsnorm)
+            if self.indata.systematic_type == "log_normal":
+                snorm = tf.exp(logsnorm)
+                snormnorm = snorm * norm
+            elif self.indata.systematic_type == "normal":
+                snormnorm = norm + logsnorm
+            else:
+                raise RuntimeError(
```
davidwalter2, line 757: Do we need to check this downstream everytime? Suggest to check upstream in indata once

combinetf2/fitter.py
```diff
@@ -19,9 +23,23 @@ def __init__(self, indata, options):
...
+            options.externalCovariance
+            and options.binByBinStat
+            and options.binByBinStat_type != "normal"
+        ):
             raise Exception(
-                'option "--binByBinStat" currently not supported for options "--externalCovariance"'
+                'option "--binByBinStat" only for options "--externalCovariance" with "--binByBinStat_type normal"'
+            )
+
+        if self.binByBinStat_type not in ["gamma", "normal"]:
```
davidwalter2, line 35: This can never happen because choices for the options are given, suggest to delete it

davidwalter2 (commented):
Can you add some lines in the readme for documentation?

combinetf2/fitter.py
bendavid: It could if the fitter was used outside of combinetf2_fit and the arguments dictionary was constructed by hand

combinetf2/inputdata.py
bendavid: Done

bin/combinetf2_fit.py
bendavid: Changed

combinetf2/fitter.py
bendavid: Done

davidwalter2:
Can some checks be added to test all these new functionalities, e.g. in the CI?

combinetf2/fitter.py
```diff
@@ -484,8 +566,8 @@ def _expvar(self, fun_exp, compute_cov=False, compute_global_impacts=False):
 
         expcov_noBBB = expcov
         if self.binByBinStat:
-            varbeta0 = tf.math.reciprocal(self.indata.kstat)
-            exp_cov_BBB = pdexpdbeta0 @ (varbeta0[:, None] * tf.transpose(pdexpdbeta0))
+            varbeta = self.varbeta
+            exp_cov_BBB = pdexpdbeta @ (varbeta[:, None] * tf.transpose(pdexpdbeta))
```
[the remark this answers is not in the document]
bendavid: This is the non-profiled case so it's actually used for the prefit uncertainties.  Previously beta0 was used because beta was set to beta0 when profile=False.  Now that beta is an explicit variable we can differentiate with respect to it explicitly (and this makes more sense logically since it is the nuisance parameters being varied not the global observables in this case also for the other systematics)

combinetf2/tensorwriter.py
```diff
@@ -281,23 +284,33 @@ def get_logk(self, syst, norm, kfac=1.0):
...
-            _logk = kfac * np.log(syst / norm)
-        _logk_view = np.where(
-            np.equal(np.sign(norm * syst), 1),
-            _logk,
-            self.logkepsilon * np.ones_like(_logk),
-        )
+        # TODO clean this up and avoid duplication
+        if self.systematic_type == "log_normal":
+            # check if there is a sign flip between systematic and nominal
+            if self.exponential_transform:
```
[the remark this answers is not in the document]
bendavid: Done

combinetf2/fitter.py
```diff
@@ -1076,43 +1251,117 @@ def loss_val_grad_hessp(self, p):
...
+                return val.__array__(), grad.__array__()
 
-        return res
+            def scipy_hessp(xval, pval):
+                self.x.assign(xval)
+                p = tf.convert_to_tensor(pval)
+                val, grad, hessp = self.loss_val_grad_hessp(p)
+                return hessp.__array__()
+
+            class FitterCallback:
```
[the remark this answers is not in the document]
bendavid: Done, although if we want to use it probably also some of the try except logic can be encapsulated to avoid duplication.  In general the fact that the scipy evaluation functions change the state of the fitter is not super-nice so maybe a more general solution can be found (allowing x to be passed in as an argument for computing the loss rather than using the stored value, etc)

bendavid:
I think the CI changes are legitimate.  The calculation of the binByBin stat contribution to the postfit uncertainty on observables (and therefore also to the total postfit uncertainty on observables) was previously bugged.

One thing which is not nice is that we are effectively using a different calculation for the total uncertainty for observables vs parameters.  (related to the fact that global impacts don't exactly sum in quadrature to the total uncertainty)

TBC whether this can/should be changed or improved.

davidwalter2:
Something with the saturated test statistic is wrong, see: https://cmsmwbot.web.cern.ch/CombineTF2/PRValidation/PR20/2025_03_24/postfit_a_b_ch1_preliminary.png

PR #24, opened 2025-03-30
Text2hdf5 script

davidwalter2:
First version of text2hdf5 script tested. The option '--root' can be used to load the histograms with ROOT together with `narf` (faster) otherwise uproot is used. rateParams are implemented as unconstrained lnN uncertainties with a 1% variation.

To synchronize with combine it also implements
- asymmetric lnN uncertainties

In the tensorwriter, to make the tensor it also works now to provide the values and variances as numpy arrays instead of providing a histogram.

What is still missing and would be nice to add are: (but could be in a later PR)
- nuisance groups
- ...

A `debug_inputdata.py` script is implemented to run on the fit input .hdf5 file and check for anything pathological.

Also, combine allows to freeze parameters in the fit. In the example I tested there were 4 signals defined but in the fit only 1 or 2 are usually free floating, maybe we should add an option to do this in the fit itself. Right now we would need to modify the cards and run the conversion again.


bendavid:
Should probably move the DataCardParser and DataCardConverter classes out of the main text2hdf5 script and into a separate file (not in the bin directory)

bendavid:
General issue:
As far as I know regular combine supports TH1, TH2, TH3, so at least the axis information should be preserved (even with axis_0, axis_1, axis_2) in the metadata.  Right now it looks like everything is flattened.

bendavid:
Is this ready to merge now?

davidwalter2:
> Is this ready to merge now?

Yes!

PR #27, opened 2025-04-08
Add minimizer fit and debug options

davidwalter2:
- Add option to set minimizer method for fit (currently supported 'trust-krylov' and 'trust-exact' 
- For trust-exact hessian is needed and when used together with debug '-v 4' the condition number and edmval is calculated and printed
- Work on tf and scipy helper functions, 
  - only do the '__array__()' view when calling the scipy helper from the tf helper and not in the scipy helper itself since it's only needed when passing a tf tensor and not with a np.array.
- Improvements in the scripts to debug the inputdata tensor

This is on top of PR #26

### DIFF
--- a/bin/combinetf2_fit.py
+++ b/bin/combinetf2_fit.py
@@ -47,6 +47,13 @@ def make_parser():
         help="Postfix to append on output file name",
     )
     parser.add_argument(
+        "--minimizerMethod",
+        default="trust-krylov",
+        type=str,
+        choices=["trust-krylov", "trust-exact"],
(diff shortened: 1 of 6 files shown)


bendavid:
I would prefer to have a separate option for calculating the edmval at each fit iteration since it's quite slow.  (ie I would like to be able to also see the fit iterations without this)

davidwalter2:
> I would prefer to have a separate option for calculating the edmval at each fit iteration since it's quite slow. (ie I would like to be able to also see the fit iterations without this)

Added a '--diagnostics' options, this could also be used for other things in the future.

PR #28, opened 2025-04-08
Linearizing likelihood also in case of pois

davidwalter2:
- Multiplicative scaling of pois with n0 only in case systematic_type is normal
- Adding a test in CI for this fully linearized likelihood


davidwalter2:
Just noticed the systematic_type normal implementation in debugdata is missing, working on it

davidwalter2:
OK, this should be good to go now. 
PR validation also looks as expected, with small changes due to systematic_type=normal in the symmetric tensor

PR #30, opened 2025-04-22
Consistent handling of mc stat uncertainties in case of zero expected events or zero uncertainty

bendavid:
On top of https://github.com/WMass/combinetf2/pull/29


bendavid:
CI looks identical to https://github.com/WMass/combinetf2/pull/29 as expected

PR #31, opened 2025-04-23
fix saturated nll with updated mc stat handling

bendavid:
After https://github.com/WMass/combinetf2/pull/29 the saturated nll was inconsistent in the presence of gamma type binByBinStat in the presence of masked channels.

(The masked channel contribution was included in the nominal likelihood but not the saturated one)


bendavid:
No change in the CI (since I don't think the saturated nll appears on any of the plots there currently)

PR #33, opened 2025-04-23
Bugfix for grouped impacts

davidwalter2:


bendavid:
CI looks identical as expected (since none of the plots here depended on the fixed reshape operation)

PR #39, opened 2025-04-28
Add 'Select', 'AngularCoefficients', and 'LamTung' physics models

davidwalter2:
- The 'Term' class in the `physicsmodels/helper.py` is now able to process 'sum' and `rebin`, the rebin is realized using tf.math.segment_sum. This is now also accessible in the Ratio and Select models. In the parsing of selections slices can be either bin numbers (integer) or  bins from axis values by providing `j` e.g. `pt:slice(0j,2j)`
- The Select model is used to select a channel and optional, processes and bins. The bin selection is using the syntax <axis name>:<axis selection> and supports integers for selecting single bins, slices, 'sum' and 'rebin' where rebin takes a list of new edges. Multiple selections for a single axis are supported e.g. `pt:slice(2j,4j),pt:rebin(2,3,4)` or `pt:slice(0,10),pt:sum`
- The AngularCoefficients model is similar as the Ratio model but it assumes that numerator and denominator have the same channel and processes.
- The LamTung model computes the difference of the angular coefficients A0 and A0, i.e. A0-A2


davidwalter2:
> In the CI this plot changes https://cmsmwbot.web.cern.ch/CombineTF2/PRValidation/PR39/2025_04_29/uncertainties_ratio_a0_ch0_ch1_Ratio_ch0_ch1_NoneNone_bslice02bsum_preliminary.png
> 
> vs
> 
> https://cmsmwbot.web.cern.ch/CombineTF2/PRValidation/ReferenceRuns/2025_04_30_47a16a4/uncertainties_ratio_a0_ch0_ch1_Ratio_ch0_ch1_NoneNone_bslice02_preliminary.png
> 
> Is it intentional? (or is this plotting a different quantity now?)

OK there was a bug for slice() selection in case it has a step argument, should be fixed now

PR #41, opened 2025-05-02
Fix chi2 calculation

davidwalter2:
For the chi2, use old logic of the covariance matrix calculation with a few improvements
- data covariance matrix is now taken into account
- Variance of beta0 is now computed automatically 

Separate functions for likelihood parts
- To compute only the necessary part of the likelihood, define separate functions `_compute_lc` and `_compute_lbeta` and use when appropriate

Fixes a bug in `combinetf2/tfhelpers.py` where return statements were missing in the function to compute the condition number

Smaller updates on plotting scrips
- replace `--noChisq` by `--chisq` to specify which chi2 value to be shown in the plot (from saturated test or linear chi2 calculation, none, or automatic). Automatic takes saturated if applicable and linear chi2 calculation otherwise. 
- added `--showVariations` which takes `upper` `lower` or `both` to show variations only in the upper, lower or in both panels
- Fixes a bug where variations were not correctly computed in the lower panel in case the lower panel shows the difference

### DIFF
--- a/bin/combinetf2_plot_hists.py
+++ b/bin/combinetf2_plot_hists.py
@@ -1258,8 +1258,11 @@ def main():
             chi2, ndf, saturated_chi2 = get_chi2(
                 (
                     fitresult
-                    if (instance_key == "Basemodel" and args.chisq != "linear")
-                    or args.chisq == "saturated"
+                    if fittype == "postfit"
+                    and (
+                        (instance_key == "Basemodel" and args.chisq != "linear")
+                        or args.chisq == "saturated"
+                    )
                     else instance
                 ),
                 args.chisq in [" ", "none", None],
--- a/bin/combinetf2_plot_hists.py
+++ b/bin/combinetf2_plot_hists.py
@@ -304,6 +304,13 @@ def parseArgs():
         help="Only plot one sided variation (1) or two default two-sided (0)",
     )
     parser.add_argument(
+        "--showVariations",
+        type=str,
+        default="lower",
+        choices=["upper", "lower", "both"],
+        help="Plot the variations in the upper, lower panels, or both",
+    )
+    parser.add_argument(
         "--scaleVariation",
         nargs="*",
         type=float,
@@ -522,39 +529,37 @@ def make_plot(
             flow="none",
         )
 
-    linestyles = [
-        "--",
-        ":",
-        "-.",
-    ]
-    linewidth = 2
-    if hup is not None:
-        hep.histplot(
-            hup,
-            histtype="step",
-            color=varColors,
-            linestyle=linestyles[: len(hup)],
-            yerr=False,
-            linewidth=linewidth,
-            label=varLabels,
-            ax=ax1,
-            flow="none",
-        )
-    if (
-        hdown is not None
-        and any(h is not None for h in hdown)
-        and len(args.varOneSided) == 0
-    ):
-        hep.histplot(
-            hdown,
-            histtype="step",
-            color=varColors,
-            linestyle=linestyles[: len(hdown)],
-            yerr=False,
-            linewidth=linewidth,
-            ax=ax1,
-            flow="none",
-        )
+    if args.showVariations in ["upper", "both"]:
+        linewidth = 2
+        if hup is not None:
+            hep.histplot(
+                hup,
+                histtype="step",
+                color=varColors,
+                linestyle="-",
+                yerr=False,
+                linewidth=linewidth,
+                label=varLabels,
+                binwnorm=binwnorm,
+                ax=ax1,
+                flow="none",
+            )
+        if (
+            hdown is not None
+            and any(h is not None for h in hdown)
+            and len(args.varOneSided) == 0
+        ):
+            hep.histplot(
+                hdown,
+                histtype="step",
+                color=varColors,
+                linestyle="--",
+                yerr=False,
+                linewidth=linewidth,
+                binwnorm=binwnorm,
+                ax=ax1,
+                flow="none",
+            )
 
     if data:
         hep.histplot(
@@ -759,7 +764,11 @@ def make_plot(
                     label=label_unc,
                 )
 
-        if hup is not None and any(h is not None for h in hup):
+        if (
+            args.showVariations in ["lower", "both"]
+            and hup is not None
+            and any(h is not None for h in hup)
+        ):
             linewidth = 2
             scaleVariation = [
                 args.scaleVariation[i] if i < len(args.scaleVariation) else 1
@@ -769,6 +778,7 @@ def make_plot(
                 args.varOneSided[i] if i < len(args.varOneSided) else 0
                 for i in range(len(varNames))
             ]
+
             for i, (hu, hd) in enumerate(zip(hup, hdown)):
 
                 if scaleVariation[i] != 1:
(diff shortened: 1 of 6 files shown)


combinetf2/fitter.py
```diff
@@ -539,7 +539,136 @@ def global_impacts_parms(self):
...
+
+        if self.binByBinStat:
+            dvars = [self.x, self.ubeta, self.nobs]
+            residuals, pdresdx, pdresdbeta, pdresdobs = compute_derivatives(dvars)
+        else:
+            dvars = [self.x, self.nobs]
+            residuals, pdresdx, pdresdobs = compute_derivatives(dvars)
+            pdresdbeta = None
+
+        cov_dresdx = tf.matmul(self.cov, pdresdx, transpose_b=True)
```
bendavid, line 612: The calculation here is not correct.  What should be done is to compute the mc stat and systematics contribution to the covariance (using the "new" logic) and then add the data stat part using the "old" logic.

Instead what's being done is to compute the total uncertainty using the "new" logic and then add the data statistical contribution using the "old" logic.  (ie the data stat contribution is being partly double counted)

combinetf2/fitter.py
```diff
@@ -539,7 +539,136 @@ def global_impacts_parms(self):
...
+                    val = self._compute_loss()
+                grad = t1.gradient(val, self.x, unconnected_gradients="zero")
+            pd2ldxdnobs = t2.jacobian(grad, self.nobs)
+
+            dxdnobs = -self.cov @ pd2ldxdnobs
+
+            dresdnobs = pdresdobs + pdresdx @ dxdnobs
+
+            if self.externalCovariance:
+                data_cov = tf.math.sqrt(tf.linalg.inv(self.data_cov_inv))
```
bendavid, line 636: this isn't correct since this is an element-wise sqrt.  You would need the cholesky decomposition, but actually this is a waste of calculation because you could just directly calculate

```python
res_cov += dresdnobs @ data_cov @ dresdnobs.T
```
(or the appropriate variation with solve etc)

For consistency can do the same for the case for the non-externalCovariance case (e.g. using tf.linalg.LinearOperatorDiag to keep the logic the same while avoiding unnecessary calculations)

combinetf2/fitter.py
```diff
@@ -539,7 +541,132 @@ def global_impacts_parms(self):
...
+            dresdnobs = pdresdnobs
+            dresdbeta0 = pdresdbeta0
+
+            res_cov = pdresdx @ tf.matmul(self.cov, pdresdx, transpose_b=True)
+
+        res_cov_stat = dresdnobs @ (self.nobs[:, None] * tf.transpose(dresdnobs))
+        res_cov += res_cov_stat
+
+        if self.binByBinStat:
+            var_beta0 = tf.math.reciprocal(self.kstat)
```
bendavid, line 645: This should be 1.0 for binByBinStatType normal, and beta0/kstat for gamma
(the proper calculation of this could also be automated based on the appropriate gradients, in general it's (d2lbeta/dbeta2)/(d2lbeta/dbetadbeta0)^2 )

Some special handling is also needed for the betamask case for gamma uncertainties (where it would be sufficient to set var_beta0 to 0)

combinetf2/fitter.py
```diff
@@ -539,7 +541,132 @@ def global_impacts_parms(self):
...
+            res_cov_BBB = dresdbeta0 @ (var_beta0[:, None] * tf.transpose(dresdbeta0))
```
bendavid, line 646: I don't think this is correct in the profile=False case, because the expected doesn't depend on beta0 (the internal self.beta variable is used instead)

The solution is probably to calculate the full uncertainty including beta at line 643 and skip this part for profile=False

combinetf2/fitter.py
```diff
@@ -539,7 +541,132 @@ def global_impacts_parms(self):
...
+            res_cov = dresdtheta0 @ (var_theta0[:, None] * tf.transpose(dresdtheta0))
+        else:
+            # full dependency in partial derivatives
+            dresdnobs = pdresdnobs
+            dresdbeta0 = pdresdbeta0
+
+            res_cov = pdresdx @ tf.matmul(self.cov, pdresdx, transpose_b=True)
+
+        res_cov_stat = dresdnobs @ (self.nobs[:, None] * tf.transpose(dresdnobs))
+        res_cov += res_cov_stat
```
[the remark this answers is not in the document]
davidwalter2: But where is the data stat uncertainty accounted for the profile=False besides here?

combinetf2/fitter.py
```diff
@@ -539,7 +537,132 @@ def global_impacts_parms(self):
...
+
+            res_cov = dresdtheta0 @ (var_theta0[:, None] * tf.transpose(dresdtheta0))
+        else:
+            # full dependency in partial derivatives
+            dresdnobs = pdresdnobs
+            dresdbeta0 = pdresdbeta0
+
+            res_cov = pdresdx @ tf.matmul(self.cov, pdresdx, transpose_b=True)
+
+        res_cov_stat = dresdnobs @ (self.nobs[:, None] * tf.transpose(dresdnobs))
```
bendavid, line 641: This is not valid in the case of externalCovariance
(would have to use the data covariance here)

The logic can probably be simplified by using LinearOperatorDiag for the diagonal case to keep things more similar with vs without the externalCovarince

bin/combinetf2_plot_hists.py
```diff
@@ -747,21 +791,28 @@ def make_plot(
                         hdiff = hh.scaleHist(hdiff, scaleVariation[i])
                         hd = hh.addHists(hdiff, h_inclusive)
 
+                if diff:
```
bendavid, line 794: what are these changes about?  add something to the PR description about that?

combinetf2/fitter.py
```diff
@@ -1058,42 +1144,44 @@ def _compute_yields(self, inclusive=True, profile=True, full=True):
...
+            # covariance matrix is in general non invertible with ndf < n
+            # compute chi2 using pseudo inverse
+            chi_square_value = tf.transpose(res) @ tf.linalg.pinv(res_cov) @ res
+        else:
+            chi_square_value = tf.transpose(res) @ tf.linalg.solve(res_cov, res)
+
+        return tf.squeeze(chi_square_value), ndf
+
+    @tf.function
+    def chi2(self, model, profile):
```
bendavid, line 1168: It's not recommended to pass python objects (model) as parameters to tf.functions

At best it will result in retracing every single time, at worst incorrect/undefined behaviour.

combinetf2/fitter.py
```diff
@@ -1058,42 +1144,44 @@ def _compute_yields(self, inclusive=True, profile=True, full=True):
...
+
+        if profile:
+            residuals, res_cov = self._residuals_profiled(model.compute_flat)
+        else:
+            data, data_var, data_cov = model.get_data(self.nobs, self.data_cov_inv)
+            pred, pred_var, pred_cov, _1, _2 = self.expected_with_variance(
```
bendavid, line 1174: nested tf.functions here probably not optimal

combinetf2/fitter.py
davidwalter2: Fixed, I think here without the LinearOperatorDiag is simpler here since we have to work with the inverse of the data covariance in one case and the data variance in the other

bin/combinetf2_plot_hists.py
davidwalter2: Updated description

combinetf2/fitter.py
```diff
@@ -1151,6 +1081,95 @@ def expected_with_variance(self, *args, **kwargs):
...
+                    lbeta, _ = self._compute_lbeta(beta)
+
+                dlbetadbeta = t1.gradient(lbeta, self.ubeta)
+            pd2lbetadbetadbeta0 = t2.gradient(dlbetadbeta, self.beta0)
+
+            var_beta0 = pd2ldbeta2 / pd2lbetadbetadbeta0**2
+
+            if self.binByBinStatType == "gamma":
+                var_beta0 = tf.where(
+                    self.betamask, tf.constant(0.0, dtype=var_beta0.dtype), var_beta0
```
bendavid, line 1152: tf.zeros_like(var_beta0)?

bendavid:
Changes in the CI look as expected (postfit chi^2 increases, saturated chi^2 printed in some cases, no other changes)

PR #42, opened 2025-05-06
Add blinding for pois and nois

davidwalter2:
- Option `--unblind <list of parameters or regexs>` added to combinetf2_fit to unblind POIs and NOIs, which are now blinded by default when fitting the data (via -t 0).
- For each POI and NOI a deterministic seed is chosen based on the string of the name of the signal (in case of NOI) or systematic (in case of NOI). This ensures the same offset is used in different fit configurations.
- The random offset is chosen from a normal distribution of mean 0 and std 5
- The offset is then added on x when computing the event yields, as it was done for combinetf1. 
- In the fit, the offset is subtracted to ensure the same initial values as if the fit would not be blinded.

### DIFF
--- a/bin/combinetf2_fit.py
+++ b/bin/combinetf2_fit.py
@@ -599,7 +599,6 @@ def main():
         fit_time = []
         for ifit in fits:
             ifitter.defaultassign()
-            ifitter.set_blinding_offsets(False)
 
             group = "results"
             if ifit == -1:
--- a/bin/combinetf2_fit.py
+++ b/bin/combinetf2_fit.py
@@ -289,8 +289,10 @@ def make_parser():
     )
     parser.add_argument(
         "--unblind",
-        action="store_true",
-        help="Unblind parameters of interest",
+        type=str,
+        default=[],
+        nargs="*",
+        help="Specify list of parameters or regex (leading by 'r:') to unblind matching parameters of interest. E.g. 'r:.*' to unblind all.",
     )
 
     return parser.parse_args()
@@ -597,7 +599,7 @@ def main():
         fit_time = []
         for ifit in fits:
             ifitter.defaultassign()
-            ifitter.do_blinding = False
+            ifitter.set_blinding(False)
 
             group = "results"
             if ifit == -1:
@@ -626,7 +628,7 @@ def main():
                 save_hists(args, models, ifitter, ws, prefit=True)
             prefit_time.append(time.time())
 
-            ifitter.do_blinding = ifit == 0 and not args.unblind
+            ifitter.set_blinding(ifit == 0, args.unblind)
             fit(args, ifitter, ws, dofit=ifit >= 0)
             fit_time.append(time.time())
 
--- a/bin/combinetf2_fit.py
+++ b/bin/combinetf2_fit.py
@@ -287,6 +287,11 @@ def make_parser():
         type=float,
         help="Assumed prefit uncertainty for unconstrained nuisances",
     )
+    parser.add_argument(
+        "--unblind",
+        action="store_true",
+        help="Unblind parameters of interest",
+    )
 
     return parser.parse_args()
 
@@ -592,6 +597,7 @@ def main():
         fit_time = []
         for ifit in fits:
             ifitter.defaultassign()
+            ifitter.do_blinding = False
 
             group = "results"
             if ifit == -1:
@@ -599,6 +605,7 @@ def main():
                 ifitter.nobs.assign(ifitter.expected_yield())
             if ifit == 0:
                 ifitter.nobs.assign(ifitter.indata.data_obs)
+
             elif ifit >= 1:
                 group += f"_toy{ifit}"
                 ifitter.toyassign(
@@ -619,6 +626,7 @@ def main():
                 save_hists(args, models, ifitter, ws, prefit=True)
             prefit_time.append(time.time())
 
+            ifitter.do_blinding = ifit == 0 and not args.unblind
             fit(args, ifitter, ws, dofit=ifit >= 0)
             fit_time.append(time.time())
 
--- a/bin/combinetf2_fit.py
+++ b/bin/combinetf2_fit.py
@@ -599,7 +599,7 @@ def main():
         fit_time = []
         for ifit in fits:
             ifitter.defaultassign()
-            ifitter.set_blinding(False)
+            ifitter.set_blinding_offsets(False)
 
             group = "results"
             if ifit == -1:
@@ -628,7 +628,7 @@ def main():
                 save_hists(args, models, ifitter, ws, prefit=True)
             prefit_time.append(time.time())
 
-            ifitter.set_blinding(ifit == 0, args.unblind)
+            ifitter.set_blinding_offsets(ifit == 0)
             fit(args, ifitter, ws, dofit=ifit >= 0)
             fit_time.append(time.time())
 
--- a/combinetf2/fitter.py
+++ b/combinetf2/fitter.py
@@ -207,54 +207,81 @@ class Fitter:
 
         self.unblind_parameters = unblind_parameters
 
-    def apply_blinding_offsets(
-        self,
-        x,
-    ):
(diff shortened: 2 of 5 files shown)


bendavid:
This risks accidentally unblinding in two ways
1) If extra diagnostics are printed the user can implicitly see the unblinding offset via the starting values of the fit parameters.  That's why the fit starting values were allowed to be shifted in combinetf1

2) If cards are fed in using the asimov as the observed data (and then running with -t 0) the postfit value of the poi/noi will reveal the offset.  This was the purpose of is_dataobs_int in combinetf1 https://github.com/bendavid/HiggsAnalysis-CombinedLimit/blob/4529e97ef743bcbedea4b592f2ca10d94a040e03/scripts/combinetf.py#L365  (admittedly this is not the nicest way to deal with this, better solutions welcome)

combinetf2/fitter.py
```diff
@@ -179,6 +183,67 @@ def __init__(self, indata, options):
...
+        # random blinding with seed taken based on string of parameter name
+
+        # backup and restore the initial random generator state
+        state = np.random.get_state()
+
+        def deterministic_random_from_string(s):
+            # Hash the string
+            hash = hashlib.sha256(s).hexdigest()
+            # Convert first 8 hex digits (32 bits) to int
+            seed = int(hash[:8], 16)
```
bendavid, line 222: This should use np.random.SeedSequence, which can probably directly consume int(hash, 16).  This less elegant version was used because the older version of numpy used in the combinetf1 environment did not have this functionality.

davidwalter2:
> This risks accidentally unblinding in two ways
> 
> 1. If extra diagnostics are printed the user can implicitly see the unblinding offset via the starting values of the fit parameters.  That's why the fit starting values were allowed to be shifted in combinetf1
> 2. If cards are fed in using the asimov as the observed data (and then running with -t 0) the postfit value of the poi/noi will reveal the offset.  This was the purpose of is_dataobs_int in combinetf1 https://github.com/bendavid/HiggsAnalysis-CombinedLimit/blob/4529e97ef743bcbedea4b592f2ca10d94a040e03/scripts/combinetf.py#L365  (admittedly this is not the nicest way to deal with this, better solutions welcome)

1. OK I remove the starting value manipulation. I'm just a bit worried in case of extreme blinding offsets. For this reason I changed signal strength modifiers being offset as `offset = e^offset - 1` to avoid negative signal strengths.

2. OK I see the point of doing this now. But I still want different seeds for different parameters. Imagine one has 2 POIs and unblinds first one and then the other, and has different setups with just one POI etc. So I'm just changing the seed in case "is_dataobs_int" is True but still different seeds for different parameters.

combinetf2/fitter.py
```diff
@@ -179,6 +183,79 @@ def __init__(self, indata, options):
...
+            x_mult[i] = np.exp(value)
+
+        for i in self.indata.noigroupidxs:
+            param = self.indata.noigroups[i]
+            if param in self.unblind_parameters:
+                continue
+            value = deterministic_random_from_string(param)
+            x_add[i + self.npoi] = value
+
+        x = x * tf.constant(x_mult, dtype=self.indata.dtype)
```
bendavid, line 254: Should this have different treatment for the case where the poi's are squared?  (or perhaps be applied after the corresponding transformation?)

bendavid:
There's actually at least one place where the blinding offset is not applied to theta
https://github.com/WMass/combinetf2/blob/1a70e53d66ddfe41be2c1c89715a37e776656aaf/combinetf2/fitter.py#L1321

Might be better to reorganize the code so that poi's and theta are only computed in one place...

combinetf2/fitter.py
davidwalter2: Blinding is applied now after transformation in dedicated `get_poi()` function

davidwalter2:
> There's actually at least one place where the blinding offset is not applied to theta
> 
> https://github.com/WMass/combinetf2/blob/1a70e53d66ddfe41be2c1c89715a37e776656aaf/combinetf2/fitter.py#L1321
> 
> Might be better to reorganize the code so that poi's and theta are only computed in one place...

But I think we don't want to apply it here, otherwise it will reveal the offset. My understanding is that the offset is only applied when the yields are computed such that self.x converges against `real value - offset`

davidwalter2:
I think with the multiplicative offset for POIs, the relative impacts are the same (not the absolute). While for the additive offsets on NOIs the absolute impacts are the same (not the relative). I think this is a choice we have to make

bendavid:
There's a structural problem here:

The blinding is being computed with numpy/python but being called from within tf functions which means that it only executes once, but in a difficult to understand way.
(Also it is computationally wasteful to re-seed and generate the random numbers for each fit iteration for example)

Probably what makes sense is to precompute any random numbers which are needed and store them as tensors or variables as members of the fitter.  (maybe some clever trick can be used to make it difficult to accidentally print this value)

bendavid:
probably having the member start with double underscore is sufficient...

davidwalter2:
OK, done

combinetf2/fitter.py
```diff
@@ -179,6 +183,91 @@ def __init__(self, indata, options):
             and ((not self.binByBinStat) or self.binByBinStatType == "normal")
         )
 
+    def set_blinding(self, blind=True, unblind_parameter_expressions=[]):
+        self._blind = blind
```
bendavid, line 187: This variable is captured as a constant inside tf.functions but then is allowed to change after the fact, which will not produce the correct behaviour.

davidwalter2:
> > But I think we don't want to apply it here, otherwise it will reveal the offset. My understanding is that the offset is only applied when the yields are computed such that self.x converges against `real value - offset`
> 
> No the offset is still needed here, otherwise the constraint is not being applied to the right value.

Can we not assume there are no constraints on POIs and NOIs?

combinetf2/fitter.py
```diff
@@ -179,6 +183,91 @@ def __init__(self, indata, options):
...
+        # multiply offset to nois
+        offsets = np.zeros(self.indata.nsyst, dtype=np.float64)
+        for i in self.indata.noigroupidxs:
+            param = self.indata.systs[i]
+            if param in unblind_parameters:
+                continue
+            logger.debug(f"Blind parameter {param}")
+            value = deterministic_random_from_string(param)
+            offsets[i] = value
+        self._blinding_offsets_theta = tf.constant(offsets, dtype=self.indata.dtype)
```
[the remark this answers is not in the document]
davidwalter2: OK, blinding 'values' computed once in '__init__' and corresponding 'offsets' defined as tf.Variable now, to switch it on or off

bin/combinetf2_fit.py
```diff
@@ -592,13 +599,15 @@ def main():
         fit_time = []
         for ifit in fits:
             ifitter.defaultassign()
+            ifitter.set_blinding_offsets(False)
```
bendavid, line 602: This might make sense to just fold into defaultassign()

davidwalter2: done

bendavid:
There are some significant changes in the CI, I guess because the fits are run blinded.  Probably not desirable, but should also check that the differences make sense in this context.

bendavid:
Ok, do you want to update the CI to remove the blinding (and/or add an additional test for the blinding?)

bendavid:
There's still a small change in https://cmsmwbot.web.cern.ch/CombineTF2/PRValidation/ReferenceRuns/2025_05_12_f11bce5/pulls_and_global_impacts_sig_data_vs_pseudodata.png vs https://cmsmwbot.web.cern.ch/CombineTF2/PRValidation/PR42/2025_05_12/pulls_and_global_impacts_sig_data_vs_pseudodata.png

does it make sense?

davidwalter2:
Yes, for the pseudodata I only unblind the POI (sig) but not the NOI (signal_slope). Just to test dome different arguments
https://github.com/davidwalter2/combinetf2/blob/250506_blinding/.github/workflows/main.yml#L201

PR #45, opened 2025-05-30
Protection for nan loss

lucalavezzo:
- added a protection to the fitter in case the loss went to nan, else it would infinitely and aimlessly iterate
- removed the default physics model in the fit
- added option to plot pseudodata instead of data in combinetf2_plot_inputdata.py

### DIFF
--- a/combinetf2/fitter.py
+++ b/combinetf2/fitter.py
@@ -1046,11 +1046,6 @@ class Fitter:
             elif self.indata.systematic_type == "normal":
                 normcentral = norm * ernorm + logsnorm
                 nexpcentral = tf.reduce_sum(normcentral, axis=-1)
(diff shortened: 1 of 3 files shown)


combinetf2/fitter.py
```diff
@@ -954,6 +956,11 @@ def _compute_yields_noBBB(self, compute_norm=False, full=True):
             elif self.indata.systematic_type == "normal":
                 normcentral = norm * ernorm + logsnorm
```
davidwalter2, line 957: can normcentral also get negative?

lucalavezzo: presumably since nexpcentral is the sum of normcentral along its last dim, yes, but I didn't see this causing problems (maybe because it's not used in `_compute_nll_components`?)

PR #53, opened 2025-06-27
Add composite physics model

davidwalter2:


lucalavezzo:
Looks good, I tested the CompositeModel with the AngularCoefficients, seems to work well.

lucalavezzo (approved):

PR #59, opened 2025-07-25
option to customize uncertainty label

cippy:


cippy:
Sorry, by mistake I deleted other changes because of using an older version of the script, I'll fix them now

PR #68, opened 2025-09-26
Update impacts and pulls plot

davidwalter2:
Add information of what is plot in the pulls column and option to switch between actual nuisance parameter and nuisance parameter difference to prefit


davidwalter2:
CI of this PR:
https://cmsmwbot.web.cern.ch/Rabbit/PRValidation/PR68/2025_09_26/
Compared to reference:
https://cmsmwbot.web.cern.ch/Rabbit/PRValidation/ReferenceRuns/2025_09_26_d59d453/
Look as expected, I think it can be mergerd

PR #69, opened 2025-10-01
added protection against null bins in chi2 fit

havynanc:


bendavid:
This protection should be optional (and off by default) since it has some non-trivial statistical implications.

PR #129, opened 2026-04-09
Sparse mode performance, SparseHist input dispatch, and low-memory --noHessian mode

bendavid:
Adds support for sparse histogram input in the TensorWriter.

Significant performance optimizations for sparse mode in
both the TensorWriter and in the Fitter.

Performance optimizations should also give a factor of ~2
improvement for dense mode in the Fitter for large models.

Reworks `--noHessian` into a true low-memory mode: the dense
`[npar, npar]` covariance matrix is no longer allocated, the
Fitter initialization no longer has quadratic hot spots on large
external Hessians, and the postfit edmval and POI/NOI
uncertainties are still reported via Hessian-free conjugate
gradient solves. On the jpsi calibration tensor (108k params,
329M-nnz external sparse Hessian) `Fitter.__init__` drops from
~370 s to ~5 s and peak memory is reduced by ~94 GB.

Note that this depends on https://github.com/WMass/wums/pull/25

Three related groups of commits:

### Group A — TensorWriter sparse dispatch (8 commits)

* `add option to treat input systematic histograms as difference with
  respect to nominal`
* `add test for sparse mode`
* `Support scipy sparse array inputs in TensorWriter and add
  as_difference option`
* `Add multi-systematic dispatch in add_systematic and use
  wums.SparseHist`
* `Add external likelihood term (gradient + hessian) support`
* `Add efficient SparseHist multi-systematic dispatch in TensorWriter`
* `Speed up TensorWriter for large multi-systematic SparseHist
  workloads`
* **TensorWriter: use h.axes.extent for SparseHist flat-index
  unravel** — `SparseHist.shape` upstream now returns the no-flow
  shape (matching `hist.Hist.shape`), but the internal
  `_flat_indices` are encoded in the with-flow layout. The two
  `np.unravel_index` calls in `_sparse_per_syst_slices` and the
  batched `_sparse_multi_systematic` path now use `h.axes.extent`
  (matching `hist.NamedAxesTuple.extent`) instead of `h.shape`;
  using `h.shape` here would silently produce wrong coordinates
  under the new convention.

### Group B — Sparse fast path performance (5 commits)

Up to ~20× HVP speedup on the jpsi calibration tensor (76800 bins,
108334 params, 62M-nnz logk, 329M-nnz external sparse Hessian):
HVP 6380 → 320 ms, loss+grad 3010 → 160 ms.

* **inputdata, parsing: prep for sparse fast path with CSR matvec**
  — canonicalize sparse index ordering at load time, pre-build a
  `CSRSparseMatrix` view of `logk`, add `--hvpMethod` and
  `--noJitCompile` CLI options.

* **fitter: dynamic loss/grad/HVP wrappers with jit_compile +
  hvpMethod** — replace class-level `@tf.function` decorators with
  instance-level wrappers built dynamically in `_make_tf_functions`,
  so jit and HVP autodiff mode can be controlled per-fit. Note that
  fwdrev HVP is intentionally never jit-compiled because
  `tf.autodiff.ForwardAccumulator` does not propagate JVPs through
  XLA-compiled subgraphs.

* **fitter: sparse fast path uses CSR matmul, no dense
  [nbins, nproc]** — reformulate the sparse branch of
  `_compute_yields_noBBB` to use `tf_sparse_csr.matmul` for the
  inner contraction `logk @ theta` (~8× faster per call than the
  equivalent gather + segment_sum) and never materialize the dense
  `[nbins, nproc]` grid in the NLL/grad/HVP path. Also forces
  jit_compile=False in sparse mode (CSR matmul has no XLA kernel)
  and falls back to revrev when fwdrev is requested in sparse mode.

* **fitter: external sparse Hessian via CSR matmul** — switch the
  external sparse-Hessian likelihood term to use CSR matmul. The
  registered gradient of `sm.matmul` is itself a single
  `sm.matmul`, so reverse-over-reverse autodiff no longer
  rematerializes a 2D gather/scatter chain in the second-order
  tape. On the jpsi 329M-nnz prefit Hessian this was the dominant
  HVP cost.

* **rabbit_fit, setup.sh: enable XLA multi-threaded Eigen on CPU** —
  set `XLA_FLAGS=--xla_cpu_multi_thread_eigen=true` so XLA's CPU
  emitter uses Eigen's multi-threaded routines for the dense
  matmuls jit_compile=True generates. ~1.3× speedup on dense
  large-model HVP/loss+grad on a many-core system. Set both in
  `setup.sh` (for sourced shells) and at the very top of
  `bin/rabbit_fit.py` before any TF import (for direct invocation).

### Group C — `--noHessian` low-memory mode (5 commits)

Reworks `--noHessian` so it no longer allocates the dense
covariance matrix while still producing edmval and POI/NOI
uncertainties. The pipeline now runs end-to-end at O(npar)
memory for jpsi-scale problems where the full cov is infeasible
(~94 GB for 108k parameters in float64).

* **fitter, rabbit_fit: skip dense cov allocation under
  --noHessian** — split `prefit_covariance` into a vector form
  (`prefit_variance`) and a `tf.linalg.LinearOperatorDiag`
  wrapper; always allocate a length-`npar` `var_prefit` vector
  and only allocate the dense `self.cov` tf.Variable when the
  postfit Hessian will actually be computed. `defaultassign`,
  `randomize_parameters`, and `load_fitresult` all handle
  `self.cov is None`. The rabbit_fit CLI now explicitly rejects
  every flag that would need the postfit covariance
  (`--doImpacts`, `--computeVariations`, `--saveHists` without
  `--noChi2`, `--computeHistErrors[PerProcess]`,
  `--computeHistCov`, `--computeHistImpacts`,
  `--computeHistGaussianImpacts`, `--externalPostfit`).

* **fitter: speed up `Fitter.__init__` on large external sparse
  Hessians** — two structural fixes that reduce
  `Fitter.__init__` on the jpsi calibration tensor from ~370 s
  to ~20 s. Replace the per-parameter `np.where` lookup of
  external term parameter names (O(n²), ~150 s) with a single
  dict lookup. Detect already-canonical sparse-Hessian indices
  and skip `np.lexsort` (~54 s) when the input is already
  sorted.

* **unify sparse-Hessian IO path; sort at write time, drop
  reorder calls** — the TensorWriter now sorts the external
  sparse-Hessian indices into canonical row-major order at
  write time (matching `hlogk_sparse` / `hnorm_sparse`), so the
  reader and Fitter can use the same `makesparsetensor` helper
  and drop their defensive `tf.sparse.reorder` calls.
  `Fitter.__init__` drops further to ~5 s on jpsi.

* **fitter: Hessian-free CG solve for is_linear case under
  --noHessian** — the purely-quadratic `is_linear` fast path in
  `Fitter.minimize()` used to build the dense Hessian and do a
  Cholesky solve. Under `--noHessian` it now solves the normal
  equation `H @ dx = -grad` via `scipy.sparse.linalg.cg` with a
  `LinearOperator` backed by `loss_val_grad_hessp`, touching
  only O(npar) memory.

* **fitter, rabbit_fit: edmval + POI/NOI uncertainties under
  --noHessian** — compute edmval and the POI+NOI rows of the
  covariance matrix via Hessian-free CG solves (one for edmval,
  one per POI/NOI index). No dense Hessian or covariance is
  ever materialized. The POI+NOI diagonal entries populate the
  `parms_variances` vector passed to `add_parms_hist`; other
  nuisances keep NaN variances, signalling that their postfit
  uncertainty was not computed. Verified against the Cholesky
  path on the small test tensor: edmval and POI/NOI
  uncertainties match to full precision.

### DIFF
--- a/bin/rabbit_fit.py
+++ b/bin/rabbit_fit.py
@@ -428,6 +428,9 @@ def fit(args, fitter, ws, dofit=True):
             ws.add_1D_integer_hist(cb.loss_history, "epoch", "loss")
             ws.add_1D_integer_hist(cb.time_history, "epoch", "time")
 
+    # prefit variances as the default fallback for add_parms_hist below
+    parms_variances = None
+
     if not args.noHessian:
         # compute the covariance matrix and estimated distance to minimum
         _, grad, hess = fitter.loss_val_grad_hess()
@@ -467,6 +470,31 @@ def fit(args, fitter, ws, dofit=True):
                 global_impacts=True,
             )
 
+        parms_variances = tf.linalg.diag_part(fitter.cov)
+    else:
+        # --noHessian: avoid the full dense Hessian. Still compute edmval
+        # and the POI+NOI uncertainties via a Hessian-free conjugate
+        # gradient solve of H @ v = grad and H @ c_i = e_i, using only
+        # Hessian-vector products. The CG solves touch O(npar) memory
+        # per call instead of O(npar^2), so this works on problems
+        # where the full covariance would be infeasible.
+        _, grad = fitter.loss_val_grad()
+        npoi = int(fitter.poi_model.npoi)
+        noi_idx_in_x = np.asarray(fitter.indata.noiidxs, dtype=np.int64) + npoi
+        poi_noi_idx = np.concatenate([np.arange(npoi, dtype=np.int64), noi_idx_in_x])
+        edmval, cov_rows = fitter.edmval_cov_rows_hessfree(grad, poi_noi_idx)
+        logger.info(f"edmval: {edmval}")
+
+        # Build a full-length variance vector with the POI+NOI entries
+        # populated from the diagonal of the CG-solved rows and the rest
+        # left as NaN (we did not compute those). add_parms_hist stores
+        # the vector verbatim into the workspace.
+        n = int(fitter.x.shape[0])
+        parms_variances_np = np.full(n, np.nan, dtype=np.float64)
+        for k, i in enumerate(poi_noi_idx):
+            parms_variances_np[int(i)] = cov_rows[k, int(i)]
+        parms_variances = tf.constant(parms_variances_np, dtype=fitter.indata.dtype)
+
     nllvalreduced = fitter.reduced_nll().numpy()
 
     ndfsat = (
@@ -497,7 +525,7 @@ def fit(args, fitter, ws, dofit=True):
 
     ws.add_parms_hist(
         values=fitter.x,
-        variances=tf.linalg.diag_part(fitter.cov) if not args.noHessian else None,
+        variances=parms_variances,
         hist_name="parms",
     )
 
--- a/bin/rabbit_fit.py
+++ b/bin/rabbit_fit.py
@@ -363,9 +363,16 @@ def save_hists(args, mappings, fitter, ws, prefit=True, profile=False):
             )
 
         if args.computeVariations:
+            if fitter.cov is None:
+                raise RuntimeError(
+                    "--computeVariations requires the parameter covariance "
+                    "matrix and so is incompatible with --noHessian."
+                )
             if prefit:
                 cov_prefit = fitter.cov.numpy()
-                fitter.cov.assign(fitter.prefit_covariance(unconstrained_err=1.0))
+                fitter.cov.assign(
+                    fitter.prefit_covariance(unconstrained_err=1.0).to_dense()
+                )
 
             exp, aux = fitter.expected_events(
                 mapping,
@@ -574,8 +581,31 @@ def main():
     if args.eager:
         tf.config.run_functions_eagerly(True)
 
-    if args.noHessian and args.doImpacts:
-        raise Exception('option "--noHessian" only works without "--doImpacts"')
+    # --noHessian skips computing the postfit Hessian, so the dense
+    # parameter covariance matrix is never available. Any feature that
+    # needs the covariance is incompatible.
+    if args.noHessian:
+        _incompat = []
+        if args.doImpacts:
+            _incompat.append("--doImpacts")
+        if args.computeVariations:
+            _incompat.append("--computeVariations")
+        if args.saveHists and not args.noChi2:
+            _incompat.append("--saveHists (without --noChi2)")
+        if args.computeHistErrors:
+            _incompat.append("--computeHistErrors")
+        if args.computeHistErrorsPerProcess:
+            _incompat.append("--computeHistErrorsPerProcess")
+        if args.computeHistCov:
+            _incompat.append("--computeHistCov")
+        if args.computeHistImpacts:
+            _incompat.append("--computeHistImpacts")
+        if args.computeHistGaussianImpacts:
+            _incompat.append("--computeHistGaussianImpacts")
+        if args.externalPostfit is not None:
+            _incompat.append("--externalPostfit")
+        if _incompat:
+            raise Exception("--noHessian is incompatible with: " + ", ".join(_incompat))
 
     global logger
     logger = logging.setup_logger(__file__, args.verbose, args.noColorLogger)
@@ -693,7 +723,7 @@ def main():
 
                 ws.add_parms_hist(
                     values=ifitter.x,
-                    variances=tf.linalg.diag_part(ifitter.cov),
+                    variances=ifitter.var_prefit,
                     hist_name="parms_prefit",
                 )
 
--- a/bin/rabbit_fit.py
+++ b/bin/rabbit_fit.py
@@ -1,5 +1,19 @@
 #!/usr/bin/env python3
 
+# Enable XLA's multi-threaded Eigen path on CPU before importing tensorflow.
+# This must be set before any TF import (including transitive) because XLA
+# parses XLA_FLAGS once during runtime initialization. Measured ~1.3x speedup
+# on dense large-model HVP/loss+grad on a many-core system, no downside.
+# Users who set their own XLA_FLAGS keep theirs and we append.
+import os as _os
+
+_xla_default = "--xla_cpu_multi_thread_eigen=true"
+_existing = _os.environ.get("XLA_FLAGS", "")
+if "xla_cpu_multi_thread_eigen" not in _existing:
+    _os.environ["XLA_FLAGS"] = (
+        f"{_existing} {_xla_default}".strip() if _existing else _xla_default
+    )
+
 import copy
 
 import tensorflow as tf
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -318,6 +318,52 @@ class Fitter:
         # one common regularization strength parameter
         self.tau = tf.Variable(1.0, trainable=True, name="tau", dtype=tf.float64)
 
+        # External likelihood terms (additive g^T x + 0.5 x^T H x contributions
+        # to the NLL). Resolve parameter name strings against the full fit
+        # parameter list (POIs + systs).
+        self.external_terms = []
+        parms_str = self.parms.astype(str)
+        for term in self.indata.external_terms:
+            params = np.asarray(term["params"]).astype(str)
+            indices = np.empty(len(params), dtype=np.int64)
+            for i, p in enumerate(params):
+                matches = np.where(parms_str == p)[0]
+                if len(matches) != 1:
+                    raise RuntimeError(
+                        f"External likelihood term '{term['name']}' parameter "
+                        f"'{p}' matched {len(matches)} entries in fit parameters"
+                    )
+                indices[i] = matches[0]
+            tf_indices = tf.constant(indices, dtype=tf.int64)
+
+            tf_grad = (
+                tf.constant(term["grad_values"], dtype=self.indata.dtype)
+                if term["grad_values"] is not None
+                else None
+            )
+
+            tf_hess_dense = None
+            tf_hess_sparse = None
+            if term["hess_dense"] is not None:
+                tf_hess_dense = tf.constant(term["hess_dense"], dtype=self.indata.dtype)
+            elif term["hess_sparse"] is not None:
+                rows, cols, vals = term["hess_sparse"]
+                tf_hess_sparse = (
+                    tf.constant(rows, dtype=tf.int64),
+                    tf.constant(cols, dtype=tf.int64),
+                    tf.constant(vals, dtype=self.indata.dtype),
+                )
+
+            self.external_terms.append(
+                {
+                    "name": term["name"],
+                    "indices": tf_indices,
+                    "grad": tf_grad,
+                    "hess_dense": tf_hess_dense,
+                    "hess_sparse": tf_hess_sparse,
+                }
+            )
+
         # constraint minima for nuisance parameters
         self.theta0 = tf.Variable(
             self.theta0default,
@@ -2087,6 +2133,27 @@ class Fitter:
 
         return ln, lc, lbeta, lpenalty, beta
 
+    def _compute_external_nll(self):
+        """Sum of external likelihood term contributions: sum_i (g_i^T x_sub + 0.5 x_sub^T H_i x_sub)."""
+        if not self.external_terms:
+            return None
+        total = tf.zeros([], dtype=self.indata.dtype)
+        for term in self.external_terms:
+            x_sub = tf.gather(self.x, term["indices"])
+            if term["grad"] is not None:
+                total = total + tf.reduce_sum(term["grad"] * x_sub)
+            if term["hess_dense"] is not None:
+                # 0.5 * x_sub^T H x_sub
+                total = total + 0.5 * tf.reduce_sum(
+                    x_sub * tf.linalg.matvec(term["hess_dense"], x_sub)
+                )
+            elif term["hess_sparse"] is not None:
+                rows, cols, vals = term["hess_sparse"]
+                total = total + 0.5 * tf.reduce_sum(
+                    vals * tf.gather(x_sub, rows) * tf.gather(x_sub, cols)
+                )
+        return total
+
     def _compute_nll(self, profile=True, full_nll=False):
         ln, lc, lbeta, lpenalty, beta = self._compute_nll_components(
             profile=profile, full_nll=full_nll
@@ -2098,6 +2165,10 @@ class Fitter:
 
         if lpenalty is not None:
             l = l + lpenalty
+
+        lext = self._compute_external_nll()
+        if lext is not None:
+            l = l + lext
         return l
 
     def _compute_loss(self, profile=True):
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -965,6 +965,72 @@ class Fitter:
         else:
             return edmval_cov(grad, hess)
 
+    def edmval_cov_rows_hessfree(self, grad, row_indices, rtol=1e-10, maxiter=None):
+        """Hessian-free edmval + selected rows of the covariance matrix.
+
+        Used under --noHessian to avoid allocating the dense [npar, npar]
+        Hessian. Solves the linear systems
+
+            H v = grad        ->  edmval = 0.5 * grad^T v
+            H c_i = e_i       ->  c_i is the i-th column/row of cov
+
+        iteratively via scipy's conjugate gradient, feeding it a
+        LinearOperator backed by self.loss_val_grad_hessp. The Hessian
+        must be positive-definite; that's the case for a converged NLL
+        minimum (including the purely-quadratic --is_linear case).
+
+        Parameters
+        ----------
+        grad : tf.Tensor or array-like, shape [npar]
+            Gradient at the current x, already computed by the caller.
+        row_indices : iterable of int
+            Parameter indices to compute covariance rows for. Typically
+            the POI indices [0, npoi) concatenated with the NOI indices
+            (npoi + noiidxs).
+        rtol : float
+            Relative residual tolerance passed to scipy.sparse.linalg.cg.
+        maxiter : int or None
+            Maximum CG iterations per solve; None lets scipy choose.
+
+        Returns
+        -------
+        edmval : float
+        cov_rows : np.ndarray, shape [len(row_indices), npar]
+            Row i is (H^{-1})[row_indices[i], :]; diag entries give the
+            variances for those parameters.
+        """
+        import scipy.sparse.linalg as _spla
+
+        n = int(self.x.shape[0])
+        dtype = np.float64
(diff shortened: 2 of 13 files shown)


rabbit/fitter.py
```diff
@@ -318,6 +321,66 @@ def init_fit_parms(
         # one common regularization strength parameter
         self.tau = tf.Variable(1.0, trainable=True, name="tau", dtype=tf.float64)
 
+        # External likelihood terms (additive g^T x + 0.5 x^T H x contributions
```
davidwalter2: Suggest to put this block into a standalone function that takes external_terms and dtype as argument and returns "external_terms" object. Suggest to put it in a new file e.g. external_likelihood.py

rabbit/h5pyutils_read.py
```diff
@@ -1,3 +1,4 @@
+import hdf5plugin  # noqa: F401  registers Blosc2/LZ4 filter used by the writer
```
davidwalter2, line 1: Why is this needed? I don't understand the comment

rabbit/parsing.py
```diff
@@ -202,6 +202,24 @@ def common_parser():
...
+        default="revrev",
+        type=str,
+        choices=["fwdrev", "revrev"],
+        help="Autodiff mode for the Hessian-vector product. 'revrev' (reverse-over-reverse) "
+        "is the default and works well in combination with --jitCompile. 'fwdrev' "
+        "(forward-over-reverse, via tf.autodiff.ForwardAccumulator) is an alternative.",
+    )
+    parser.add_argument(
+        "--noJitCompile",
+        dest="jitCompile",
```
davidwalter2: Using the dest keyword can be confusing and so far we managed to do without it, could we keep that convention?

rabbit/fitter.py
```diff
@@ -2098,47 +2242,84 @@ def _compute_nll(self, profile=True, full_nll=False):
...
+
+    def _make_tf_functions(self):
+        # Build tf.function wrappers at instance construction time so that
+        # jit_compile and the HVP autodiff mode can be controlled via fit
+        # options without redefining the class.
+        #
+        # SparseMatrixMatMul has no XLA kernel, so any tf.function that
+        # uses it (via _compute_yields_noBBB in sparse mode) cannot be
+        # jit-compiled. Force jit_compile off in sparse mode regardless
+        # of the user's --jitCompile setting.
```
davidwalter2: There is no "--jitCompile" setting but the option is "--noJitCompile", maybe got confused by the dest keyword?

rabbit/fitter.py
```diff
@@ -2098,47 +2242,84 @@ def _compute_nll(self, profile=True, full_nll=False):
...
+        jit = self.jit_compile and not self.indata.sparse
```
davidwalter2: Suggest to add a warning with something like"
"jit_compile set but input data is sparse and jit compilation will be disabled."

rabbit/fitter.py
```diff
@@ -2087,6 +2198,39 @@ def _compute_nll_components(self, profile=True, full_nll=False):
 
         return ln, lc, lbeta, lpenalty, beta
 
+    def _compute_external_nll(self):
```
davidwalter2, line 2309: This could also go in a standalone function, which would be more modular and reusable which is the direction I think we should go. Arguments would be terms, params.

rabbit/parsing.py
```diff
@@ -202,6 +202,24 @@ def common_parser():
...
+        "--hvpMethod",
+        default="revrev",
+        type=str,
+        choices=["fwdrev", "revrev"],
+        help="Autodiff mode for the Hessian-vector product. 'revrev' (reverse-over-reverse) "
+        "is the default and works well in combination with --jitCompile. 'fwdrev' "
+        "(forward-over-reverse, via tf.autodiff.ForwardAccumulator) is an alternative.",
+    )
+    parser.add_argument(
+        "--noJitCompile",
```
davidwalter2: How does the option interplay with --eager? Do we need both options or can this also not be controlled with --eager? At a minimum --eager should also trigger no jit compile

rabbit/tensorwriter.py
```diff
@@ -1014,6 +2152,51 @@ def create_dataset(
                 )
                 beta_variations = None
 
+        # Write external likelihood terms
+        if self.external_terms:
+            ext_group = f.create_group("external_terms")
+            create_dataset(
```
davidwalter2: are the names needed? They are already stored as the keys of the ext_group groups added below

rabbit/inputdata.py
```diff
@@ -182,6 +196,59 @@ def __init__(self, filename, pseudodata=None):
...
+            #   hess_dense: 2D float ndarray or None
+            #   hess_sparse: tuple (rows, cols, values) or None
+            self.external_terms = []
+            if "external_terms" in f.keys():
+                names = [
+                    s.decode() if isinstance(s, bytes) else s
+                    for s in f["hexternal_term_names"][...]
+                ]
+                ext_group = f["external_terms"]
+                for tname in names:
```
davidwalter2: Does the following work?
```
for tname, tg, in ext_group.items()
```
It would be more pythonic IMO and no need for storing the names separately

rabbit/inputdata.py
```diff
@@ -182,6 +196,59 @@ def __init__(self, filename, pseudodata=None):
 
             self.axis_procs = hist.axis.StrCategory(self.procs, name="processes")
 
+            # Load external likelihood terms (optional).
```
davidwalter2: Could also maybe go into a standalone function, but not sure about that

davidwalter2 (commented):
First bunch of comments

rabbit/h5pyutils_read.py
bendavid: The compression type is explicitly set when writing, but it's implicitly set when reading.  If the package isn't loaded then the underlying decompressor is not guaranteed to be available in the C code and reading could fail.  (Of course it might still work if the package was imported elsewhere, but putting it here explicitly ensures it won't break from unrelated changes later)

rabbit/fitter.py
bendavid: Almost certainly because it was --jitCompile before and then got swapped when the default changed.

rabbit/fitter.py
bendavid: Since this is the default it probably doesn't make sense to have a warning.  Might make sense to have a default Non/ "auto" option for jitCompile rather than just on/off

rabbit/parsing.py
bendavid: if --eager is used I think it will implicitly skip jitCompile because it won't even process the tf;functions

These are two different things though 
tf function by itself switches from eager to graph mode.  jit_compile recompiles the instructions within the graph (allowing things like add-multiply fusion like what a C++ compiler would do).  Otherwise the graph internally still executes the tf operations one by one as written.

So yes we need both options and I think the current behaviour should be fine.

rabbit/tensorwriter.py
bendavid: Not sure...

rabbit/inputdata.py
bendavid: Yes I think that should work

rabbit/tensorwriter.py
bendavid: Oh I see what you mean now.   Yes probably they don't need to be stored seperately indeed.

rabbit/inputdata.py
bendavid: Yah probably.  The __init__ of this is already not very modular so maybe can leave that for subsequent refactoring.

rabbit/parsing.py
bendavid: Yah ok let me see if that can be avoided

rabbit/fitter.py
bendavid: Ok yes the makes sense actually

rabbit/inputdata.py
bendavid: Ok actually given the similar pattern in the fitter init I agree this can be modularized.

rabbit/fitter.py
bendavid: Done in fad47bc -- the external-term construction now lives in a new `rabbit/external_likelihood.py` module, and `Fitter.__init__` calls `external_likelihood.build_tf_external_terms(self.indata.external_terms, self.parms, self.indata.dtype)`. The matching scalar evaluator and the h5 reader live in the same module.

rabbit/h5pyutils_read.py
bendavid: (Already explained in https://github.com/WMass/rabbit/pull/129#discussion_r3060723867 -- this preloads the Blosc2/LZ4 decompressor before any read so it's guaranteed to be available, regardless of whether some other module happens to import hdf5plugin first.)

rabbit/parsing.py
bendavid: Done in fad47bc -- replaced `--noJitCompile` with a tri-state `--jitCompile {auto,on,off}` (default `auto`), no `dest` keyword needed.

rabbit/fitter.py
bendavid: Done in fad47bc -- the option is now `--jitCompile auto|on|off` and `Fitter.__init__` reads it as a string. Backwards-compatible with True/False from programmatic callers.

rabbit/fitter.py
bendavid: Done in fad47bc -- the new tri-state `--jitCompile` defaults to `auto`, which silently disables jit in sparse mode. The warning fires only when the user explicitly passes `--jitCompile on` while running on sparse input, since at that point they've asked for something XLA can't actually do.

rabbit/fitter.py
bendavid: Done in fad47bc -- moved into `rabbit.external_likelihood.compute_external_nll(terms, x, dtype)`. `Fitter._compute_external_nll` is now a one-line dispatch.

rabbit/parsing.py
bendavid: No code change here. `--eager` and `--jitCompile` control different things (eager mode skips graph building entirely, jit_compile is XLA fusion within an existing graph) so they remain orthogonal. `--eager` continues to bypass jit by virtue of skipping the `tf.function` wrappers.

rabbit/tensorwriter.py
bendavid: Done in fad47bc -- the writer no longer creates `hexternal_term_names`; the reader now iterates the `external_terms` h5 group's subgroups directly via `ext_group.items()`.

rabbit/inputdata.py
bendavid: Done in fad47bc -- the reader now uses `for tname, tg in ext_group.items()` (in `rabbit.external_likelihood.read_external_terms_from_h5`) and the writer-side names list is gone.

rabbit/inputdata.py
bendavid: Done in fad47bc -- moved into `rabbit.external_likelihood.read_external_terms_from_h5(ext_group)`. `FitInputData` calls `self.external_terms = read_external_terms_from_h5(f.get("external_terms"))`.

bin/rabbit_fit.py
```diff
@@ -407,6 +428,9 @@ def fit(args, fitter, ws, dofit=True):
             ws.add_1D_integer_hist(cb.loss_history, "epoch", "loss")
             ws.add_1D_integer_hist(cb.time_history, "epoch", "time")
 
+    # prefit variances as the default fallback for add_parms_hist below
+    parms_variances = None
```
davidwalter2, line 432: I don't see why this is needed, it follows an if ... else statement in the code where parms_variances is defined either way

rabbit/fitter.py
```diff
@@ -121,6 +122,27 @@ def __init__(
...
+        _jit_opt = getattr(options, "jitCompile", "auto")
+        if _jit_opt is True:
+            _jit_opt = "on"
+        elif _jit_opt is False:
+            _jit_opt = "off"
+        if _jit_opt not in ("auto", "on", "off"):
+            raise ValueError(
+                f"jitCompile must be one of 'auto', 'on', 'off'; got {_jit_opt!r}"
+            )
+        self.jit_compile = _jit_opt
```
davidwalter2: Can jit_compile be a boolean and set to True if _jit_opt == "on" and False otherwise?

rabbit/fitter.py
```diff
@@ -2098,47 +2296,104 @@ def _compute_nll(self, profile=True, full_nll=False):
...
+    def _make_tf_functions(self):
+        # Build tf.function wrappers at instance construction time so that
+        # jit_compile and the HVP autodiff mode can be controlled via fit
+        # options without redefining the class.
+        #
+        # SparseMatrixMatMul has no XLA kernel, so any tf.function that
+        # uses it (via _compute_yields_noBBB in sparse mode) cannot be
+        # jit-compiled. Resolve the tri-state self.jit_compile setting:
+        #
+        #   "auto" -> enable jit in dense mode, silently disable in
```
davidwalter2: can the "auto" be resolved at fitter init stage?

rabbit/inputdata.py
```diff
@@ -82,11 +82,22 @@ def __init__(self, filename, pseudodata=None):
...
                 self.norm = makesparsetensor(f["hnorm_sparse"])
                 self.logk = makesparsetensor(f["hlogk_sparse"])
+                # Pre-build a CSRSparseMatrix view of logk for use in the
+                # fitter's sparse matvec path via sm.matmul, which dispatches
+                # to a multi-threaded CSR kernel and is much faster per call
+                # than the equivalent gather + unsorted_segment_sum. NOTE:
+                # SparseMatrixMatMul has no XLA kernel, so any tf.function
+                # that calls sm.matmul must be built with jit_compile=False.
+                from tensorflow.python.ops.linalg.sparse import (
+                    sparse_csr_matrix_ops as _tf_sparse_csr,
```
davidwalter2: in the fitter.py it is imported as tf_sparse_csr without the leading "_", I suggest to keep a consistent convention

rabbit/tensorwriter.py
```diff
@@ -334,6 +488,639 @@ def add_norm_systematic(
...
+        logkavg_proc = (nnz_idx, logkavg_vals, size)
+        self.book_logk_avg(logkavg_proc, channel, process, var_name_out)
+        self.book_systematic(
+            var_name_out, add_to_data_covariance=add_to_data_covariance, **kargs
+        )
+
+    @staticmethod
+    def _bin_label(ax, idx):
+        """Return a string label for a hist axis bin, preferring string values."""
+        try:
```
davidwalter2: why try except? In what cases would this be triggered and do we want to continue or error out better?

tests/test_external_term.py
davidwalter2, line 1: Suggest to add this test in the github CI

tests/test_multi_systematic.py
davidwalter2, line 1: Same here, can it be added to the github CI?

tests/test_sparse_fit.py
davidwalter2, line 1: Can this also be added to the github CI? To what degree is it already covered or orthogonal by the sparse tensor fit that is already in the github CI?

bin/rabbit_fit.py
bendavid: I think this is a coding convention thing to maintain C-like variable scope semantics.

rabbit/fitter.py
bendavid: Addressed in 54530e2 — `self.jit_compile` is now resolved to a plain `bool` right after the tri-state parsing in `Fitter.__init__`, using `self.indata.sparse` (the only runtime condition it can depend on). `_make_tf_functions` just does `jit = self.jit_compile`.

*— Claude (Opus 4.7, acting on @bendavid's behalf)*

rabbit/fitter.py
bendavid: Yes — done in 54530e2, same commit as the other jitCompile comment. The "auto"/"on"/"off" → `bool` collapse now happens once in `Fitter.__init__` (right after the tri-state is parsed), and `_make_tf_functions` just reads the resolved boolean. The warning for `--jitCompile=on` + sparse mode now fires at init time rather than being deferred to `_make_tf_functions`, which is arguably more useful (earlier feedback).

*— Claude (Opus 4.7, acting on @bendavid's behalf)*

rabbit/inputdata.py
bendavid: Done in a43fc47 — renamed to `tf_sparse_csr` to match the existing `rabbit/fitter.py` import.

*— Claude (Opus 4.7, acting on @bendavid's behalf)*

rabbit/tensorwriter.py
bendavid: Removed in 21babae. Looking at it again, every stock `hist.axis` type (Regular / Integer / Variable / StrCategory / IntCategory) implements `.value(idx)` and returns either a scalar or a string for any in-range index, so there's no realistic failure mode that needed the blanket `except Exception`. The body is now the unconditional `ax.value(idx)` + `isinstance` check — if a bug ever does produce a bad index we'd rather see the traceback than silently get `str(idx)`.

*— Claude (Opus 4.7, acting on @bendavid's behalf)*

tests/test_external_term.py
bendavid: Added in f2f8b07. The new `unit-tests` CI job runs `test_external_term.py`, `test_multi_systematic.py`, and `test_sparse_fit.py` in parallel after `setenv` (matrix of three). Each test is self-contained — it builds its own temp tensor, so there's no dependency on the `make-tensor` stage.

*— Claude (Opus 4.7, acting on @bendavid's behalf)*

tests/test_multi_systematic.py
bendavid: Added in f2f8b07 together with the other two tests — see reply on `tests/test_external_term.py`.

*— Claude (Opus 4.7, acting on @bendavid's behalf)*

tests/test_sparse_fit.py
bendavid: Added in f2f8b07 together with the other two tests. On overlap with the existing `sparse-fits` job: there's some, but the two are largely orthogonal. `sparse-fits` runs the main `rabbit_fit.py` CLI against a pre-made sparse tensor and exercises impacts/variations/plots — an end-to-end check that the full pipeline works. `tests/test_sparse_fit.py` instead builds a small tensor in **both** dense and sparse modes and asserts that the fit results match element-wise, so it's a regression-style consistency check specifically on the sparse code path. Keeping both makes sense: the existing job catches pipeline-level breakage, the new one catches sparse/dense divergence.

*— Claude (Opus 4.7, acting on @bendavid's behalf)*

PR #133, opened 2026-04-30
asymmetric traditional impacts; unblinding groups

lucalavezzo:


bin/rabbit_fit.py
```diff
@@ -226,6 +226,107 @@ def make_parser():
...
+        nargs="+",
+        help="Regex(es) restricting which nuisances are scanned for --asymImpacts.",
+    )
+    parser.add_argument(
+        "--asymImpactsExclude",
+        default=None,
+        nargs="+",
+        help="Regex(es) excluding nuisances from --asymImpacts.",
+    )
+    parser.add_argument(
```
davidwalter2, line 249: I would remove that option and make it default to compute all nuisances with `asymImpacts`, then exclude the ones you like with `--asymImpactsExclude` or specify a list with the `--asymImpactsInclude` otherwise we have too many CLAs

bin/rabbit_fit.py
```diff
@@ -226,6 +226,107 @@ def make_parser():
...
+        "'exact' rebuilds the full N x N NLL Hessian each iteration (slow, "
+        "reference). 'hvp' uses Hessian-vector products via a LinearOperator "
+        "(exact, typically fastest for large N). 'frozen' uses cov^-1 from "
+        "the postfit as a constant Hessian (cheapest but produces silent "
+        "failures on non-Gaussian profiles -- speed reference only). "
+        "'bfgs'/'sr1' use a quasi-Newton estimate built up from the gradient "
+        "sequence (no extra TF calls per iteration; may need more iterations).",
+    )
+    parser.add_argument(
+        "--asymImpactsMaxiter",
```
davidwalter2, line 270: I think we should remove that option. The user should use `--earlyStopping x` instead, which says if the loss doesn't improve after `x` iterations it will stop. This makes sure to not get on nuisances that don't converge while it still keeps it running until the end.

bin/rabbit_fit.py
```diff
@@ -358,6 +459,37 @@ def save_hists(args, mappings, fitter, ws, prefit=True, profile=False):
                 logger.info(f"    2*deltaNLL: {round(chi2val, 2)}")
                 logger.info(rf"    p-value: {round(p_val * 100, 2)}%")
 
+                # DEBUG: decompose chi2_sat = 2*(NLL_nom - NLL_sat)
```
davidwalter2, line 462: Is this still needed or leftover from some debugging and can be removed?

bin/rabbit_fit.py
```diff
@@ -434,24 +566,41 @@ def fit(args, fitter, ws, dofit=True):
...
             logger.info(f"profile beta")
             fitter._profile_beta()
+            logger.debug(
+                f"[timing] _profile_beta(): {time.perf_counter() - _t_bb:.1f} s"
+            )
 
         if cb is not None:
             ws.add_1D_integer_hist(cb.loss_history, "epoch", "loss")
             ws.add_1D_integer_hist(cb.time_history, "epoch", "time")
+            try:
```
davidwalter2, line 589: Is a try-except statement here needed? I don't think so

bin/rabbit_fit.py
```diff
@@ -434,24 +566,41 @@ def fit(args, fitter, ws, dofit=True):
...
+                n_iter = len(cb.loss_history)
+                logger.debug(f"[timing] L-BFGS: {n_iter} epochs recorded")
```
davidwalter2, line 591: I think this is the minimizer iterations, which doesn't have to be L-BFGS

bin/rabbit_fit.py
```diff
@@ -434,24 +566,41 @@ def fit(args, fitter, ws, dofit=True):
...
     parms_variances = None
 
-    # take covariance from externalPostfit in case the fit was skipped
-    if dofit or args.externalPostfit is None:
+    # take covariance from externalPostfit in case the fit was skipped —
+    # unless a Hessian is explicitly wanted (no --noHessian): the two-pass
+    # straight-through covariance recipe (--externalPostfit ... --noFit,
+    # WITHOUT --noHessian) relies on recomputing the Hessian at the loaded
+    # postfit point.
+    if dofit or args.externalPostfit is None or not args.noHessian:
```
davidwalter2, line 603: I think the ` or not args.noHessian` should be removed here because if you run with `--externalPostfit <fitresult> --noFit` you expect the covariance to be taken from the fitresult

bin/rabbit_fit.py
```diff
@@ -520,6 +682,22 @@ def fit(args, fitter, ws, dofit=True):
 
     nllvalreduced = fitter.reduced_nll().numpy()
 
+    # DEBUG: decompose nominal-fit NLL into (Poisson data, nuisance constraints, BBB)
```
davidwalter2, line 685: Do we want to print that always or is this something left from testing that can be removed?

bin/rabbit_fit.py
```diff
@@ -714,6 +921,17 @@ def main():
         "nois": ifitter.parms[ifitter.param_model.nparams :][indata.noiidxs],
     }
 
+    # ParamModel Gaussian priors (if --paramModelPriors was used and the
+    # model declared sigmas). Stored as a small dict so downstream tooling
+    # can see what was applied without parsing the rabbit log.
+    if getattr(ifitter, "param_prior_active", False):
+        meta["param_priors"] = {
+            "params": ifitter.param_model.params,  # all nparams names
+            "mask": ifitter.param_prior_mask_np,  # bool array
```
davidwalter2, line 930: What does `np` stand for? Nuisance Parameter? It could be that one want's to declare a parameter as POI and still want to add a prior to it, so this is unnecessarily specific and can be removed

bin/rabbit_fit.py
```diff
@@ -748,6 +966,19 @@ def main():
             for j, (data_values, data_variances) in enumerate(datasets):
 
                 ifitter.defaultassign()
+                # If --externalPostfit was supplied AND this is an
```
davidwalter2, line 969: Should we not expect the user to run with `-t 0`? Then we don't need that

rabbit/impacts/asym_impacts.py
```diff
@@ -0,0 +1,168 @@
+"""
+Traditional asymmetric impacts.
+
+For each selected nuisance, find the asymmetric +/- 1 sigma points on the
+Delta(2NLL)=q likelihood contour via constrained minimization (contour_scan).
+The shifts of every fitter parameter at those points are the asymmetric
+impacts. Group impacts are obtained by quadrature envelope of the contained
```
davidwalter2, line 7: I think the convention for grouped impacts in the case of traditional impacts is different, we should use the analogue to what is done in the gaussian case

rabbit/impacts/asym_impacts.py
```diff
@@ -0,0 +1,168 @@
...
+nuisances, separately for the down and up sides.
+
+Nuisances that are structurally symmetric (logkhalfdiff identically zero)
```
davidwalter2, line 10: I don't think this is true because nonlinear effects can lead to asymmetric constraints and impacts also for symmetric nuisances. I think by default we should do the most generic and compute for all nuisances and the user has to specify explicitly which he wants to include/exclude

rabbit/impacts/asym_impacts.py
```diff
@@ -0,0 +1,168 @@
...
+
+def asym_impacts_parms(
+    fitter,
+    nll_min,
+    selected_idxs,
+    selected_names,
+    q=1,
+    contour_xtol=1e-4,
+    contour_gtol=1e-4,
+    contour_maxiter=200,
```
davidwalter2, line 77: I think the default should be set to effectively infinite, 5000 or so, since the user does not expect it to stop early

rabbit/impacts/traditional_impacts.py
```diff
@@ -16,58 +16,76 @@ def _compute_impact_group(cov, v, idxs, nsignal_params=0):
     return tf.sqrt(v_invC_v)
 
 
-def _gather_poi_noi_vector(v, noiidxs, nsignal_params=0):
+def _gather_poi_noi_vector(v, noiidxs, nsignal_params=0, nmodel_params=None):
```
davidwalter2, line 19: Is this computing the impacts also for POUs? If so, why?

rabbit/parsing.py
```diff
@@ -244,7 +244,44 @@ def common_parser():
         "--noEDM",
         default=False,
         action="store_true",
-        help="Don't compute the estimated distance to minimum as fit quality evaluation",
+        help="Skip the Hessian-free EDM/CG postfit step (only meaningful with "
+        "--noHessian). Skips both the edmval estimate and the POI+NOI "
+        "uncertainty rows; their entries in the output are NaN.",
+    )
+    parser.add_argument(
+        "--paramModelPriors",
```
davidwalter2, line 252: I think there should be no CLA for that, if the model defines priors they should be applied by default (since naively the user expects the priors to be applied in case he defines them). So either remove the option and leave it to the user to implement something in the model parser, or add the inverse of this option to disable model priors.

davidwalter2 (commented):
Here are a few comments to start with, please also add coverage for the new features in the CI, the asymmetric impacts but also maybe other things. If it's not too much work maybe break it up in separate PRs

bin/rabbit_fit.py
lucalavezzo: numpy!

bin/rabbit_fit.py
lucalavezzo: postfit-expected Asimov isn't the same as fit to data? I'm not sure if it makes sense

davidwalter2: Ah you want to construct the asimov data from the external postfit, but why? And is this not confusing? After all the argument is `--externalPostfit` and not `--externalPrefit`

bin/rabbit_fit.py
davidwalter2: Can you not use `ifitter.param_prior_mask.numpy()` instead, I think there is no need for separate _np attribues

lucalavezzo:
Thanks @davidwalter2 for the comments, I've addressed them and split up the PR into 3 more managable ones:
- https://github.com/WMass/rabbit/pull/138
- https://github.com/WMass/rabbit/pull/139
- https://github.com/WMass/rabbit/pull/140
Closing this one

PR #138, opened 2026-06-12
asym impacts

lucalavezzo:
- traditional and global asymmetric impacts, with test coverage
- NOI gathering now uses `nmodel_params = npoi + npou`, so impacts are correct  when the ParamModel has POU (model-nuisance) parameters. POUs never get  impact rows; they act only as impact sources.

### DIFF
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -24,6 +24,11 @@ from rabbit.tfhelpers import edmval_cov
 
 logger = logging.child_logger(__name__)
 
+# Supported constraint-Hessian modes for contour_scan (see its docstring for
+# what each mode does). Single source of truth for the CLI choices and the
+# contour_scan validation.
+CONTOUR_HESS_MODES = ("exact", "hvp", "frozen", "bfgs", "sr1")
+
 
 def match_regexp_params(regular_expressions, parameter_names):
     if isinstance(regular_expressions, str):
@@ -1006,7 +1011,6 @@ class Fitter:
         include=None,
         exclude=None,
         skip_symmetric=False,
-        skip_unconstrained=True,
         contour_xtol=1e-6,
         contour_gtol=1e-6,
         contour_maxiter=5000,
@@ -1014,8 +1018,10 @@ class Fitter:
     ):
         """Traditional asymmetric impacts via per-nuisance contour scan.
 
-        All constrained nuisances are scanned by default; use include/exclude
-        to restrict the selection.
+        All nuisances are scanned by default, including unconstrained ones —
+        like the symmetric traditional impacts, the data still gives them a
+        finite postfit uncertainty and hence a finite Delta(2NLL)=q contour.
+        Use include/exclude to restrict the selection.
 
         Args:
             nll_min: postfit reduced NLL. Computed from the current fit state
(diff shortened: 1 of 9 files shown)


rabbit/impacts/asym_impacts.py
```diff
@@ -0,0 +1,203 @@
...
+The shifts of every fitter parameter at those points are the asymmetric
+impacts. Group impacts follow the traditional grouped-impact convention,
+generalized to the asymmetric case: per side, sqrt(delta^T rho_GG^-1 delta)
+over the group's scanned nuisances, with rho_GG the postfit correlation
+submatrix (see _group_impact).
+
+All constrained nuisances are scanned by default (nonlinear effects can
+produce asymmetric impacts even for structurally symmetric templates);
+use include/exclude to restrict the selection. Unconstrained nuisances
+(constraintweight = 0) are skipped since they have no finite Delta(2NLL)
```
davidwalter2: In the traditional impacts they should have a finite contour, i.e. postfit uncertainty. It is only for the global impacts where they don't have a prefit uncertainty and thus are excluded from the global impacts. But I think they should be included in the traditional impacts

rabbit/impacts/asym_impacts.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def _group_impact(values, rho_inv):
+    """Correlation-aware group combination of asymmetric impacts.
+
+    Asymmetric analogue of the traditional grouped-impact convention
+    (sqrt(v^T C_GG^-1 v) in traditional_impacts._compute_impact_group):
+    per side s, the group impact on each parameter is
+
+        sqrt( delta_s^T rho_GG^-1 delta_s )
```
davidwalter2: This seems like a hybrid method but maybe we should do it in the way that combine is doing it, this is summarized e.g. on slide 3 here: https://indico.cern.ch/event/1641647/contributions/6926122/attachments/3219379/5736736/CMSWeek-GlobalImpacts%20.pdf under "Freeze groups one at a time" @bendavid what do you think?

rabbit/fitter.py
```diff
@@ -936,6 +994,138 @@ def gaussian_global_impacts_parms(self):
...
+            nll_min: postfit reduced NLL. Computed from the current fit state
+                if None.
+            q: contour level (q=1 -> 1 sigma).
+            include: optional regex(es) restricting which nuisances to scan.
+            exclude: optional regex(es) excluding nuisances from the scan.
+            skip_symmetric: optionally skip nuisances whose template content is
+                structurally symmetric (logkhalfdiff identically zero). Off by
+                default since nonlinear effects can produce asymmetric impacts
+                even for symmetric templates.
+            skip_unconstrained: skip nuisances with constraintweight=0 (no
```
davidwalter2: are you sure this is the case?

rabbit/fitter.py
```diff
@@ -936,6 +994,138 @@ def gaussian_global_impacts_parms(self):
...
+        """Fully likelihood-based asymmetric global impacts.
+
+        For each selected nuisance i, shift theta0[i] by +/- sigma (in units of
+        the prefit constraint width) and re-run the full fit. POI shifts at
+        each sign are the asymmetric global impacts.
+
+        Args:
+            include: optional regex(es) restricting which nuisances to scan.
+            exclude: optional regex(es) excluding nuisances from the scan.
+            skip_unconstrained: skip nuisances with constraintweight=0; their
```
davidwalter2: Here I agree, no prefit uncertainty so unconstrained nuisances don't have an impact and should be skipped. Do we need a boolean for this at all? What happens if `skip_unconstrained=False`? Maybe it can be removed

rabbit/fitter.py
```diff
@@ -2073,23 +2263,112 @@ def nll_scan2D(self, param_tuple, scan_range, scan_points, use_prefit=False):
...
+                lg_cache["grad"] = grad.numpy()
+
+            def scipy_hess(x, v):
+                _ensure_hess(x)
+                return v[0] * h_cache["hess"]
+
+        else:
+            raise ValueError(
+                f"contour_scan: unknown hess_mode={hess_mode!r}; "
+                "expected one of 'exact', 'hvp', 'frozen', 'bfgs', 'sr1'."
```
davidwalter2: a list of supported modes should be defined and used everywhere instead of hard-coding multiple times

PR #139, opened 2026-06-12
Unified constraint structure (cw/x0); ParamModel Gaussian priors

lucalavezzo:
Based on #140 #143 #144 to be merged after those.

- unified constraint structure over the full parameter vector — fitter.cw (1/σ²; 0 = free), x0 (centers), var_x0 (1/cw), all index-aligned with x = [POI, POU, NOI, SYST], replacing theta0/param_x0/param_cw/var_theta0. The loss constraint term, prefit variances, toys, and every impact path now read this one structure; the ±nparams offsets cancel, so it's a net code reduction. Breaking: fitter.theta0 is gone (use fitter.x0[fitter.param_model.nparams:]).
- Gaussian priors on ParamModel parameters — opt-in, declared by the model (prior_sigmas/prior_means, no CLA). A prior is just a constraint on a param-block entry (cw=1/σ², x0=mean), so it flows through prefit variances, pulls/constraints, toys (frequentist fluctuates centers, bayesian samples values — fixes the old FIXMEs), and all global impacts (likelihood/gaussian/asym, where priors show up as <param>_prior sources). Priors on POIs require allowNegativeParam=True (guarded).


rabbit/fitter.py
```diff
@@ -300,6 +307,59 @@ def init_fit_parms(
...
+        # disable them (e.g. a token in the --paramModel spec) is up to the
+        # model. The penalty 0.5 * (x - μ)^2 / σ^2 is added to _compute_lc.
+        # param_prior_sigmas / param_prior_means hold the priors as declared
+        # (NaN where no prior); the compute-safe forms are derived in
+        # _compute_lc.
+        self.param_prior_active = False
+        pm = self.param_model
+        sigmas = getattr(pm, "prior_sigmas", None)
+        if sigmas is not None:
+            sigmas = np.asarray(sigmas, dtype=np.float64)
```
davidwalter2: Use `self.indata.dtype` instead to stay consistent with other places, also further down

davidwalter2 (commented):
This looks overall very good but I have one design question, could we align the constraints mechanism with the thetas (as discussed offline)?

rabbit/fitter.py
```diff
@@ -274,6 +285,116 @@ def init_fit_parms(
...
+                    raise ValueError(
+                        "Gaussian priors on POIs require allowNegativeParam="
+                        "True: with allowNegativeParam=False the stored "
+                        "parameter is sqrt(poi), so the Gaussian penalty "
+                        "would apply to sqrt(poi) rather than to the POI "
+                        "itself."
+                    )
+                self.param_prior_active = True
+                self.param_prior_mask = tf.constant(mask, dtype=tf.bool)
+                self.param_prior_sigmas = tf.constant(
```
davidwalter2: These are duplicated in ParamModel and fitter class which calls for trouble, we should probably only define them in only one of them.

rabbit/impacts/global_asym_impacts.py
```diff
@@ -78,76 +90,104 @@ def global_asym_impacts_parms(
...
+    warmstart_col = None
     if linear_warmstart:
         if fitter.cov is None:
             raise RuntimeError(
                 "global_asym_impacts: linear_warmstart requires fitter.cov "
                 "(incompatible with --noHessian)."
             )
         t_lws = time.perf_counter()
-        dxdtheta0_tf, _, _ = fitter._dxdvars()
+        dxdtheta0_tf, dxdparam0_tf, _, _ = fitter._dxdvars()
```
davidwalter2: I think theta0 and param0 should be treated on equal footing, so here instead of returning the two `dxdtheta0_tf` and `dxdparam0_tf` collection return one `dxdx0_tf`

rabbit/fitter.py
```diff
@@ -1189,46 +1389,91 @@ def _pd2ldbeta2(self, profile=False):
...
+        return vartheta0, varparam0
+
     def _dxdvars(self):
+        # Differentiate the loss only w.r.t. the constraint sources (the
+        # nuisance block plus any priored params) by perturbing those x0
+        # entries with `ds` and scattering back into the full vector; see
+        # _x0_constraint_source_idxs. The resulting columns are ordered
+        # [theta block | priored params] and split below.
+        src_idxs = self._x0_constraint_source_idxs()
+        n_src = self.indata.nsyst + (
```
davidwalter2: This looks very error prone, what do we gain from sticking together two collections, I think it is much cleaner to work with a single vector `self.x0` which replaces the previous `self.theta0` and contains all params and systs.

rabbit/impacts/global_impacts.py
```diff
@@ -267,12 +279,25 @@ def global_impacts_parms(
             pd2ldbeta2_pdexpdbeta=None,
         )
 
+    # impacts_x0 covers the full constraint term: the nuisance block AND any
+    # ParamModel prior penalties (rows in the ParamModel block, nonzero where
+    # the model declares priors). Per-1-sigma units come out automatically
+    # since sc = sqrt(d2lc/dx2) = sqrt(cw) = 1/sigma.
     impacts_x0 = _compute_global_impacts_x0(x, compute_lc_fn, cov_dexpdx)
```
davidwalter2: When params and systs are treated on equal footing I think it should be possible to work with `impacts_x0` without splitting them into the two groups, or am I missing something?

davidwalter2 (commented):
The params and systs are not really unified in this PR, there are lot's of places where the two are split and/or merged together with specific treatment which isn't getting clear to my why it is needed. Is it possible to unify them better and work with one collection in most cases?  

Can the modifications on blinding groups and the bugfix on composite param models each one be moved into separate PRs?

lucalavezzo:
Thanks @davidwalter2 I tried to improve a bit the unification after discussing. Everything should be more folded into a single cw / x0 / var_x0. Can you take a look to see if this is closer to what had in mind?

bin/rabbit_fit.py
```diff
@@ -410,16 +410,34 @@ def save_hists(args, mappings, fitter, ws, prefit=True, profile=False):
...
+                fitter_saturated.x0[composite_model.nparams :].assign(
+                    toy_x0[orig_model.nparams :]
+                )
+                if orig_model.npoi > 0:
+                    fitter_saturated.x0[: orig_model.npoi].assign(
+                        toy_x0[: orig_model.npoi]
+                    )
+                if orig_model.npou > 0:
+                    fitter_saturated.x0[
+                        composite_model.npoi : composite_model.npoi + orig_model.npou
```
davidwalter2, line 439: Use `composite_model.nparams` as in the line below?

bin/rabbit_fit.py
```diff
@@ -859,6 +877,26 @@ def main():
         "nois": ifitter.parms[ifitter.param_model.nparams :][indata.noiidxs],
     }
 
+    # ParamModel Gaussian priors (if the model declared sigmas). Read straight
+    # from the model (the single source of truth) so downstream tooling can see
+    # what was applied without parsing the rabbit log.
+    if getattr(ifitter, "param_prior_active", False):
+        pm = ifitter.param_model
+        sigmas = np.asarray(pm.prior_sigmas, dtype=np.float64)
```
davidwalter2: Should this not rather be `tf` objects instead of `np`? And use `ifitter.indata.dtype` here and below?

rabbit/impacts/global_asym_impacts.py
```diff
@@ -48,28 +48,40 @@ def _envelope(values):
...
+            to build the grouped quadrature envelopes. Covers both syst groups
+            and ParamModel impact groups.
         sigma: shift magnitude in units of the prefit constraint width
             (constraints are unit-sigma in rabbit, so 1.0 = 1 prefit sigma).
         signs: sequence (down, up). Bin 0 of axis_downUpVar -> first sign.
         linear_warmstart: experimental. If True, warm-start each refit at
-            x_nom + dxdtheta0[:, i] * shift, the Gaussian-approximation new
-            minimum for the shifted theta0. Should drastically reduce the
-            number of optimizer iterations on near-Gaussian nuisances.
+            x_nom + dxds[:, source] * shift, the Gaussian-approximation new
```
davidwalter2: should it not be `dxdx0` rather than `dxds`?

rabbit/impacts/global_asym_impacts.py
```diff
@@ -78,76 +90,94 @@ def global_asym_impacts_parms(
...
     if linear_warmstart:
         if fitter.cov is None:
             raise RuntimeError(
                 "global_asym_impacts: linear_warmstart requires fitter.cov "
                 "(incompatible with --noHessian)."
             )
         t_lws = time.perf_counter()
-        dxdtheta0_tf, _, _ = fitter._dxdvars()
-        dxdtheta0_np = dxdtheta0_tf.numpy()
+        dxdx0_tf, _, _ = fitter._dxdvars()
```
davidwalter2, line 134: just call it `dxdx0` since `tf` is the default used elsewhere

rabbit/impacts/global_impacts.py
```diff
@@ -197,12 +197,24 @@ def _compute_grouped_impacts(
     bin_by_bin_stat,
     bin_by_bin_stat_mode,
     systgroupidxs,
-    impacts_theta0_sq,
+    impacts_sources_sq,
```
davidwalter2: should that not be `impacts_x0_sq`?

rabbit/impacts/global_impacts.py
```diff
@@ -267,26 +286,33 @@ def global_impacts_parms(
...
         )
 
+    # impacts_x0 covers the full constraint term over the whole parameter
+    # vector. The reported sources (constrained nuisances + priored params)
+    # are a single collection picked out by source_idxs; params and systs are
+    # treated identically here. Per-1-sigma units come out automatically since
+    # sc = sqrt(d2lc/dx2) = sqrt(cw) = 1/sigma.
     impacts_x0 = _compute_global_impacts_x0(x, compute_lc_fn, cov_dexpdx)
-    impacts_theta0 = tf.transpose(impacts_x0[nmodel_params:])
+    impacts_sources = tf.transpose(tf.gather(impacts_x0, source_idxs, axis=0))
```
davidwalter2: Why is this needed? Can we not work with `impacts_x0` directly?

rabbit/impacts/global_impacts.py
```diff
@@ -357,25 +385,28 @@ def global_impacts_obs(
         )
 
     impacts_x0 = _compute_global_impacts_x0(x, compute_lc_fn, cov_dexpdx)
-    impacts_theta0 = tf.transpose(impacts_x0[nmodel_params:])
+    # single source collection (constrained nuisances + priored params)
+    impacts_sources = tf.transpose(tf.gather(impacts_x0, source_idxs, axis=0))
```
davidwalter2: same comment as above, can we not use `impacts_x0` to compute 
```
impacts_x0_sq = tf.square(impacts_x0)
var_x0 = tf.reduce_sum(impacts_x0_sq, axis=-1)
```
and pass `impacts_x0_sq` into `_compute_grouped_impacts` instead of `impacts_sources_sq`

rabbit/impacts/global_impacts.py
```diff
@@ -197,12 +197,24 @@ def _compute_grouped_impacts(
...
+    impacts_sources_sq,
     impacts_nobs,
     impacts_beta0_total,
     impacts_beta0_process,
+    n_param_sources=0,
 ):
-    """Assemble the grouped impacts tensor from all contributions."""
+    """Assemble the grouped impacts tensor from all contributions.
+
+    impacts_sources_sq holds the squared per-source impacts as a single
```
davidwalter2: I'd prefer to treat them as one `impacts_x0_sq` collection. If you compute them separate, how to combine param type sources with syst type sources into a single group?

rabbit/impacts/global_impacts.py
```diff
@@ -384,18 +415,34 @@ def global_impacts_obs(
...
     bin_by_bin_stat,
     bin_by_bin_stat_mode,
     beta_shape,
     systgroupidxs,
+    n_param_sources=0,
     data_cov_inv=None,
 ):
+    # dxdx0 / varx0 are the single source collection ordered
+    # [nuisance block | priored params]. The leading nuisance columns and the
+    # trailing priored-param columns are split out only to assemble the
```
davidwalter2: Similar here, can the split be avoided?

rabbit/fitter.py
```diff
@@ -298,6 +298,125 @@ def init_fit_parms(
 
         self.x = tf.Variable(xdefault, trainable=True, name="x")
 
+        # ParamModel Gaussian priors. The ParamModel itself decides whether
```
davidwalter2: This description is param model specific and should be removed (I think everything is already stated in the param model already)

rabbit/fitter.py
```diff
@@ -298,6 +298,125 @@ def init_fit_parms(
...
+        #
+        # Priors use the same constraint structure as the nuisance
+        # constraints: declared priors set the constraint weight
+        # (1/sigma^2; 0 = free) and the constraint center (the prior mean)
+        # of the ParamModel block in the full-length cw / x0 vectors built
+        # below. _compute_lc adds 0.5 * cw * (x - x0)^2 over the full
+        # parameter vector, prefit variances are 1/cw, and the toy
+        # randomization fluctuates the centers wherever cw > 0.
+        pm = self.param_model
+        np_dtype = self.indata.dtype.as_numpy_dtype
```
davidwalter2, line 307: Is this `tf`->`np`->`tf` conversion forth and back necessary or can it be in `tf`?

rabbit/fitter.py
```diff
@@ -708,15 +810,22 @@ def set_nobs(self, values, variances=None):
...
+        self.x0.assign(self.x0default)
 
     def xdefaultassign(self):
         if self.param_model.nparams == 0:
-            self.x.assign(self.theta0)
+            self.x.assign(self.x0)
         else:
             self.x.assign(
-                tf.concat([self.param_model.xparamdefault, self.theta0], axis=0)
+                tf.concat(
```
davidwalter2: Is there now a `self.x0default` attribute defined above that let's you get rid of this if-else block?

rabbit/fitter.py
```diff
@@ -740,32 +849,35 @@ def defaultassign(self):
...
-                    [
-                        self.param_model.xparamdefault,
-                        self.theta0
-                        + tf.random.normal(
-                            shape=self.theta0.shape, dtype=self.theta0.dtype
-                        ),
-                    ],
-                    axis=0,
-                )
+            defaults = tf.concat(
```
davidwalter2: Same comment as above, can `self.x0default` be used instead?

rabbit/fitter.py
```diff
@@ -298,6 +298,125 @@ def init_fit_parms(
...
+        )
+        self.param_prior_active = bool(int(self.param_prior_idxs.shape[0]) > 0)
+        self.x0_source_idxs = tf.concat(
+            [
+                tf.range(pm.nparams, len(self.parms), dtype=tf.int64),
+                self.param_prior_idxs,
+            ],
+            axis=0,
+        )
+        self.var_x0_sources = tf.gather(self.var_x0, self.x0_source_idxs)
```
davidwalter2: I think we should get rid of this attribute and use `self.var_x0` everywhere instead, even if we carry some zeros with us, they should not matter

rabbit/fitter.py
```diff
@@ -1263,45 +1414,61 @@ def _pd2ldbeta2(self, profile=False):
         return pd2ldbeta2
 
     def _dxdvars(self):
+        # Response of the postfit minimum to a unit shift of each constraint
```
davidwalter2, line 1373: Instead of working with this reduced set of `source` attributes I suggest to replace the `self.theta0` from the previous implementation with `self.x0` even if we have parameters without priors that we carry for no reason, I think it is better to have code simplicity versus overly specific logic

rabbit/fitter.py
```diff
@@ -1899,13 +2076,26 @@ def full_nll(self):
     def reduced_nll(self):
         return self._compute_nll(full_nll=False)
 
-    def _compute_lc(self, full_nll=False):
-        # constraints
-        theta = self.get_theta()
-        lc = self.indata.constraintweights * 0.5 * tf.square(theta - self.theta0)
+    def _compute_lc(self, full_nll=False, x0=None):
```
davidwalter2: Related to the other comments, I think we should get rid of the `x0` function argument since it's leading to function branches

davidwalter2 (commented):
A few more comments

bin/rabbit_fit.py
lucalavezzo: done

PR #140, opened 2026-06-12
fit ops qol

lucalavezzo:
- pass some minimizer options via CLI  `--minimizerMaxiter`, `--minimizerGtol`, `--minimizerFtol`
- timing output in debug mode
- `--externalPostfit` covariance handling: when the external fitresult has no covariance (e.g. produced with `--noHessian`), recompute the Hessian at the loaded postfit point — enabling the two-pass recipe (fit with `--noHessian`, then `--externalPostfit … --noFit` to get the covariance)

### DIFF
--- a/bin/rabbit_fit.py
+++ b/bin/rabbit_fit.py
@@ -434,24 +434,41 @@ def fit(args, fitter, ws, dofit=True):
         ws.add_1D_integer_hist([max_curvature], "best", "lcurve")
 
     if dofit:
+        _t_min = time.perf_counter()
         cb = fitter.minimize()
+        logger.debug(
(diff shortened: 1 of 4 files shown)


davidwalter2:
I still don't agree that the asimov should be done with the `--externalPostfit` IMO it is more intuitive to expect that the asimov is done from the nominal model before the `--externalPostfit` is loaded and only the postfit is made with it. What was your thought about this?

lucalavezzo:
Thanks @davidwalter2. The idea originally was to be able to get the expected sensitivity at the data-driven point. I don't think this is currently supported.  For now, I've reverted, as indeed for a full Asimov fit, you are correct (i.e. one must use the prefit Asimov even with --externalPrefit is used).

PR #141, opened 2026-06-15
Limited-MC-statistics de-biasing for binned template fits (continuous-M + cross-fit)

bendavid:
## Limited-MC-statistics de-biasing for binned template fits

Adds opt-in de-biasing of the bias and under-coverage that finite MC template
statistics induce in chi-squared / profile-likelihood template fits (the
errors-in-variables / Neyman–Scott "incidental parameters" effect). Two
complementary methods, both injected at the scalar-NLL level so they flow through
rabbit's existing TF-autodiff grad/Hessian/HVP machinery and SciPy minimizers.

**Draft** — opening for design review and discussion of the interfaces before polish
(the test scripts are standalone validation drivers, not yet pytest/CI-wired).

### Methods
- **continuous-M** — a frozen noise-floor matrix `M` (supplied to the `TensorWriter`)
  enters the objective as `-½ θᵀ M θ`, giving the de-biased point `(H−M)⁻¹g` and
  curvature `H−M`. Lowest variance; requires the de-biased parameters to enter the
  prediction **linearly** (e.g. `--allowNegativeParam` POIs); a warning fires otherwise.
- **two-half / k-fold cross-fit** — the user fills histograms with an extra MC-stat
  `fold_axis`; the writer slices it (full derived by projecting it out). The fitter
  forms the jackknife `2L_full − ½L_A − ½L_B`; the `--covMode fisher` curvature uses
  the complete pairwise **U-statistic** over all k folds (k-fold averaging, O(k), no
  partition enumeration, → continuous-M as k→N). General / nonlinearity-robust.

### Interfaces
- `TensorWriter.add_mc_stat_moment(M, param_names)`; `add_process(..., fold_axis=)`,
  `add_systematic(..., fold_axis=)` (split-logk, symmetric/asymmetric, dense/sparse).
- `--mcStatDebias {none,continuousM,twoHalf,kfold}`,
  `--mcStatDebiasCov {curvature,sandwich,dataPropagated}`,
  `--covMode {observed,fisher}` (applies to the standard cov too).
- Robust (sandwich) covariance; BB-lite compatible; the de-biased covariance flows
  through to impacts (a new `mcStatDebias` impact group) and output histograms.

### Validation
- Per-toy **exact match to an independent numpy reference** (de-biased point,
  curvature, sandwich) for continuous-M and two-half; k-fold U-statistic == numpy and
  RMS↓ with k; sparse == dense and split-logk sparse == dense to 1e-6.
- A native **template-fluctuating coverage harness** (`tests/mcstat_coverage.py`):
  the de-bias removes the central-value attenuation bias and inflates σ→σ∞; the
  **proper** sandwich (BB-lite σ_MC in the meat) covers ~0.68 with no calibration,
  while the bare-Hessian meat under-covers and the data-propagated meat over-covers.

### Known limitations (documented)
- continuous-M needs linear parameters; for nonlinear use two-half.
- BB-lite on pathologically degenerate problems can give a rank-deficient covariance.
- `dataPropagated`/sparse-split-logk are continuous-M/dense and log_normal-symmetric
  respectively; large-`nparams` `--noHessian` (matrix-free) de-biased covariance is
  not yet implemented (the HVPs are automatic; only a sandwich-aware Hessian-free row
  extractor is missing).

### DIFF
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -1959,6 +1959,42 @@ class Fitter:
                 axis=-1,
             )
 
+            if templates != "full":
+                # De-bias fold/half yields in sparse mode: the per-fold/half norm
+                # templates are stored DENSE, so scatter the (shared) systematic
+                # factor from the sparse logk into a dense [nbinsfull, nproc] grid
+                # and contract with the dense norm. Split-logk (per-fold logk) is
+                # not supported in sparse mode.
+                if templates == "A":
+                    norm_dense = self.norm_A
+                elif templates == "B":
+                    norm_dense = self.norm_B
+                else:
+                    norm_dense = self.indata.norm_folds[fold_index]
+                if self.logk_folds_scaled is not None:
+                    raise NotImplementedError(
+                        "split-logk (per-fold logk) is not supported in sparse "
+                        "mode; use shared logk (no fold_axis on systematics)."
+                    )
+                idx = self.indata.norm.indices  # [nnz, 2] = (bin, proc)
+                shape = [self.indata.nbinsfull, self.indata.nproc]
+                if self.indata.systematic_type == "log_normal":
+                    factor = tf.tensor_scatter_nd_update(
+                        tf.ones(shape, dtype=norm_dense.dtype), idx, tf.exp(logsnorm)
+                    )
+                    normcentral = norm_dense * rnorm * factor
+                else:  # "normal": additive variation (norm-independent)
+                    add = tf.tensor_scatter_nd_update(
+                        tf.zeros(shape, dtype=norm_dense.dtype), idx, logsnorm
+                    )
+                    normcentral = norm_dense * rnorm + add
+                if not (full or self.indata.nbinsmasked == 0):
+                    normcentral = normcentral[: self.indata.nbins]
+                nexpcentral = tf.reduce_sum(normcentral, axis=-1)
+                if not compute_norm:
+                    normcentral = None
+                return nexpcentral, normcentral
+
             # Build a sparse [nbinsfull, nproc] tensor whose values absorb
             # the per-entry syst variation and the per-(bin, proc) POI
             # scaling rnorm. The sparsity pattern is unchanged from
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -266,9 +266,7 @@ class Fitter:
             # syst (and asym) axes.
             if self.indata.systematic_type == "normal" and self.param_model.nparams > 0:
                 rnorm_init = tf.broadcast_to(
-                    self.param_model.compute(
-                        self.param_model.xparamdefault, full=True
-                    ),
+                    self.param_model.compute(self.param_model.xparamdefault, full=True),
                     [self.indata.nbinsfull, self.indata.nproc],
                 )
                 ntrail = len(logk_folds.shape) - 3  # 1 (sym) or 2 (asym)
@@ -316,11 +314,7 @@ class Fitter:
                 "templates (hnorm_folds) found in the input; two-half de-biasing "
                 "is inactive. Add processes with a fold_axis in the TensorWriter."
             )
-        if (
-            self.n_folds is not None
-            and self.n_folds >= 3
-            and self.covMode != "fisher"
-        ):
+        if self.n_folds is not None and self.n_folds >= 3 and self.covMode != "fisher":
             logger.info(
                 f"k-fold de-biasing with k={self.n_folds} folds: the k-fold "
                 "AVERAGING (complete U-statistic, lower curvature-estimate "
@@ -1213,10 +1207,10 @@ class Fitter:
         pd2ldxdnobs, pd2ldxdnorm = t2.jacobian(
             grad, [self.nobs, norm], unconnected_gradients="zero"
         )
-        dxdnobs = -Ainv @ pd2ldxdnobs                              # [npar, nbins]
+        dxdnobs = -Ainv @ pd2ldxdnobs  # [npar, nbins]
         dxdnorm = -Ainv @ tf.reshape(
             pd2ldxdnorm, [tf.shape(pd2ldxdnorm)[0], -1]
-        )                                                          # [npar, nbinsfull*nproc]
+        )  # [npar, nbinsfull*nproc]
         var_nobs = self.varnobs if self.varnobs is not None else self.nobs
         sumw2_flat = tf.reshape(self.indata.sumw2, [-1])
         return (dxdnobs * var_nobs[None, :]) @ tf.transpose(dxdnobs) + (
@@ -2035,10 +2029,7 @@ class Fitter:
                 # systematic-template noise is de-biased. Curvature path only
                 # (templates='fold'); halves use the shared logk.
                 split_corr = None
-                if (
-                    self.logk_folds_delta is not None
-                    and templates == "fold"
-                ):
+                if self.logk_folds_delta is not None and templates == "fold":
                     theta_folded = tf.gather(theta, self.mcstat_folded_syst_idx)
                     split_corr = tf.einsum(
                         "bpj,j->bp", self.logk_folds_delta[fold_index], theta_folded
@@ -2428,9 +2419,7 @@ class Fitter:
 
         nexp = nexpfullcentral[: self.indata.nbins]
 
-        debiased = (
-            self.mcStatDebias in ("twoHalf", "kfold") and self.norm_A is not None
-        )
+        debiased = self.mcStatDebias in ("twoHalf", "kfold") and self.norm_A is not None
         if debiased:
             # Cross-fit jackknife combination L_cf = 2 L_full - 1/2 L_A - 1/2 L_B.
             # Since mu_bar = 1/2(n_A + n_B) = n_full exactly, this de-biases the
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -247,20 +247,31 @@ class Fitter:
                 "templates (hnorm_folds) found in the input; two-half de-biasing "
                 "is inactive. Add processes with a fold_axis in the TensorWriter."
             )
-        if self.mcStatDebias == "continuousM" and self.mcstat_M is None:
-            logger.warning(
-                "--mcStatDebias continuousM requested but no 'mcstat' external "
-                "moment term found in the input; continuous-M is inactive. Supply "
-                "M via TensorWriter.add_mc_stat_moment(M, param_names)."
-            )
-        if self.mcStatDebias == "continuousM" and not self.param_model.is_linear:
-            logger.warning(
-                "--mcStatDebias continuousM: the param model is NONLINEAR "
-                "(e.g. the default rabbit x^2 POI transform). The -1/2 theta^T M theta "
-                "penalty is cancelled by Fisher growth at the shifted minimum and will "
-                "NOT de-bias the point or covariance (RESULTS.md S9a). Use "
-                "--allowNegativeParam (linear POI) or the two-half method instead."
-            )
+        if self.mcStatDebias == "continuousM":
+            if self.mcstat_M is None:
+                logger.warning(
+                    "--mcStatDebias continuousM requested but no 'mcstat' external "
+                    "moment term found in the input; continuous-M is inactive. Supply "
+                    "M via TensorWriter.add_mc_stat_moment(M, param_names)."
+                )
+            elif not self._mcstat_M_params_linear():
+                # The -1/2 theta^T M theta penalty only de-biases parameters that
+                # enter the prediction LINEARLY. If any parameter that M touches
+                # is nonlinear (e.g. the default x^2 POI transform, or log-normal
+                # systematics), the data Fisher grows as the -M theta gradient
+                # shifts the minimum and cancels the -M curvature subtraction
+                # (RESULTS.md S9a) -> no de-bias of the point or covariance for
+                # those parameters. Two-half de-biasing has no such restriction.
+                logger.warning(
+                    "--mcStatDebias continuousM: the supplied M acts on parameters "
+                    "that enter the prediction NONLINEARLY (e.g. the default rabbit "
+                    "x^2 POI transform [use --allowNegativeParam for a linear POI], "
+                    "or log_normal systematics). The -1/2 theta^T M theta penalty is "
+                    "cancelled by Fisher growth at the shifted minimum and will NOT "
+                    "de-bias the point or covariance for those parameters "
+                    "(RESULTS.md S9a). Use a linear parametrization for the de-biased "
+                    "parameters, or the two-half method (--mcStatDebias twoHalf)."
+                )
 
     def init_fit_parms(
         self,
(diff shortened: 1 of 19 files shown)


davidwalter2 (commented):
Design review as asked. I built the branch, ran the toy generators and the verification drivers, and benchmarked the fold paths, rather than reading only.

**The method works and the validation is real.** `verify_twohalf` reproduces the numpy reference exactly (`rabbit == numpy reference: True`, cf point `r=[1.1478 0.8568]`, curv 0.1367, sandwich 0.2224 on both sides), `verify_kfold` matches the U-statistic and shows the RMS falling toward σ∞ with k (0.0318 → 0.0233 → 0.0211 at k=4/8/16 against σ∞=0.1343), and `verify_mcstat` reproduces the documented linear-vs-squared POI behaviour (0.0840 → 0.1548 de-biased with `--allowNegativeParam`; unchanged with the default x² transform). The jackknife algebra checks out — for a quadratic ℓ, `2L_full − ½L_A − ½L_B` has Gauss-Newton part `J_Aᵀ D J_B` exactly, and `k/(k−1)[J_fullᵀDJ_full − Σ_i J_iᵀDJ_i]` has the right normalisation for the full-sample Fisher. black clean; the 5 F541s in `tensorwriter.py` are pre-existing on the base.

So my comments are about **scale and packaging**, not about whether the statistics are right.

### The one that decides whether this is usable

`--covMode fisher` is **O(nbins²)** — measured, steady state, at **2 parameters**:

| nbins | `fisher_curvature` (2nd+ call) | trace |
|---|---|---|
| 800 | 67 ms | 4.5 s |
| 1600 | 355 ms | 3.8 s |
| 3200 | 1147 ms | 7.5 s |
| 6400 | 8167 ms | 16.8 s |

8× in nbins → 122× in time. Separately it grows ~linearly in nparams (nparams 2→32 at nbins=800: 0.10 s → 2.18 s). A real rabbit fit has nbins in the 10⁴–10⁵ range and nparams in the 10³–10⁴ range, so this does not reach production as written. Details and a candidate fix inline.

That matters more than a normal performance note because the PR states the k-fold averaging "is realized only in `--covMode fisher`" — so the headline k>2 capability is exactly the part that doesn't scale. The **objective** is fine: the fold NLL is 0.8–2.0× the plain NLL across the same range, so `--mcStatDebias kfold --covMode observed` costs essentially nothing in the minimisation.

### Packaging

- **The base is 3 months stale.** Merge-base is `ffb8039` (2026-06-12); **63 commits** on main since, including 862 lines of churn in `fitter.py` alone (preconditioning, the restart loop, snapshots). GitHub reports `CONFLICTING / DIRTY`. The good news is that a test merge produces only **2 conflict hunks**, both in `fitter.py` — so this is a rebase, not a rewrite, but it should happen before review goes much further, because several of the new `@tf.function`s touch code paths that were restructured in that window.

- **None of the 1400 lines of test code runs in CI.** pytest collects 15 tests on this branch — the same set as the base. All twelve new files are named `verify_*.py` / `toy_*.py` / `mcstat_coverage.py` and contain **zero** `test_` functions, so pytest's default `python_files` pattern never picks them up. You flag this as a known limitation, and I agree it needn't block a draft, but it's worth knowing the number is zero rather than "not yet polished": nothing here is protected against regression today. They also print rather than assert (`verdict = "DE-BIASED" if s_M > 1.2*s_no else ...`), so wiring them up is more than a rename.

- **18 code references to two documents that aren't in the repo.** `RABBIT_MCSTAT_DESIGN.md` (7×) and `RESULTS.md` (5×, plus variants) are cited for every substantive design decision — §1b for the BB-lite denominator, §2c/§2d for the Fisher construction, §9a for the linearity cancellation — and neither file exists on the branch or is added by the PR. For a PR opened specifically for design review, that's the part I'd fix first: right now the reasoning is only checkable by re-deriving it.

Interface comments inline. Overall this is a substantial and careful piece of work — my reservations are about whether the expensive path can reach production scale and about the reader being able to follow the argument, not about the statistics.

rabbit/fitter.py
```diff
@@ -789,6 +943,331 @@ def toyassign(
...
+        # bread; using each fold term's own raw n_i (~nexp_full/k) breaks the
+        # full-minus-self identity F_sumfold - sum_i F_i = sum_{i!=j} cross
+        # (with per-term D it leaves a -sum_i F_i self-term, driving the bread
+        # indefinite -> Cholesky fails). fisher_terms[0] is the full / sum-of-
+        # folds prediction; for every non-fold path there is a single term with
+        # n == nexp_full, so nref reproduces the previous 1/n exactly.
+        nref = fisher_terms[0][0]
+        F = tf.zeros_like(hess_obj)
+        for n, c in fisher_terms:
+            J = t1.jacobian(n, self.x)  # [nbins, nparams]
```
davidwalter2, line 1156: This is where the O(nbins²) comes from, and it's the thing that decides whether `--covMode fisher` is usable.

`t1.jacobian(n, self.x)` takes a reverse-mode jacobian of an **[nbins]** output w.r.t. an **[nparams]** input, k+1 times, on a `persistent=True` tape that has also recorded the `_ln_terms_for_fisher` graph. Reverse mode costs one backward pass per *output* element, so this is the wrong orientation whenever nparams ≪ nbins — which is the normal rabbit regime for the data term.

Measured steady state (2nd+ call, so not tracing), at **npar = 2**:

```
nbins    fisher_curvature    trace
 800          67 ms          4.5 s
1600         355 ms          3.8 s
3200        1147 ms          7.5 s
6400        8167 ms         16.8 s
```

8× in nbins → 122× in time. On top of that it grows ~linearly in nparams (0.10 s → 2.18 s for nparams 2 → 32 at nbins=800). Extrapolating the quadratic to a realistic nbins ~ 5e4 gives ~8 minutes at *two* parameters, before the nparams factor.

Forward mode over the parameter axis is the natural fix: nparams JVPs, each O(nbins), giving O(nbins·nparams) instead of O(nbins²). I checked it produces a **bit-identical** jacobian and is already 4–9× faster in eager mode at npar=2:

```
nbins   reverse J   forward J   speedup   max|J_rev - J_fwd|
 800      186 ms       44 ms      4.2x         0.00e+00
1600      296 ms       33 ms      9.1x         0.00e+00
3200      825 ms      154 ms      5.3x         0.00e+00
6400     1474 ms      236 ms      6.3x         0.00e+00
```

via `tf.autodiff.ForwardAccumulator(self.x, e_i)` around the yield computation. That won't be the whole story — the isolated jacobian is roughly linear in nbins in eager mode, so some of the quadratic is likely the persistent tape being replayed k+1 times, or pfor's graph inside `@tf.function` — but it is the piece with a clear O() improvement and it's verifiable against the current code to zero difference.

Worth noting the trace times too (4–17 s): with `@tf.function` these are paid per distinct k, and the per-fold Python loop unrolls the graph k-fold.

Since the k-fold averaging is only realized in `--covMode fisher`, this path is the flagship capability of the PR, so I'd want the scaling settled before the interfaces are frozen. `--mcStatDebias kfold --covMode observed` is unaffected — the fold objective measures 0.8–2.0× the plain NLL across the same nbins range, i.e. essentially free.

rabbit/tensorwriter.py
```diff
@@ -2104,6 +2566,28 @@ def create_dataset(
             sumw2, f, "hsumw2", maxChunkBytes=self.chunkSize
         )
 
+        # MC-stat cross-fit fold templates [k, nbinsfull, nproc] (dense), only
+        # written when at least one process was added with a fold_axis.
+        if norm_folds is not None:
+            nbytes += h5pyutils_write.writeFlatInChunks(
+                norm_folds, f, "hnorm_folds", maxChunkBytes=self.chunkSize
```
davidwalter2, line 2573: `hnorm_folds` is written dense `[k, nbinsfull, nproc]` unconditionally — I confirmed it is exactly k× the size of `hnorm` (4.0× at k=4 in every configuration I built).

In dense mode that's the expected price and it's fine. In **sparse** mode it isn't: `norm` is stored CSR precisely because it is mostly zeros, and this stores the same information densified and multiplied by k. `_add_process_folded`'s docstring is honest about it ("the per-fold fold templates are densified"), but the consequence is worth stating in the option help rather than only in a writer docstring: turning on `fold_axis` removes the memory benefit of sparse mode for the nominal templates.

That bites hardest exactly where sparse mode is load-bearing — an unfolding setup where each signal bin is its own process, so `nproc` is O(10²–10³) and `norm` is nearly diagonal in (bin, proc). There `k × nbinsfull × nproc` dense is the configuration sparse mode exists to avoid.

`hlogk_folds` `[k, nbinsfull, nproc, nsyst]` has the same shape problem one dimension worse; I see the sparse split-logk path routes around it with `logk_folds_delta` over folded systs only, which looks like the right instinct. Is there a reason the same trick (store per-fold *deltas* against the full template, over the folds' support only) wouldn't work for `norm_folds`? For k folds of a common template the deltas are the only new information, and they inherit the sparsity pattern of `norm`.

rabbit/tensorwriter.py
```diff
@@ -246,6 +268,114 @@ def add_process(self, h, name, channel="ch0", signal=False, variances=None):
...
+        # sumw / sumw2 and all downstream machinery are identical to a normal
+        # process. variances default to the full hist's variances (= sum of
+        # fold variances), which is what BB-lite needs.
+        self.add_process(
+            h_full, name, channel=channel, signal=signal, variances=variances
+        )
+        self.dict_norm_folds[channel][name] = norm_folds.astype(self.dtype)
+        self.fold_k[channel][name] = k
+
+    def add_mc_stat_moment(self, M, param_names, name="mcstat_M"):
```
davidwalter2, line 327: Design comment on `add_mc_stat_moment`, since interfaces are what you asked about.

This asks the histmaker to accumulate an `nparams × nparams` matrix in the event loop whose denominator is *the fit's* per-bin variance — and, as the docstring says, that denominator has to be the BB-lite-**inflated** `mu_b + Σ_proc sumw2_b` when BB-lite is on, or `H − M` goes non-PD and the fit diverges.

That makes the input file depend on the fit configuration. Today a rabbit tensor is a description of the templates and a fit option is a description of the fit; with this, `binByBinStat` on/off changes what a *correct* input file contains. The failure mode is also bad: silently wrong `M` → indefinite curvature → Cholesky failure, far from the histmaker that produced it, with no way for the fitter to detect it. The runtime warning in `Fitter.__init__` helps but it can only say "you might have this wrong", not check.

Two things that would make it safer without changing the method:

1. Store what's configuration-independent and let the fitter finish the job. If the writer stored the score outer products with the *bare* denominator (or the un-normalised `Σ_e w_e² s_i s_j` per bin), the fitter could divide by whichever variance it is actually using. That removes the coupling entirely and the file stays valid under both BB-lite settings.
2. Failing that, record the denominator convention as an attribute next to the term (e.g. `mcstat_M_variance = "bblite_inflated" | "bare"`) and have the fitter **raise** on a mismatch rather than warn. A wrong `M` is not a preference.

The `fold_axis` interface is much nicer by comparison — it asks the histmaker only for an extra axis and derives everything else. Its own subtle requirement (all oversampled copies of one *generated* event in the same fold, assignment independent of the fitted observables) is stated in `_add_process_folded`'s docstring but can't be checked either; a note in the `--mcStatDebias` help pointing at it would help, since that's where a user starts.

rabbit/fitter.py
```diff
@@ -214,6 +214,160 @@ def __init__(
...
+        self.norm_A = None
+        self.norm_B = None
+        # n_folds and the raw per-fold templates are kept for the complete
+        # k-fold U-statistic curvature (k-fold averaging, see _ln_terms_for_fisher).
+        self.n_folds = None
+        norm_folds = getattr(self.indata, "norm_folds", None)
+        if norm_folds is not None:
+            k = int(norm_folds.shape[0])
+            if k % 2 != 0:
+                raise NotImplementedError(
```
davidwalter2, line 240: The even-k restriction is a property of the *objective*, not of the method, and it might be worth not baking it into a hard `NotImplementedError`.

The complete pairwise U-statistic in `_fisher_data_terms` is `k/(k−1)[J_sumfoldᵀ D J_sumfold − Σ_i J_iᵀ D J_i]`, which is well defined for any k ≥ 2 — nothing there needs k even. The evenness comes only from `norm_A`/`norm_B` being literal first/second halves for the objective's `2L_full − ½L_A − ½L_B`.

So k=3 or k=5 could still give a valid de-biased *curvature* while the point falls back to the shared-template objective — which is already the documented behaviour for k>2 with split-logk ("the objective uses shared logk"). Rejecting the input outright means a user who picks k=3 gets an error rather than the k>2 behaviour they'd get at k=4.

If the restriction stays, the message could say which half of the machinery needs it — currently "two-half de-biasing needs an even number of folds" reads as a property of the method rather than of the A/B split in the objective.

rabbit/parsing.py
```diff
@@ -377,6 +377,49 @@ def common_parser():
...
+        help="Covariance reported when --mcStatDebias != none. 'curvature' = inverse "
+        "de-biased Hessian A^-1 (undercovers). 'sandwich' = A^-1 H A^-1 = "
+        "A^-1 + A^-1 M A^-1 (analytic meat; calibration-free ~0.68 with BB-lite, "
+        "undercovers without). 'dataPropagated' = Huber-White delta-method sandwich "
+        "propagating the data AND template (sumw2) variances through theta_hat "
+        "(continuous-M, dense): CONSERVATIVE, over-covers (~0.81, needs a downward "
+        "calibration; RESULTS §7d). Default 'sandwich'.",
+    )
+    parser.add_argument(
+        "--mcStatKfold",
```
davidwalter2, line 405: This adds a user-facing option whose entire help text explains that it does nothing:

> "DEPRECATED / unused: the number of folds k is inferred from the fold-axis size in the input …"

Nothing can be deprecated before it ships. `Fitter.__init__` reads it into `self.mcStatKfold` and nothing else uses it — I grepped. I'd just drop both the argument and the attribute; the "k is inferred from the input, nothing to choose here" point is worth keeping, but it belongs in the `--mcStatDebias` help where a user looking for a fold count would actually look.

rabbit/fitter.py
```diff
@@ -214,6 +214,160 @@ def __init__(
...
+                    "parameters, or the two-half method (--mcStatDebias twoHalf)."
+                )
+            if self.mcstat_M is not None and self.bbstat.enabled:
+                logger.warning(
+                    "--mcStatDebias continuousM with bin-by-bin stat (BB-lite) ON: "
+                    "M must be computed with the BB-lite-inflated variance "
+                    "(mu_b + sum_proc sumw2) in its denominator, NOT the bare data "
+                    "variance. Otherwise M over-subtracts the (already BB-inflated) "
+                    "curvature and H - M becomes non-positive-definite (the fit will "
+                    "diverge / Cholesky will fail). See RABBIT_MCSTAT_DESIGN.md §1b."
```
davidwalter2, line 368: This is one of 18 places in the diff that cite `RABBIT_MCSTAT_DESIGN.md` (7×) or `RESULTS.md` (5×, plus §-only variants) — and neither file exists on the branch or is added by the PR (`git diff --name-only` shows zero `.md` files).

The citations are load-bearing rather than decorative. §1b is the justification for the BB-lite denominator warning right here; §2c/§2d justify dropping the residual·d²μ term in `_fisher_core`; §9a is the whole reason continuous-M is documented as not working with the default POI transform; §7d carries the coverage numbers behind the `--mcStatDebiasCov` default. A reader who wants to check any of those has to re-derive them.

Since the PR is explicitly opened for design review, adding those two documents would do more for reviewability than any code change in it. If they're meant to stay out of the repo, the handful of results the code actually depends on (the BB-lite denominator argument, the GN truncation, the linearity cancellation, the coverage table) could move into the module docstring instead — that's where someone reading `fitter.py` in two years will look.

PR #152, opened 2026-08-26
Avoid copying logk for the normal systematic type

davidwalter2:
## Summary

For `systematic_type == "normal"`, `_init_logk_scaled` absorbed the param-model factor `rnorm_init` into `logk` at construction time. That allocates a second full-size copy of the `[nbins, nproc, nsyst]` tensor and, in sparse mode, rebuilds the CSR matrix.

`rnorm_init` carries no systematic index, so it factors straight out of the contraction over systematics:

```
sum_s (rnorm_init[b,p] * logk[b,p,s]) * theta[s]
  == rnorm_init[b,p] * sum_s logk[b,p,s] * theta[s]
```

This keeps `logk` / `logk_csr` aliased to the indata tensors and applies the factor to the `[nbins, nproc]` contraction *result* instead of the `[nbins, nproc, nsyst]` input.

Only the `normal` systematic type is affected; `log_normal` already carried the scaling multiplicatively and takes the same early return as before.

## Measured effect

The scaling now runs once per loss/HVP evaluation instead of once over the full tensor at construction, so it is worth checking that this does not cost more than it saves. Measured on a 10000-bin x 2048-systematic `normal`-type model (`logk` = 819 MB), with `--minimizerMaxiter` pinned so both versions do matched work:

| | peak RSS | wall time | n_hvp |
|---|---|---|---|
| before | 2115 +- 211 MB | 178 +- 36 s | 1100 |
| after | 1350 +- 18 MB | 190 +- 95 s | 1089 |

- **Memory: 766 +- 122 MB saved** (a 6.3 sigma effect), matching the 819 MB `logk` tensor -- i.e. the copy is gone.
- **Runtime: neutral.** The wall-time difference is 12 +- 59 s, consistent with zero. Call counts are deterministic and differ by 1% (1089 vs 1100), the expected small trajectory shift from the reassociation.

The per-call cost is one elementwise multiply on `[nbins, nproc]`, a factor `nsyst` smaller than the matmul it follows, and it lands on data already in cache.

The saving scales with the tensor: for a 100k-bin x 8192-systematic model the avoided copy is ~33 GB. In sparse mode the win is structural rather than just an allocation -- the old path rebuilt the CSR matrix via `CSRSparseMatrix(...)`, which is replaced by a one-time `gather_nd` onto the `[norm_nnz]` layout the CSR contraction already returns.

`rnorm_init` remains a constant evaluated at `xparamdefault`, so the loss stays strictly linear in theta and derivatives are unchanged.

## Validation

Compared against unpatched `main` using `tests/make_tensor.py`, running the fit with both and comparing fitted parameter values and variances:

| case | max relative difference |
|---|---|
| dense + normal | 5.7e-12 |
| sparse + normal | 4.7e-11 |
| dense + log_normal (control) | 0 (bit-identical) |

Run with `--expectSignal sig 2.5` so that `rnorm_init != 1` and the scaling is genuinely exercised -- with the default `expectSignal=1` the factor is exactly 1.0 and both orderings agree trivially. The residual differences are consistent with the change in summation order (summing then scaling instead of scaling then summing); the new ordering performs one multiply per `(bin, proc)` rather than `nsyst` of them.

### DIFF
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -1620,26 +1620,44 @@ class Fitter:
         return expvars
 
     def _init_logk_scaled(self):
-        """Build an internal copy of indata.logk for the yield-computation
-        hot path, pre-multiplied per (bin, proc) by the param-model factor
-        evaluated at xparamdefault.
+        """Prepare the param-model scaling of the systematic variations for
+        the yield-computation hot path.
 
         For systematic_type == "log_normal" the multiplicative form
         ``rnorm * exp(θ·logk) * norm`` already carries the param-model
-        scaling through to the variation, so no copy is needed and
+        scaling through to the variation, so nothing is needed and
         self.logk / self.logk_csr alias the indata tensors.
 
         For systematic_type == "normal" the linearized variation
-        ``rnorm * norm + θ·logk`` does not scale with rnorm. We absorb a
-        constant rnorm_init = param_model.compute(xparamdefault) into logk
-        once, so the relative size of an additive variation matches the
-        multiplicative case at the linearization point. The scaling is a
-        constant, so the hot path remains strictly linear in θ.
+        ``rnorm * norm + θ·logk`` does not scale with rnorm, so a constant
+        rnorm_init = param_model.compute(xparamdefault) has to be absorbed
+        into the variation, so that the relative size of an additive
+        variation matches the multiplicative case at the linearization
+        point.
+
+        rnorm_init carries no systematic index, so it factors out of the
+        contraction over systematics:
+
+            sum_s (rnorm_init[b,p] * logk[b,p,s]) * theta[s]
+              == rnorm_init[b,p] * sum_s logk[b,p,s] * theta[s]
+
+        We therefore keep logk aliased and apply the factor to the
+        [nbins, nproc] contraction *result* instead of to the
+        [nbins, nproc, nsyst] input. This is exactly equivalent, and avoids
+        allocating a second full-size copy of logk (and, in sparse mode,
+        rebuilding the CSR matrix) at construction time. The scaling is
+        still a constant, so the hot path remains strictly linear in θ.
         """
+        # scaling factors applied to the contraction result; None when the
+        # multiplicative form already carries the param-model scaling
+        self.rnorm_init = None
+        self.rnorm_init_at_norm = None
+
+        self.logk = self.indata.logk
+        if self.indata.sparse:
+            self.logk_csr = self.indata.logk_csr
+
         if self.indata.systematic_type != "normal" or self.param_model.nparams == 0:
-            self.logk = self.indata.logk
-            if self.indata.sparse:
-                self.logk_csr = self.indata.logk_csr
             return
 
         rnorm_init = self.param_model.compute(self.param_model.xparamdefault, full=True)
@@ -1648,27 +1666,14 @@ class Fitter:
         )
 
         if self.indata.sparse:
-            # logk dense shape is [norm_nnz, nsyst_or_2nsyst]; each value
-            # at logk.indices[i] = (norm_pos, syst_pos) corresponds to the
-            # (bin, proc) pair stored at norm.indices[norm_pos]. Gather
-            # rnorm_init through this two-level mapping.
-            rnorm_at_norm = tf.gather_nd(rnorm_init, self.indata.norm.indices)
-            scale_per_logk = tf.gather(rnorm_at_norm, self.indata.logk.indices[:, 0])
-            new_values = self.indata.logk.values * scale_per_logk
-            self.logk = tf.SparseTensor(
-                self.indata.logk.indices,
-                new_values,
-                self.indata.logk.dense_shape,
-            )
-            self.logk_csr = tf_sparse_csr.CSRSparseMatrix(self.logk)
+            # The CSR contraction returns one value per non-zero of norm, so
+            # the factor is gathered onto the same [norm_nnz] layout. This is
+            # much smaller than the per-logk-entry scaling it replaces.
+            self.rnorm_init_at_norm = tf.gather_nd(rnorm_init, self.indata.norm.indices)
         else:
-            # Dense logk: [nbinsfull, nproc, nsyst] symmetric, or
-            # [nbinsfull, nproc, 2, nsyst] asymmetric. Broadcast rnorm_init
-            # over the trailing axes.
-            if self.indata.symmetric_tensor:
-                self.logk = self.indata.logk * rnorm_init[..., None]
-            else:
-                self.logk = self.indata.logk * rnorm_init[..., None, None]
+            # Dense: logsnorm is [nbins, nproc], matching rnorm_init directly
+            # (sliced to nbins when masked channels are excluded).
+            self.rnorm_init = rnorm_init
 
     def _compute_yields_noBBB(self, full=True, compute_norm=True):
         # full: compute yields inclduing masked channels
@@ -1725,6 +1730,11 @@ class Fitter:
                 snormnorm_sparse = snormnorm_sparse * rnorm
             else:  # "normal"
                 # values[i] = norm[i] * rnorm[bin, proc] + logsnorm[i]
+                # rnorm_init factors out of the contraction (see
+                # _init_logk_scaled), so it is applied here rather than
+                # being baked into logk.
+                if self.rnorm_init_at_norm is not None:
+                    logsnorm = logsnorm * self.rnorm_init_at_norm
                 snormnorm_sparse = self.indata.norm * rnorm
                 snormnorm_sparse = snormnorm_sparse.with_values(
                     snormnorm_sparse.values + logsnorm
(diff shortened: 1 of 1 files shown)


davidwalter2:
Fixed and rebased onto main (`2e8c24f`), now at `e94b9f3`.

The P3 was right and the contradiction was self-contained: the call-site comment still described the pre-multiplied scaled copy that this very commit removes, while `_init_logk_scaled`'s own docstring ~1400 lines below already explained the replacement (`rnorm_init` factors out of the contraction, so it is applied to the `[nbins, nproc]` result and `logk` stays aliased). Trimmed the call site to a pointer rather than restating the docstring:

```python
# Param-model scaling of the systematic variations for the
# yield-computation hot path; see _init_logk_scaled.
self._init_logk_scaled()
```

I also checked the other two `_init_logk_scaled` references, at the sparse and dense `"normal"` consumers — both already describe the new behaviour ("applied here rather than being baked into logk") and needed no change. No other mention of a scaled copy survives in the file.

Rebase was clean (single commit, no conflicts) and picks up #166, #167 and the 153/154 work. 152 tests pass on the rebased tree; `black` clean. `isort` flags `fitter.py` locally, but it does so on unmodified main too and this change touches no import lines, so it is pre-existing and not from this PR.

Amended rather than stacked a fixup, since the stale comment was a defect of this commit and the rebase rewrote the branch anyway.

davidwalter2:
Added the regression test this PR was missing — 50f9031.

The refactor was verified by hand against the unpatched code and nothing protected it afterwards, so a reintroduced copy would have been silent. Two properties now pinned:

- **`logk` (and `logk_csr`) are aliased, not copied** — the point of the change, directly assertable, dense and sparse. Reintroducing the copy fails it.
- **`rnorm_init` factors out of the contraction** — the premise that licenses applying it to the `[nbins, nproc]` result, checked against the fitter's own `logk`/`rnorm_init` so a misaligned or wrongly-shaped factor fails there rather than silently broadcasting.

Two things worth being explicit about, since they shaped the test:

**It is not an end-to-end old-vs-new comparison.** The yield path captures `rnorm_init` at construction, so it cannot be swapped out afterwards to run both forms. I tried exactly that first — build one fitter, pre-multiply `logk` and clear `rnorm_init` on a second, compare yields. It passed, and proved nothing: mutating the attribute after construction has no effect on `expected_yield()`, so both sides computed the same thing for the same reason. A deliberate control (assert the yields *differ* when the scaling is dropped) is what caught it; without that control it would have shipped as a green, meaningless test. That control survives in the final version.

**A `Mu` model at the default `mu = 1` cannot see this code path at all.** `rnorm_init` is then exactly 1 and both forms are the identity, so the test uses `expectSignal = 1.7` and asserts a systematic actually reaches a process whose `rnorm_init != 1` — my first attempt put all the systematics on `bkg`, where the factor is 1, and was vacuous for that reason too.

Also registered in the CI matrix. That matters more than it sounds: CI runs named files rather than collecting them, so an unlisted test file never runs. See #173 — 137 of ~152 unit tests on `main` are currently in that state.

lucalavezzo:
Before raising anything I checked the refactor itself rather than reasoning about it: I reimplemented the pre-PR path (pre-multiply `logk`, then contract) in numpy and compared it against `_compute_yields_noBBB()` with **both** the POI and θ moved off their defaults, on the `normal`-type tensor from the new test.

```
[dense ]  max rel diff (new vs OLD pre-multiplied): 0.000e+00
[sparse]  max rel diff (new vs OLD pre-multiplied): 0.000e+00
sparse vs dense yields:                             0.000e+00
```

Bit-identical here only because the tensor has `nsyst=3`; the 1e-12 / 1e-11 in the description is the large-`nsyst` version of the same thing. The part I actually wanted to see was the sparse branch — the old code gathered onto per-`logk`-entry positions and the new one gathers onto the `[norm_nnz]` layout — and that mapping is right.

Two things to raise.

## 1. The `main.yml` hunk removes six test entries against current `main`

The branch is based on `2e8c24f`, which predates the blinding / saturated / ndfsat entries now on `main`. `git diff main <pr-head> -- .github/workflows/main.yml`:

```diff
           - test_global_asym_impacts.py
-          - test_blinding.py
-          - test_blinding_multiplicative.py
-          - test_blinding_noi.py
-          - test_saturated_blinding.py
+          - test_logk_no_copy.py
           - test_multi_systematic.py
-          - test_ndfsat.py
-          - test_saturated_composite.py
           - test_sparse_fit.py
```

GitHub reports the PR as `CONFLICTING` on exactly that file. Given #173 — an unlisted test file runs nowhere, and nothing reports the omission — resolving that conflict by taking the PR side would silently disable six test files and still go green. Worth another rebase that keeps every entry, rather than a hand-resolved merge.

## 2. `rnorm_init` is √μ, not μ

Pre-existing, not introduced here — this PR moves the factor from before the contraction to after it and preserves its value exactly, which is the right thing for a perf change to do. But this PR is the only thing in the repo that deliberately exercises `rnorm_init != 1`, so it is where it becomes visible.

For POIs with `allowNegativeParam=False` (the default) rabbit fits **√μ**, not μ. `set_param_default` stores the square root:

```python
xpoi_part = tf.sqrt(poi_part)      # paramdefault holds mu, xparamdefault holds sqrt(mu)
```

and `get_poi()` undoes it on the way into the model:

```python
poi = tf.square(xpoi)
```

So `param_model.compute()` consumes **physical** values — in the hot path it is always fed `get_poi()`. `_init_logk_scaled` is the one call site that skips the squaring:

```python
rnorm_init = self.param_model.compute(self.param_model.xparamdefault, full=True)
```

`Mu.compute` returns its argument verbatim, so it hands back √μ. With `expectSignal = ("sig", 1.7)`:

```
x-space default   pm.xparamdefault        : [1.30384048]      <- sqrt(1.7)
HOT PATH    compute(get_poi())  -> rnorm  : [1.7  1.0]
_init_logk_scaled           -> rnorm_init : [1.30384048  1.0]
```

Same point in parameter space, two different answers. `914b10b` is titled *"Scale additive systematics by param-model value at linearization point"*, and the param-model value at that point is 1.7 — it is what the yield computation itself uses there. So the additive variations on signal processes end up at `1/√μ = 0.767` of the intended size.

Mirroring `get_poi()` fixes it:

```python
xdef = self.param_model.xparamdefault
if not self.param_model.allowNegativeParam:
    npoi = self.param_model.npoi
    xdef = tf.concat([tf.square(xdef[:npoi]), xdef[npoi:]], axis=0)
rnorm_init = self.param_model.compute(xdef, full=True)
```

Three conditions have to line up for it to bite — `systematic_type == "normal"`, a POI default ≠ 1 (in practice `--expectSignal` ≠ 1), and `allowNegativeParam=False` — which is why it has gone unnoticed: at μ=1, √1 = 1 exactly, and every test in the repo bar the new one runs at μ=1. The ABCD-family models set `xparamdefault` to ones or zeros directly and are unaffected for the same reason.

Worth noting that the new test cannot catch this either: its guard is `assert np.any(touched & (rnorm_init != 1.0))`, and 1.304 satisfies that as happily as 1.7 would. Happy for this to be a separate issue rather than scope creep on a perf PR — flagging it here because this is where the factor is under the microscope.

davidwalter2:
Thanks for reimplementing the old path rather than reasoning about it — and for checking the sparse gather specifically, which was the half I was least able to verify from the algebra alone.

## 1. The `main.yml` hunk — already resolved

The branch has been rebased since you wrote that. Merge-base is now `8898336`, i.e. current `main`, and the hunk is a single addition with all six entries intact:

```diff
           - test_saturated_blinding.py
+          - test_logk_no_copy.py
           - test_multi_systematic.py
```

GitHub reports `MERGEABLE/CLEAN`. Your reasoning for *why* it mattered stands though, and is worth recording against #173: resolving that conflict the wrong way would have disabled six test files and still gone green, because nothing reports an unlisted file. A hand-resolved merge on that particular file is unusually dangerous for exactly that reason.

## 2. `rnorm_init` is √μ — confirmed, and fixed in 3574ac4

Reproduced before changing anything:

```
xparamdefault (x-space)            : [1.30384048]    <- sqrt(1.7)
hot path   compute(get_poi())      : [1.7  1.0]
_init_logk_scaled -> rnorm_init    : [1.30384048  1.0]
ratio init/hot = 0.766965          (1/sqrt(1.7) = 0.766965)
```

Exactly your diagnosis. `compute()` consumes physical values everywhere else, because everywhere else it is handed `get_poi()`; this one call site passed the stored coordinate straight through.

I took it here rather than into a separate PR, and I want to be explicit that this is a scope judgement rather than an oversight. Three reasons: the fix is four lines inside the function this PR rewrites, so a separate PR would conflict with this one immediately and one of them would need rebasing onto the other; this branch is the only thing in the repo that runs with `rnorm_init != 1`, so it is also the only place the fix can be tested today; and the test you correctly identified as unable to catch it is this PR's test, so it needed strengthening here regardless. It is a separate commit, so it can be dropped or split if you would rather. Say the word and I will move it.

On the test: you are right that `rnorm_init != 1` is no guard at all here, since `sqrt(1.7) = 1.304` satisfies it as happily as `1.7` would. Replaced with an agreement check against `compute(get_poi())` — the claim is that two call sites evaluate the model at the same point, not that the factor takes any particular value — plus a guard that the POI default really is off 1, since `sqrt(1) = 1` would make the whole thing vacuous. Reverting the fix makes it fail.

244 tests pass, lint clean.

PR #153, opened 2026-08-26
Native TensorFlow trust-region minimizers (tf-trust-exact, tf-trust-ncg, tf-trust-krylov)

davidwalter2:
## What

Three native TensorFlow implementations of the trust-region minimizers, selectable via `--minimizerMethod`, sharing one outer loop (`rabbit/minimizer/base.py`, mirroring scipy's `_minimize_trust_region` and plugging into the existing callback / early-stopping / restart / `minimizer_result` plumbing unchanged):

- **`tf-trust-exact`** (`exact.py`): Moré–Sorensen, with `H + λI`, the Cholesky factorizations and the triangular solves on the TF device. No custom op for the LAPACK `potrf` failure index: `tf.linalg.cholesky` signals non-PD input by a NaN-filled factor on every backend, and a failed factorization already proves λ is a valid lower bound, so the safeguarded bracket update converges without it. The hard-case machinery stays on device too (inverse-iteration smallest-singular-value estimate) and its acceptance is guarded by an explicit model-value check.
- **`tf-trust-ncg`** (`krylov.py`): Steihaug–Toint CG with the whole inner loop as a single `tf.function` `while_loop` — one graph dispatch per subproblem solve instead of one python round trip (assign, two numpy conversions, forced device sync) per Hessian-vector product. `Hz` is carried through the CG recurrence, so the boundary/negative-curvature exits and the outer loop's model value cost no extra HVPs (scipy spends 2–3 there).
- **`tf-trust-krylov`** (`gltr.py`): GLTR (Gould–Lucidi–Roma–Toint, the trust-krylov algorithm). Lanczos with full CGS2 reorthogonalization on device against a fixed-shape basis buffer; the k×k tridiagonal trust-region subproblems (secular solve, hard case included) on the host where they cost microseconds. The Krylov data is radius-independent, so re-solves after rejected outer steps reuse it — usually zero new HVPs.

The preconditioner gains `tf_transforms()`: graph-side block T/Tᵀ applications, so the reparameterisation runs inside the compiled CG/Lanczos loops (once per HVP, where the numpy path cannot).

## Validation

`tests/test_native_minimizer.py` (33 tests):
- nearly-exact subproblem vs the exact optimum by eigendecomposition across PD/indefinite models and radii (worst case 96.6% of the optimal model reduction vs scipy trust-exact's 96.0%);
- Steihaug-CG steps agree with scipy's `CGSteihaugSubproblem` to float precision (same algorithm, deterministic);
- GLTR reaches ≥ 99.9% of the exact optimum where truncated CG has no guarantee, hard-case KKT check, and a test counting HVPs across a shrunken-radius re-solve;
- full `Fitter` fits for all three methods match the trust-krylov reference, with and without `--precondition`.

## Benchmarks (GTX 1080 Ti, fp64 at 1/32 rate — the least favorable GPU available)

2D Z in-situ fit (24k bins, 3603 nuisances), identical conditions, identical minimum 11249.6383:
- `tf-trust-krylov` **1043 s** vs scipy `trust-krylov` **1435 s** (93 vs 96 outer iterations — the win is per-iteration subproblem cost, largest in the endgame where scipy's iterations reach ~200 s vs 67–102 s native).

Synthetic (analytic-Hessian objective, n=2000/4000): `tf-trust-exact` 5.7 s / 16.6 s vs scipy trust-exact 7.7 s / 24.7 s, identical minima; raw fp64 Cholesky on device 2–3× faster than 8-core LAPACK. On CPU-only nodes the native paths are slower than scipy (TF's CPU Cholesky kernel is single-threaded Eigen) — a warning is logged and the scipy methods remain the default; nothing changes unless a `tf-` method is selected.

Multi-GPU support is intentionally **not** part of this PR; it will come separately.

https://claude.ai/code/session_01VCVDNM26GpDMggr8GDFiwY

### DIFF
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -14,11 +14,7 @@ from rabbit import external_likelihood, io_tools
 from rabbit import preconditioner as precond
 from rabbit import tfhelpers as tfh
 from rabbit.bbstat.bbstat import BinByBinStat
-from rabbit.callbacks import (
-    RESTART_MIN_IMPROVEMENT,
-    FitterCallback,
-    merge_callbacks,
-)
+from rabbit.callbacks import RESTART_MIN_IMPROVEMENT, FitterCallback, merge_callbacks
 from rabbit.impacts import (
     asym_impacts,
     global_asym_impacts,
@@ -26,6 +22,7 @@ from rabbit.impacts import (
     nonprofiled_impacts,
     traditional_impacts,
 )
+from rabbit.minimizer import minimize_trust_exact
 from rabbit.tfhelpers import edmval_cov
 
 logger = logging.child_logger(__name__)
@@ -2383,6 +2380,29 @@ class Fitter:
                 logger.info(f"  - edmval: {edmval}")
             return pc.hess_to_internal(hess.__array__())
 
+        # Native (TF) minimizer counterparts of the callbacks above. Same
+        # contract and the same internal coordinates, but the gradient and
+        # Hessian stay tf tensors: with preconditioning off they never leave
+        # the device, and the subproblem factorizes there either way.
+        def native_loss(yval):
+            pc = pc_cell[0]
+            self.x.assign(pc.to_physical(yval))
+            return float(self.loss_val())
+
+        def native_closure(yval):
+            pc = pc_cell[0]
+            self.x.assign(pc.to_physical(yval))
+            val, grad, hess = self.loss_val_grad_hess()
+            if self.diagnostics:
+                cond_number = tfh.cond_number(hess)
+                logger.info(f"  - Condition number: {cond_number}")
+                edmval = tfh.edmval(grad, hess)
+                logger.info(f"  - edmval: {edmval}")
+            if pc.enabled:
+                grad = tf.constant(pc.grad_to_internal(grad.__array__()))
+                hess = tf.constant(pc.hess_to_internal(hess.__array__()))
+            return float(val), grad, hess
+
         # scipy works in internal coordinates throughout; y = 0 at the point the
         # transform was built.
         xval = pc_cell[0].from_physical(self.x.numpy())
@@ -2430,16 +2450,26 @@ class Fitter:
         while True:
             cb = FitterCallback(xval, self.earlyStopping)
             try:
-                res = scipy.optimize.minimize(
-                    scipy_loss,
-                    xval,
-                    method=self.minimizer_method,
-                    jac=True,
-                    tol=0.0,
-                    callback=cb,
-                    options=sci_opts,
-                    **info_minimize,
-                )
+                if self.minimizer_method == "tf-trust-exact":
+                    res = minimize_trust_exact(
+                        native_loss,
+                        native_closure,
+                        xval,
+                        gtol=sci_opts.get("gtol", 0.0),
+                        maxiter=sci_opts.get("maxiter"),
+                        callback=cb,
+                    )
+                else:
+                    res = scipy.optimize.minimize(
+                        scipy_loss,
+                        xval,
+                        method=self.minimizer_method,
+                        jac=True,
+                        tol=0.0,
+                        callback=cb,
+                        options=sci_opts,
+                        **info_minimize,
+                    )
             except Exception as ex:
                 # minimizer could have called the loss or hessp functions with "random" values, so restore the
                 # state from the end of the last iteration before the exception
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -22,7 +22,7 @@ from rabbit.impacts import (
     nonprofiled_impacts,
     traditional_impacts,
 )
-from rabbit.minimizer import minimize_trust_exact
+from rabbit.minimizer import minimize_trust_exact, minimize_trust_ncg
 from rabbit.tfhelpers import edmval_cov
 
 logger = logging.child_logger(__name__)
@@ -2403,6 +2403,39 @@ class Fitter:
                 hess = tf.constant(pc.hess_to_internal(hess.__array__()))
             return float(val), grad, hess
 
+        def native_grad_closure(yval):
+            pc = pc_cell[0]
+            self.x.assign(pc.to_physical(yval))
+            val, grad = self.loss_val_grad()
+            if pc.enabled:
+                grad = tf.constant(pc.grad_to_internal(grad.__array__()))
+            return float(val), grad
+
+        def native_set_point(yval):
+            pc = pc_cell[0]
+            self.x.assign(pc.to_physical(yval))
+
+        def native_hessp():
+            # internal-coordinate HVP, graph-compatible: the pc transform runs
+            # inside the compiled CG loop (numpy per CG iteration would defeat
+            # the on-device solve). Rebuilt per restart along with pc.
+            pc = pc_cell[0]
+            transforms = pc.tf_transforms()
+            if transforms is None:
+
+                def hessp(v):
+                    _, _, hp = self.loss_val_grad_hessp(v)
+                    return hp
+
+            else:
+                apply_T, apply_TT = transforms
+
+                def hessp(v):
+                    _, _, hp = self.loss_val_grad_hessp(apply_T(v))
+                    return apply_TT(hp)
+
+            return hessp
+
         # scipy works in internal coordinates throughout; y = 0 at the point the
         # transform was built.
         xval = pc_cell[0].from_physical(self.x.numpy())
@@ -2454,6 +2487,17 @@ class Fitter:
                     res = minimize_trust_exact(
                         native_loss,
                         native_closure,
+                        xval,
+                        gtol=sci_opts.get("gtol", 0.0),
+                        maxiter=sci_opts.get("maxiter"),
+                        callback=cb,
+                    )
+                elif self.minimizer_method == "tf-trust-ncg":
+                    res = minimize_trust_ncg(
+                        native_loss,
+                        native_grad_closure,
+                        native_hessp(),
+                        native_set_point,
                         xval,
                         gtol=sci_opts.get("gtol", 0.0),
                         maxiter=sci_opts.get("maxiter"),
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -22,7 +22,11 @@ from rabbit.impacts import (
     nonprofiled_impacts,
     traditional_impacts,
 )
-from rabbit.minimizer import minimize_trust_exact, minimize_trust_ncg
+from rabbit.minimizer import (
+    minimize_trust_exact,
+    minimize_trust_krylov,
+    minimize_trust_ncg,
+)
 from rabbit.tfhelpers import edmval_cov
 
 logger = logging.child_logger(__name__)
@@ -2494,6 +2498,17 @@ class Fitter:
                     )
                 elif self.minimizer_method == "tf-trust-ncg":
                     res = minimize_trust_ncg(
+                        native_loss,
+                        native_grad_closure,
+                        native_hessp(),
+                        native_set_point,
+                        xval,
+                        gtol=sci_opts.get("gtol", 0.0),
+                        maxiter=sci_opts.get("maxiter"),
+                        callback=cb,
+                    )
+                elif self.minimizer_method == "tf-trust-krylov":
+                    res = minimize_trust_krylov(
                         native_loss,
                         native_grad_closure,
                         native_hessp(),
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -67,6 +67,16 @@ def match_regexp_params(regular_expressions, parameter_names):
     return matched
 
 
+# Options from --minimizerMaxiter/--minimizerGtol/--minimizerFtol that each
+# native minimizer actually reads. Anything else passed for these methods is
+# warned about rather than silently dropped (see Fitter.fit).
+NATIVE_MINIMIZER_OPTIONS = {
+    "tf-trust-exact": {"gtol", "maxiter"},
+    "tf-trust-ncg": {"gtol", "maxiter"},
+    "tf-trust-krylov": {"gtol", "maxiter"},
+}
+
+
 class Fitter:
     valid_systematic_types = ["log_normal", "normal"]
 
@@ -2477,6 +2487,24 @@ class Fitter:
         if self.minimizer_ftol is not None:
             sci_opts["ftol"] = float(self.minimizer_ftol)
         logger.info(f"[minimize] method={self.minimizer_method} options={sci_opts}")
+
+        # The native minimizers take gtol and maxiter and nothing else, so any
+        # other key here is silently dropped -- scipy at least raises an
+        # OptimizeWarning for an option its method does not recognize, and
+        # without this the user sees the option echoed in the line above and
+        # then has no signal that it did nothing. --minimizerFtol is the one
+        # that reaches this today.
+        if self.minimizer_method in NATIVE_MINIMIZER_OPTIONS:
+            ignored = sorted(
+                set(sci_opts) - NATIVE_MINIMIZER_OPTIONS[self.minimizer_method]
+            )
+            if ignored:
+                logger.warning(
+                    f"{self.minimizer_method} does not implement "
+                    f"{', '.join(ignored)}; ignoring "
+                    f"{', '.join(f'--minimizer{o.capitalize()}' for o in ignored)}. "
+                    "Use a scipy --minimizerMethod if you need it."
+                )
 
         # Restart loop. scipy's trust-region methods shrink the trust radius by
         # 4x on every rejected step with no lower bound, and the radius is a
--- a/rabbit/minimizer/exact.py
+++ b/rabbit/minimizer/exact.py
@@ -1,0 +1,400 @@
+"""Nearly-exact trust-region subproblem in TensorFlow.
+
+This is a port of scipy's ``IterativeSubproblem``
+(``scipy/optimize/_trustregion_exact.py``, the More-Sorensen algorithm of
+[3]_ as described in [1]_ ch. 7.3) with the linear algebra kept on the
+TensorFlow device: building H + lambda*I, the Cholesky factorizations and
+the triangular solves all run where the Hessian already lives, so the
+n x n matrix never has to cross to the host per lambda trial. Only O(n)
+vectors and control-flow scalars are synchronized.
+
+The one structural difference from scipy: LAPACK's ``potrf`` reports the
+index k of the first non-positive-definite leading minor on failure, which
+scipy feeds to ``singular_leading_submatrix`` to tighten ``lambda_lb``
+below the critical damping. ``tf.linalg.cholesky`` reports no such index
+(it fills the factor with NaNs instead of raising, on CPU, GPU and under
+XLA alike), and getting k out of TF would need a custom op. We don't need
+it: a failed factorization of H + lambda*I proves lambda < -lambda_min(H),
+so ``lambda_current`` itself is a valid lower bound, and the standard
+safeguarded update max(sqrt(lb*ub), lb + theta*(ub - lb)) still converges,
+just without the accelerated bound. The price is a few extra
+factorizations per solve in the indefinite case -- each of which is the
+thing this port makes cheap. With the fit preconditioner active the
+internal-coordinate Hessian is near the identity and this branch is
+essentially never taken.
+
+The rarely-hit "hard case" refinement (interior solution with lambda > 0)
+needs the smallest-singular-value estimate of the factor; that runs on the
+host exactly as in scipy, paying one factor download when triggered.
+
+References
+----------
+.. [1] A.R. Conn, N.I. Gould, and P.L. Toint, "Trust region methods",
+       Siam, pp. 169-200, 2000.
+.. [2] J. Nocedal and S. Wright, "Numerical optimization",
+       Springer, pp. 83-91, 2006.
+.. [3] J.J. More and D.C. Sorensen, "Computing a trust region step",
+       SIAM J. Sci. Stat. Comput., vol. 4(3), pp. 553-572, 1983.
+"""
+
+import math
+
+import numpy as np
+import scipy.linalg
+import tensorflow as tf
+from wums import logging
+
+logger = logging.child_logger(__name__)
+
+
+def gershgorin_bounds(hess):
+    """Lower and upper bounds on the eigenvalues of ``hess`` ([1]_ p. 19).
+
+    Runs on device; returns python floats.
+    """
+    h_diag = tf.linalg.diag_part(hess)
+    h_diag_abs = tf.abs(h_diag)
+    row_sums = tf.reduce_sum(tf.abs(hess), axis=1)
+    lb = tf.reduce_min(h_diag + h_diag_abs - row_sums)
+    ub = tf.reduce_max(h_diag - h_diag_abs + row_sums)
+    return float(lb), float(ub)
+
+
+def estimate_smallest_singular_value(L):
+    """Estimate the smallest singular value/vector of lower-triangular ``L``.
+
+    Direct port of scipy's version (Cline et al. 1979), which works on the
+    upper factor U; here U = L^T so U^T = L and the recurrence reads off
+    columns of L. O(n^2), host-side numpy: it is inherently sequential and
+    only runs in the rare hard case.
+    """
+    L = np.atleast_2d(L)
+    n = L.shape[0]
+
+    p = np.zeros(n)
+    w = np.empty(n)
+
+    for k in range(n):
+        wp = (1 - p[k]) / L[k, k]
+        wm = (-1 - p[k]) / L[k, k]
+        pp = p[k + 1 :] + L[k + 1 :, k] * wp
+        pm = p[k + 1 :] + L[k + 1 :, k] * wm
+
+        if abs(wp) + np.linalg.norm(pp, 1) >= abs(wm) + np.linalg.norm(pm, 1):
+            w[k] = wp
+            p[k + 1 :] = pp
+        else:
+            w[k] = wm
+            p[k + 1 :] = pm
+
+    # w solves L w = e (e in {+1,-1}^n chosen for growth); now L^T v = w
+    v = scipy.linalg.solve_triangular(L, w, lower=True, trans="T")
+    v_norm = np.linalg.norm(v)
+
+    s_min = np.linalg.norm(w) / v_norm
+    z_min = v / v_norm
+    return s_min, z_min
+
+
+def get_boundaries_intersections(z, d, trust_radius):
+    """Solve ||z + t d|| == trust_radius for t; return [t_low, t_high]."""
+    a = float(np.dot(d, d))
+    b = 2 * float(np.dot(z, d))
+    c = float(np.dot(z, z)) - trust_radius**2
+    sqrt_discriminant = math.sqrt(b * b - 4 * a * c)
+
+    # numerically stable form (avoids cancellation), as in scipy
+    aux = b + math.copysign(sqrt_discriminant, b)
+    ta = -aux / (2 * a)
+    tb = -2 * c / aux
+    return sorted([ta, tb])
+
+
+class IterativeSubproblem:
+    """Quadratic subproblem solved by the nearly-exact iterative method.
+
+    Constructed from the objective value, gradient and dense Hessian at the
+    current point (gradient and Hessian as tf tensors on device). One
+    instance corresponds to one linearization point; ``solve`` may be
+    called repeatedly with shrinking radii (rejected outer steps) and
+    warm-starts its lambda search from the previous call, as in scipy.
+    """
+
+    # "theta" of [1]_ formula 7.3.14 (p. 190)
+    UPDATE_COEFF = 0.01
+
+    EPS = np.finfo(np.float64).eps
+
+    # scipy caps the lambda search at 25 (gh-12513); ours approaches the
+    # critical damping without the potrf-index acceleration, so allow more --
+    # the factorizations are the operation this port makes cheap
+    MAXITER_DEFAULT = 50
+
+    def __init__(self, fun_val, jac, hess, k_easy=0.1, k_hard=0.2, maxiter=None):
+        self.fun = float(fun_val)
+        self.jac = tf.convert_to_tensor(jac)
+        self.hess = tf.convert_to_tensor(hess)
+
+        self.jac_mag = float(tf.norm(self.jac))
+
+        # lambda-search warm start across solve() calls at this same point:
+        # when the trust radius shrinks, the previous lower bound is reusable
+        self.previous_tr_radius = -1.0
+        self.lambda_lb = None
+
+        # stop-criteria parameters, [1]_ pp. 194-197
+        self.k_easy = k_easy
+        self.k_hard = k_hard
+
+        self.maxiter = self.MAXITER_DEFAULT if maxiter is None else maxiter
+
+        self.dimension = int(self.hess.shape[0])
+        self._eye = tf.eye(self.dimension, dtype=self.hess.dtype)
+        self.hess_gersh_lb, self.hess_gersh_ub = gershgorin_bounds(self.hess)
+        # NB axis=[-2, -1] requests the *matrix* norms; tf.norm's default
+        # axis=None flattens the tensor, and the resulting max|H_ij| can sit
+        # below |lambda_min|, silently invalidating the lambda_ub bracket
+        self.hess_inf = float(tf.norm(self.hess, ord=np.inf, axis=[-2, -1]))
+        self.hess_fro = float(tf.norm(self.hess, ord="fro", axis=[-2, -1]))
+        self.CLOSE_TO_ZERO = self.dimension * self.EPS * self.hess_inf
+
+    # --- model evaluation -------------------------------------------------
+
+    def model_value(self, p):
+        """m(p) = f + g.p + p.H p / 2 for a step ``p`` (numpy or tensor)."""
+        p = tf.convert_to_tensor(p, dtype=self.hess.dtype)
+        quad = tf.tensordot(p, tf.linalg.matvec(self.hess, p), axes=1)
+        lin = tf.tensordot(self.jac, p, axes=1)
+        return self.fun + float(lin) + 0.5 * float(quad)
+
+    # --- linear algebra helpers ------------------------------------------
+
+    def _factorize(self, lambda_current):
+        """Cholesky of H + lambda*I. Returns (L, ok).
+
+        tf.linalg.cholesky signals a non-positive-definite input by filling
+        the factor with NaNs (never raising) on every backend, so success is
+        a NaN check -- one scalar readback.
+        """
+        H = self.hess + lambda_current * self._eye
+        L = tf.linalg.cholesky(H)
+        ok = not bool(tf.reduce_any(tf.math.is_nan(L)))
+        return L, ok
+
+    @staticmethod
+    def _cho_solve(L, b):
+        """Solve (L L^T) x = b for vector b."""
+        return tf.squeeze(
+            tf.linalg.cholesky_solve(L, tf.expand_dims(b, axis=-1)), axis=-1
+        )
+
+    @staticmethod
+    def _tri_solve_t(L, b):
+        """Solve L^T x = b for vector b."""
+        return tf.squeeze(
+            tf.linalg.triangular_solve(
+                L, tf.expand_dims(b, axis=-1), lower=True, adjoint=True
+            ),
+            axis=-1,
+        )
+
+    # --- lambda search ----------------------------------------------------
+
+    def _initial_values(self, tr_radius):
+        """Initial damping factor and bracket, [1]_ sec. 7.3.8 (p. 192)."""
+        # upper bound
+        hess_norm = min(self.hess_fro, self.hess_inf)
+        lambda_ub = self.jac_mag / tr_radius + min(-self.hess_gersh_lb, hess_norm)
+        lambda_ub = max(0.0, lambda_ub)
+
+        # lower bound
+        lambda_lb = self.jac_mag / tr_radius - min(self.hess_gersh_ub, hess_norm)
+        lambda_lb = max(
+            lambda_lb, -float(tf.reduce_min(tf.linalg.diag_part(self.hess)))
+        )
+        lambda_lb = max(0.0, lambda_lb)
+
+        # improve the bracket with the previous solve at this point
+        if tr_radius < self.previous_tr_radius and self.lambda_lb is not None:
+            lambda_lb = max(self.lambda_lb, lambda_lb)
+
+        if lambda_lb == 0.0:
+            lambda_initial = 0.0
+        else:
+            lambda_initial = max(
+                math.sqrt(lambda_lb * lambda_ub),
+                lambda_lb + self.UPDATE_COEFF * (lambda_ub - lambda_lb),
+            )
+        return lambda_initial, lambda_lb, lambda_ub
+
+    def solve(self, tr_radius):
+        """Solve the subproblem for the given radius.
+
+        Returns (p, hits_boundary) with ``p`` a numpy array.
+        """
+        lambda_current, lambda_lb, lambda_ub = self._initial_values(tr_radius)
+        n = self.dimension
+        hits_boundary = True
+        already_factorized = False
+        niter = 0
+
+        p = None
+        L = None
+        factorized_ok = False
+
+        while True:
+            if already_factorized:
+                already_factorized = False
+            else:
+                L, factorized_ok = self._factorize(lambda_current)
+
+            if niter >= self.maxiter:
+                # scipy caps here too (gh-12513). Return the best step
+                # available rather than looping: the last computed p clipped
+                # to the radius, or a boundary Cauchy step along -g.
+                logger.warning(
+                    f"trust-region subproblem lambda search hit maxiter="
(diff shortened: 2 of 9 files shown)


rabbit/minimizer/exact.py
```diff
@@ -0,0 +1,448 @@
...
+                p = self._cho_solve(L, -self.jac)
+                p_norm = float(tf.norm(p))
+
+                # interior convergence
+                if p_norm <= tr_radius and lambda_current == 0.0:
+                    hits_boundary = False
+                    break
+
+                # Newton step on the secular equation, [2]_ (4.44) p. 87
+                w = self._tri_solve_t(L, p)
```
lucalavezzo: **Wrong triangular solve in the Moré–Sorensen Newton step** — this is the one I'd fix before merge.

`_tri_solve_t` solves `Lᵀx = p`, i.e. `w = L⁻ᵀp`. Formula (4.44) / N&W Alg. 4.3 needs `w = L⁻¹p`. scipy computes `solve_triangular(U, p, trans='T')` with `B = UᵀU`; since `U = Lᵀ`, scipy's `w` is `U⁻ᵀp = L⁻¹p`. The adjoint got flipped in the port.

Checked numerically (n=40, eigenvalues `logspace(-6,3)`, cond 1e9), against a finite difference of the secular equation:

| λ | `d‖p‖/dλ` (FD) | `-‖L⁻¹p‖²/‖p‖` | `-‖L⁻ᵀp‖²/‖p‖` | step too large by |
|---|---|---|---|---|
| 1e-4 | -2.494135e+08 | -2.494126e+08 ✓ | -6.161462e+07 ✗ | 4.05× |
| 1e-2 | -4.696568e+04 | -4.696567e+04 ✓ | -1.736449e+04 ✗ | 2.71× |
| 1    | -6.100695e+00 | -6.100695e+00 ✓ | -4.668389e+00 ✗ | 1.31× |

The sign is still carried by `(p_norm - tr_radius)`, so the safeguarded bracket keeps this correct-but-slow rather than divergent. But the overshoot grows with condition number, and in the `else:  # outside the boundary` branch below it's taken verbatim (`lambda_current = lambda_new`), so it can jump past `lambda_ub`, thrash the bracket, burn the 50-factorization budget and drop into the safeguarded/Cauchy fallback — i.e. worse steps than scipy `trust-exact` exactly in the ill-conditioned regime this port exists for.

Minimal fix (or add a `_tri_solve` with `adjoint=False` next to `_tri_solve_t`, which reads better):

```suggestion
                w = tf.squeeze(
                    tf.linalg.triangular_solve(
                        L, tf.expand_dims(p, axis=-1), lower=True
                    ),
                    axis=-1,
                )
```

rabbit/minimizer/exact.py
```diff
@@ -0,0 +1,448 @@
...
+        return lambda_initial, lambda_lb, lambda_ub
+
+    def solve(self, tr_radius):
+        """Solve the subproblem for the given radius.
+
+        Returns (p, hits_boundary) with ``p`` a numpy array.
+        """
+        lambda_current, lambda_lb, lambda_ub = self._initial_values(tr_radius)
+        n = self.dimension
+        hits_boundary = True
```
lucalavezzo, line 239: **`hits_boundary` can report `True` for a strictly interior step.**

It's initialised `True` here and the `niter >= self.maxiter` branch below `break`s without ever clearing it. Both fallback candidates can be interior: the clipped `p` when `‖p‖ < tr_radius`, and the Cauchy step when `t = ‖g‖²/gᵀHg < tr_radius/‖g‖`.

The outer loop then acts on the false claim — `base.py::_minimize_trust_region`:

```python
elif rho > 0.75 and hits_boundary:
    trust_radius = min(2 * trust_radius, max_trust_radius)
```

so the radius **doubles right after a λ-search failure**, making the next subproblem harder and the next failure more likely. Setting `hits_boundary = np.linalg.norm(p) >= tr_radius * (1 - 1e-12)` on the chosen candidate before the `break` would do it.

rabbit/minimizer/exact.py
```diff
@@ -0,0 +1,448 @@
...
+                        tf.tensordot(
+                            self.jac, tf.linalg.matvec(self.hess, self.jac), axes=1
+                        )
+                    )
+                    if gHg <= 0:
+                        t = tr_radius / g_norm
+                    else:
+                        t = min(tr_radius / g_norm, g_norm**2 / gHg)
+                    candidates.append(-t * g)
+                p = min(candidates, key=self.model_value)
```
lucalavezzo, line 300: **`ValueError: min() arg is an empty sequence` here aborts the fit.**

At an exact saddle (`grad == 0`, indefinite `H`), `candidates` can be empty:
- `jac_mag > self.CLOSE_TO_ZERO` is False, so the general branch never runs and `p` stays `None` — hence no clipped-`p` candidate;
- the `elif factorized_ok` branch only assigns `p` on a path that immediately `break`s, so if the `k_hard` test never passes within the 50 iterations `p` is still `None`;
- `g_norm > 0` is False, so no Cauchy candidate is appended.

`min([])` then raises, and `base.py`'s `except (np.linalg.LinAlgError, ValueError)` turns it into `warnflag=3` → the fit terminates with `status=3` ("A linalg error occurred") **at the starting point**, instead of stepping off the saddle. A guard for the empty case (e.g. fall back to a small step along the estimated bottom eigenvector, or `p = zeros` with `hits_boundary=False`) keeps it recoverable.

rabbit/minimizer/exact.py
```diff
@@ -0,0 +1,448 @@
...
+        self.lambda_lb = None
+
+        # stop-criteria parameters, [1]_ pp. 194-197
+        self.k_easy = k_easy
+        self.k_hard = k_hard
+
+        self.maxiter = self.MAXITER_DEFAULT if maxiter is None else maxiter
+
+        self.dimension = int(self.hess.shape[0])
+        self._eye = tf.eye(self.dimension, dtype=self.hess.dtype)
```
lucalavezzo: **Dense `n×n` identity allocated per subproblem — i.e. per accepted outer step.**

A new `IterativeSubproblem` is constructed on every accepted step in `_minimize_trust_region`, each allocating an `n×n` float64 eye: 200 MB at n=5000, **1.4 GB at n=13000** — coexisting with the previous subproblem's copy plus the Hessian itself until GC. That's a plausible OOM on exactly the GPU this path targets. `_factorize` also pays a full dense multiply-add per λ trial to build `H + λI`.

`self.hess_diag = tf.linalg.diag_part(self.hess)` cached once (it's already needed just below) plus

```python
tf.linalg.set_diag(self.hess, self.hess_diag + lambda_current)
```

removes the eye entirely and drops one `n²` kernel per λ trial.

rabbit/minimizer/exact.py
```diff
@@ -0,0 +1,448 @@
...
+    """
+    h_diag = tf.linalg.diag_part(hess)
+    h_diag_abs = tf.abs(h_diag)
+    row_sums = tf.reduce_sum(tf.abs(hess), axis=1)
+    lb = tf.reduce_min(h_diag + h_diag_abs - row_sums)
+    ub = tf.reduce_max(h_diag - h_diag_abs + row_sums)
+    return float(lb), float(ub)
+
+
+def estimate_smallest_singular_value(L):
```
lucalavezzo: **Dead code:** a verbatim re-implementation of `scipy.optimize._trustregion_exact.estimate_smallest_singular_value` that nothing calls and no test covers.

Grep over `rabbit/` and `tests/` finds the name only at this definition and in the docstring of `estimate_smallest_singular_value_device`; both call sites in `solve()` use the device version. These 33 lines of sequential Cline-et-al. recurrence are also the sole consumer of `import scipy.linalg` in this module.

The tests already import scipy's private trust-region internals (`from scipy.optimize._trustregion_exact import IterativeSubproblem as ScipySubproblem`), so if a host-side reference is wanted, importing scipy's beats maintaining a copy.

rabbit/minimizer/gltr.py
```diff
@@ -0,0 +1,293 @@
...
+            kmax = min(kmax, self._cg_maxiter)
+
+        if not self._started:
+            self._solver.reset()
+            self._q_prev = tf.zeros_like(self.jac)
+            self._q = self.jac / self.jac_mag
+            self._started = True
+
+        h = lam = hits_boundary = None
+        can_extend = True
```
lucalavezzo: **`can_extend` is per-call state where it should be instance state.**

`_started`, `_deltas`, `_gammas` and `_q` all persist across `solve()` calls — that's the whole point of the rejected-step reuse — but `can_extend` is reset to `True` here on every call. So a subproblem that already hit Lanczos breakdown will call `_extend()` again on a re-solve at a shrunken radius.

Usually the residual test `self._gammas[k-1]*abs(h[-1]) <= tol_eff` saves it. But with a small gradient (`jac_mag=1e-6` → `tol_eff = 1e-10`) and large Hessian entries, the breakdown γ (up to `1e-14*abs(delta)`) times `abs(h[-1]) ~ Delta` can exceed `tol_eff` — and `_extend()` then runs on the post-breakdown `self._q`, which is roundoff noise normalised to unit length rather than a genuine Lanczos direction. Store it next to `_started`.

rabbit/minimizer/gltr.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+
+    def __init__(self, hessp_fn, kmax=None):
+        self.hessp_fn = hessp_fn
+        self.kmax = kmax
+        self.Q = None
+        self._step = tf.function(self._step_impl)
+
+    def ensure(self, n, dtype):
+        kmax = self.kmax or min(n, 1000)
```
lucalavezzo: **Falsy-zero `or` default:** an explicit `kmax=0` silently becomes `min(n, 1000)`.

`GLTRSolver(hessp, kmax=0)` — or a value computed as `max(0, budget - used)` — asks for no inner iterations and gets an unbounded inner solve instead. Same pattern in `GLTRSubproblem.solve` and at `krylov.py:194` (`maxiter = self._cg_maxiter or int(self.jac.shape[0])`).

```suggestion
        kmax = min(n, 1000) if self.kmax is None else self.kmax
```

rabbit/minimizer/krylov.py
```diff
@@ -0,0 +1,221 @@
...
+        self._cg_maxiter = cg_maxiter
+        self._last_p = None
+        self._last_model = None
+        self.niter = 0
+
+    def solve(self, tr_radius):
+        # scipy's forcing sequence: superlinear local convergence
+        tolerance = min(0.5, math.sqrt(self.jac_mag)) * self.jac_mag
+
+        if self.jac_mag < tolerance:  # only at an exactly-zero gradient
```
lucalavezzo: **This guard is unreachable, and the comment asserts the opposite.**

`tolerance = min(0.5, sqrt(jac_mag)) * jac_mag <= 0.5 * jac_mag`, so `jac_mag < tolerance` is False for every `jac_mag >= 0` — including exactly 0, since `0 < 0` is False. The comment claims it fires at a zero gradient; it does not.

With `jac_mag == 0` control falls into the CG graph with `r = d = 0`, takes the `dBd <= 0` negative-curvature exit with `a=0`, `c=-Delta²`, and returns `p = 0` with **`hits_boundary=True`** — a boundary claim for a zero step, which then feeds the radius-doubling test in the outer loop.

`GLTRSubproblem.solve` gets this right with an explicit `if self.jac_mag == 0.0`. For the record scipy's own `CGSteihaugSubproblem` has the identical dead check, so this is a faithful port of scipy's bug — but the comment is new and actively wrong.

```suggestion
        if self.jac_mag == 0.0:  # exactly-zero gradient: nothing to do
```

rabbit/fitter.py
```diff
@@ -2389,6 +2390,62 @@ def scipy_hess(yval):
...
+            self.x.assign(pc.to_physical(yval))
+            val, grad, hess = self.loss_val_grad_hess()
+            if self.diagnostics:
+                cond_number = tfh.cond_number(hess)
+                logger.info(f"  - Condition number: {cond_number}")
+                edmval = tfh.edmval(grad, hess)
+                logger.info(f"  - edmval: {edmval}")
+            if pc.enabled:
+                grad = tf.constant(pc.grad_to_internal(grad.__array__()))
+                hess = tf.constant(pc.hess_to_internal(hess.__array__()))
```
lucalavezzo, line 2423: **With `--precondition` this round-trips the full `n×n` Hessian device → host → device on every accepted step** — which is precisely the transfer the native minimizer exists to eliminate, and it's *two* transfers where the scipy path pays one.

`hess.__array__()` forces a device→host copy of `[npar, npar]` float64, then a numpy block transform, then `tf.constant` copies it back. At npar=5000 that's 200 MB each way, per accepted outer iteration. `exact.py`'s docstring claim that "the `n×n` matrix never has to cross to the host" holds for the λ trials, but the stated win is lost in exactly the configuration the preconditioner was added for.

The fix already exists in this PR: `Preconditioner.tf_transforms()` gives device-side `T`/`Tᵀ`, so `H_y = TᵀHT` can be applied on device (per-block gather + two matmuls) and the Hessian never leaves it.

rabbit/fitter.py
```diff
@@ -2451,16 +2508,48 @@ def scipy_hess(yval):
...
-                        options=sci_opts,
-                        **info_minimize,
-                    )
+                    if self.minimizer_method == "tf-trust-exact":
+                        res = minimize_trust_exact(
+                            native_loss,
+                            native_closure,
+                            xval,
+                            gtol=sci_opts.get("gtol", 0.0),
+                            maxiter=sci_opts.get("maxiter"),
```
lucalavezzo, line 2545: **`--minimizerFtol` is silently dropped for all three `tf-` methods.**

`minimize_trust_exact/ncg/krylov` are called with only `gtol` and `maxiter`; `sci_opts["ftol"]` (set at line 2478) is never read. So `rabbit_fit.py --minimizerMethod tf-trust-exact --minimizerFtol 1e-8` logs

```
[minimize] method=tf-trust-exact options={'ftol': 1e-08}
```

and then ignores it — whereas the scipy path at least raises an `OptimizeWarning` for an unrecognised option, so the user gets no signal either way here.

Either warn on unconsumed `sci_opts` keys for the native methods, or wire `ftol` into the `predicted_reduction` / loss-change termination test in `base.py`.

lucalavezzo (commented):
Reviewed the three native minimizers against scipy's originals and the underlying algorithms (Moré–Sorensen, Steihaug–Toint, GLTR). The architecture is sound — one shared outer loop mirroring scipy's `_minimize_trust_region`, existing callback/early-stopping/restart plumbing untouched, scipy still the default with a CPU warning — and carrying `Hz` through the CG recurrence to save the 2–3 HVPs scipy spends on the exit tests, plus the radius-independent Krylov reuse in GLTR, are both genuinely good calls. Lint is clean (container `black`/`isort`/CI `flake8` on all changed files) and `[tool.setuptools.packages.find]` auto-discovers `rabbit.minimizer`, so no packaging gap.

**One thing I'd fix before merge:** `exact.py:333` uses `L⁻ᵀp` where the Moré–Sorensen Newton step on the secular equation needs `L⁻¹p`. Confirmed against both scipy's source and a finite difference of `d‖p‖/dλ`; the resulting λ step is 1.3–4× too large at cond(H)=1e9. Details inline.

Two more real defects in the same `solve()` — `hits_boundary` reporting `True` for interior fallback steps (which makes the outer loop *double* the radius after a λ-search failure), and `min()` on an empty candidate list aborting the fit at a saddle. Everything in `exact.py`, note; the two Krylov paths carry only the lighter findings.

Two tests don't test what they claim: `test_gltr_warm_restart_reuses_krylov_data` counts `tf.function` traces rather than HVPs, so it stays green even with the reuse guard deleted; `test_gltr_hard_case` puts g *on* the bottom eigenvector and never enters the hard-case branch — which I verified separately is itself correct, so that one is coverage, not a bug.

The rest is performance and cleanup: the `--precondition` path round-trips the full `n×n` Hessian host↔device every accepted step (the transfer this PR exists to remove, and `tf_transforms()` in this same PR is the fix), a per-subproblem `tf.eye(n)` (1.4 GB at n=13000), the per-block scatter inside the compiled HVP, two falsy-zero `or` defaults, a dead `estimate_smallest_singular_value`, and `--minimizerFtol` silently ignored by all three `tf-` methods.

rabbit/fitter.py
```diff
@@ -2451,16 +2508,48 @@ def scipy_hess(yval):
...
-                        scipy_loss,
-                        xval,
-                        method=self.minimizer_method,
-                        jac=True,
-                        tol=0.0,
-                        callback=cb,
-                        options=sci_opts,
-                        **info_minimize,
-                    )
+                    if self.minimizer_method == "tf-trust-exact":
```
lucalavezzo, line 2539: **Cleanup:** the three `tf-` branches are near-identical 10-line call blocks, and the five `native_*` closures above all open with the same two-line pin

```python
pc = pc_cell[0]
self.x.assign(pc.to_physical(yval))
```

while `native_closure` also duplicates the `cond_number`/`edmval` diagnostics block from `scipy_hess` verbatim. This `if/elif/elif` chain differs only in the callable and in whether `hessp`/`set_point` are passed, so each minimizer's argument order now lives in the fitter — adding a fourth native method, or changing one signature, means editing three places here.

A `_NATIVE_MINIMIZERS` table (or a single `rabbit.minimizer.minimize_native(method, ...)` entry point taking the closures as a bundle) plus one `_pin(yval)` helper collapses ~60 lines and keeps the argument order next to the implementations.

rabbit/preconditioner.py
```diff
@@ -313,6 +313,53 @@ def _apply_TT(self, v):
...
+                if linv is not None:
+                    new = tf.linalg.matvec(linv, sub, transpose_a=transpose)
+                else:
+                    new = tf.squeeze(
+                        tf.linalg.triangular_solve(
+                            chol, sub[:, None], lower=True, adjoint=transpose
+                        ),
+                        axis=-1,
+                    )
+                out = tf.tensor_scatter_nd_update(out, sidx, new)
```
lucalavezzo, line 352: **One full-vector scatter per block, inside the compiled loop and twice per HVP** — so cost scales with the number of blocks.

`tensor_scatter_nd_update` is not in-place in a graph: it materialises a new length-`n` tensor per block. This module's own docstring reports that `auto_blocks` "recovers 239 components" on the in-situ efficiency fit, so a single HVP inside `SteihaugCGSolver._solve_impl` costs **2 × 239 = 478 full-vector copies plus 478 gathers**, and the traced graph carries 239 captured `tf.constant` block matrices.

Concatenating all block indices once at construction and doing a single gather → single batched block-diagonal matvec → single scatter makes it 2 copies per HVP instead of 2·B.

tests/test_native_minimizer.py
```diff
@@ -0,0 +1,583 @@
...
+
+    solver = GLTRSolver(hessp)
+    m = GLTRSubproblem(0.0, tf.constant(g, tf.float64), np.zeros(30), solver)
+    p1, _ = m.solve(1.0)
+    n_first = count[0]
+    assert n_first > 0
+
+    p2, hb2 = m.solve(0.25)  # shrunken radius, same point
+    n_second = count[0] - n_first
+    assert n_second <= 2  # essentially free re-solve
```
lucalavezzo: **This test counts `tf.function` traces, not Hessian-vector products, so both assertions pass whether or not the Krylov data is reused.**

`hessp` increments `count[0]` in its python body, but that body only executes while `GLTRSolver._step = tf.function(self._step_impl)` is *tracing*. Every subsequent `step()` call has an identical input signature (`int32` scalar, `float64[30]`, `float64[30]`, `float64` scalar) and reuses the cached graph, so the python body never runs again.

Concretely: after `m.solve(1.0)` — which runs many Lanczos steps — `count[0] == 1`, and after `m.solve(0.25)`, `n_second == 0`. **Deleting the `if not self._started` guard in `GLTRSubproblem.solve`** — i.e. restarting Lanczos from scratch on every re-solve, destroying the feature under test — **leaves this test green.**

Assert on `len(m._deltas)` or `m.niter` instead; those are real python state that tracks actual Lanczos extension.

tests/test_native_minimizer.py
```diff
@@ -0,0 +1,583 @@
...
+        p_exact = _exact_subproblem_solution(g, H, tr_radius)
+        best = model(p_exact)
+        assert best < 0
+        assert np.linalg.norm(p) <= tr_radius * (1 + 1e-8)
+        assert model(p) <= 0.999 * best
+        # the cached model value prices the returned step
+        assert abs(m.model_value(p) - model(p)) < 1e-8 * (1 + abs(model(p)))
+
+
+def test_gltr_hard_case():
```
lucalavezzo, line 482: **This doesn't exercise the hard case** — the hard-case branch has zero coverage.

Ran `solve_tridiag_trust_region` with the test's own inputs (`diag=[-2,1,3]`, `off=[0,0.5]`, `gamma0=1`, `Delta=10`): eigenvalues `[-2, 0.882, 3.118]`, `gproj = [1, 0, 0]` — the gradient is *exactly* the bottom eigenvector, the opposite of the docstring's "g orthogonal to the bottom eigenvector". So `hnorm(lam_eps) = 5.0e11 >> Delta = 10`, control goes down the ordinary bisection path, and it returns `lam = 2.1 > -wmin`. The `degenerate` mask, the pseudo-inverse solve and `h[idx] += tau` are never executed by any test in the file.

Good news: I hand-triggered it with `diag=[3,1,-2]`, `off=[0.5,0]` and the machinery **is correct** — `lam = 2.0 = -wmin`, KKT residual 2.8e-17. So this is a coverage gap, not a live bug; those inputs would make the test do what it says.

Also, the half-finished comment `gamma0 = 1.0  # g = e1... wait e1 couples to mode 1` and the following "build instead directly" note were left in the committed test.

tests/test_native_minimizer.py
```diff
@@ -0,0 +1,583 @@
...
+@pytest.mark.parametrize(
+    "method", ["tf-trust-exact", "tf-trust-ncg", "tf-trust-krylov"]
+)
+@pytest.mark.parametrize("precondition", [False, True])
+def test_fit_matches_scipy(method, precondition):
+    with tempfile.TemporaryDirectory() as tmpdir:
+        fname = make_test_tensor(tmpdir)
+
+        res_native = run_fit_native(fname, method, precondition)
+        res_ref = run_fit(fname)  # trust-krylov: same likelihood, same minimum
```
lucalavezzo: **Cleanup:** this re-runs the identical scipy reference fit for all six parametrizations.

The test is parametrized 3 methods × 2 `precondition` settings, and each case calls `make_test_tensor(tmpdir)` (an HDF5 write) plus this `run_fit(fname)` — a full scipy `trust-krylov` fit whose result depends on neither parameter — before the one native fit. That's 6 tensor writes and 6 identical reference fits where 1 of each suffices, roughly doubling this file's CI runtime.

A module- or session-scoped fixture returning `(fname, res_ref)` fixes it.

rabbit/minimizer/exact.py
davidwalter2: Confirmed, and it was worse than "correct-but-slow" — thank you for catching it. Fixed in 69d9bbf.

I reproduced your finite-difference table to the digit, and the derivation settles it independently: `d||p||/dlambda = -p^T (H + lambda I)^-1 p / ||p|| = -||L^-1 p||^2 / ||p||`.

What I hadn't expected is the effect on step *quality*, not just on the number of factorizations. Model reduction relative to scipy `trust-exact` on the same subproblem, before -> after the one-line change:

| n | cond | radius | committed | fixed |
|---|---|---|---|---|
| 40 | 1e6 | 1e2 | 0.541 | **1.000000** |
| 40 | 1e9 | 1e2 | 0.644 | **1.000000** |
| 80 | 1e9 | 1e2 | 0.573 | **1.000000** |
| 120 | 1e12 | 1e2 | 0.379 | **1.000000** |

All twelve configurations I tried land on exactly 1.000000 after the fix. At cond 1e12 the committed code was getting 38% of scipy's model reduction.

I took the `_tri_solve` helper you suggested rather than the inline form — it does read better, and `_tri_solve_t` had no other caller so it's gone.

On why the suite missed it, see the reply on the test file: `_random_model` draws eigenvalues from `U(0.1, 10)`, so cond ~ 100, exactly where the safeguard absorbs the overshoot.

rabbit/minimizer/exact.py
davidwalter2: Confirmed and fixed. `hits_boundary` is now measured on the step actually returned:

```python
hits_boundary = bool(np.linalg.norm(p) >= tr_radius * (1.0 - 1e-12))
```

Your reading of the consequence is the part that made me prioritise it: the radius doubling in `base.py:131` fires precisely after a lambda-search failure, so the false claim compounds — harder subproblem, more likely to fail again.

rabbit/minimizer/exact.py
davidwalter2: Confirmed and guarded. Your path analysis holds exactly: no general branch (so `p` stays `None`), no Cauchy step, empty list, `ValueError` -> `warnflag=3` at the starting point.

I couldn't construct a real fit that reaches it — it needs an exact saddle — so it's latent rather than observed, but `min([])` aborting a fit is not something to leave in. Rather than `p = zeros`, it now steps along the estimated bottom eigenvector when a factor is available (sign chosen by model value), which is a descent direction wherever `H` is indefinite, so the outer loop can actually leave the saddle instead of stalling on a zero step. Falls back to zeros only when there is no usable factor.

rabbit/minimizer/exact.py
davidwalter2: Confirmed and removed. `self.hess_diag = tf.linalg.diag_part(self.hess)` cached once, and

```python
H = tf.linalg.set_diag(self.hess, self.hess_diag + lambda_current)
```

per lambda trial. Your 1.4 GB at n=13000 checks out (13000^2 x 8 B = 1.35 GB), and `base.py:136` does build a fresh subproblem on every accepted step, so the two would indeed have overlapped.

rabbit/minimizer/exact.py
davidwalter2: Confirmed by grep — defined once, referenced only from the device version's docstring, no test coverage. Removed, along with the `import scipy.linalg` it was the only consumer of. The device function's docstring keeps the comparison to scipy's Cline et al. recurrence, now as a reference to scipy rather than to a local copy.

rabbit/minimizer/gltr.py
davidwalter2: Confirmed and moved to `__init__` as `self._can_extend`, next to `_started`. Agreed on the reasoning: breakdown is a property of the linearization point, not of one `solve()` call, so a re-solve extending from the post-breakdown `q` is extending from normalized roundoff.

I couldn't construct a case where the residual test actually fails to save it, so this is latent as far as I can tell — but the asymmetry with the rest of the Krylov state was clearly unintended.

rabbit/minimizer/gltr.py
davidwalter2: Confirmed and fixed at both sites — `gltr.py` `ensure()` and `krylov.py:194` — using the `is None` form you suggest.

rabbit/minimizer/krylov.py
davidwalter2: Confirmed both halves. `tolerance = min(0.5, sqrt(jac_mag)) * jac_mag <= 0.5 * jac_mag`, so the guard is false for every `jac_mag >= 0`; I checked at 0, 1e-30, 1e-12, 1 and 4 and it never fires.

And the consequence is real — I ran it:

```
CGSteihaug at g=0: |p|=0.000e+00  hits_boundary=True
GLTR       at g=0: |p|=0.000e+00  hits_boundary=False
```

so `GLTRSubproblem` was already right and this one was claiming the boundary for a zero step. Taking your `if self.jac_mag == 0.0`, with the comment rewritten to say why the inequality form doesn't work rather than asserting that it does.

tests/test_native_minimizer.py
davidwalter2: Confirmed by deleting the feature: with the `if not self._started` guard removed, the test stayed green exactly as you predicted.

Your suggested assertion needed one more bound than I first wrote. Asserting only `len(m._deltas) - k_first <= 2` still passes when the basis is rebuilt, because the difference goes *negative* — a shrunken radius terminates earlier, so the rebuilt basis is shorter than the original. It's bounded both ways now:

```python
assert k_first <= len(m._deltas) <= k_first + 2
```

which fails with `assert 4 <= 2` when the guard is deleted. Kept a weak `count[0] > 0` so the traced graph is still known to call `hessp` at all, with a comment explaining why the counter can't carry more than that.

tests/test_native_minimizer.py
davidwalter2: Confirmed exactly — `gproj = [1, 0, 0]`, `||h(lam_eps)|| = 5e11 >> Delta = 10`, ordinary bisection, `lam = 2.1 > -wmin = 2`. Your inputs do enter the branch: `lam = 2.0 = -wmin`, KKT residual 1.1e-16 here.

Taken as suggested, plus one assertion so the test can't silently drift back onto the bisection path:

```python
assert abs(lam + wmin) <= 1e-9 * max(1.0, abs(wmin))
```

In the hard case `lam` sits exactly at `-lambda_min` (the secular equation has no root above it), so that pins the degenerate/pseudo-inverse branch rather than merely permitting it. And the half-finished comment is gone — the docstring now records why the ordering matters, since it's the thing that decides which path runs.

davidwalter2:
Thanks — this was a genuinely useful review, and the `L⁻¹` catch in particular. I verified every finding before touching anything; all fifteen are real. Ten are fixed in 69d9bbf, four I'd like to discuss before doing, and one I think is overstated.

## Fixed

The `L⁻ᵀ` → `L⁻¹` fix turned out to matter more than "correct-but-slow". Model reduction relative to scipy `trust-exact`, before → after:

| n | cond | radius | committed | fixed |
|---|---|---|---|---|
| 40 | 1e6 | 1e2 | 0.541 | **1.000000** |
| 80 | 1e9 | 1e2 | 0.573 | **1.000000** |
| 120 | 1e12 | 1e2 | 0.379 | **1.000000** |

Twelve of twelve configurations land on exactly 1.000000 after the change.

**The suite couldn't see it**, which is the part I want to flag back. `_random_model` draws eigenvalues from `U(0.1, 10)` — cond ≈ 100, exactly where the safeguarded bracket absorbs the overshoot. All 34 tests passed with the bug and all 34 passed without it. `_random_model` now takes a `cond` argument and `test_subproblem_matches_scipy` is parametrized over `cond ∈ {None, 1e6}`; reverting the fix fails 8 of the 18 subproblem cases.

One thing fell out of that: at cond 1e6, indefinite, radius 100, **scipy** drops to 0.90 of the exact optimum while the native solver reaches 0.987. So `assert model(p_sp) <= 0.95 * best` — the line asserting scipy meets the same bar — fails, on scipy. I turned it into `model(p_tf) <= 0.98 * model(p_sp)`: the native solver keeps the absolute bar, and scipy is held to being a reference rather than a guarantee. Worth knowing that `trust-exact` is the weaker of the two in that corner now.

The rest, all with replies inline: `hits_boundary` on the returned step; the empty-`candidates` guard (stepping along the estimated bottom eigenvector rather than returning zeros, so the outer loop can actually leave the saddle); `set_diag` instead of the `n×n` eye; the dead `estimate_smallest_singular_value` and its `scipy.linalg` import; `_can_extend` as instance state; both falsy-zero `or` defaults; the `jac_mag == 0.0` guard and its comment.

Both test fixes are mutation-checked rather than just re-run. The warm-restart one needed a bound you didn't mention: `len(m._deltas) - k_first <= 2` still passes when the basis is rebuilt, because a shrunken radius terminates earlier and the difference goes *negative*. Bounded both ways it fails with `assert 4 <= 2` when the reuse guard is deleted.

121 pass on the full suite (40 in this file, up from 34), black/isort/pylint clean via the pre-commit hooks.

## Overstated, I think — the redundant reference fits

Real, but not "roughly doubling this file's CI runtime". Measured: the scipy reference fit is 3.19 s and `make_test_tensor` is 0.02 s, so the redundancy across six parametrizations is ~16 s of a ~101 s file, about 16%. Still worth a session-scoped fixture, just not urgent — and I'd rather do it together with the other test-time work than alone.

While measuring it I also found the 101 s figure itself is unstable: back-to-back runs of the unmodified branch and the fixed one gave 38.25 s and 34.53 s. The file is load-sensitive, so I'd treat any CI-runtime number here as ±3× until it's pinned down.

## Deferred, and I'd like your view

Four I've deliberately not touched, because each is a design decision rather than a fix:

1. **The `--precondition` Hessian round-trip** (`fitter.py:2413`). You're right that it's the transfer this PR exists to remove, and right that `tf_transforms()` is the fix. But moving `H_y = TᵀHT` on-device is a real change to the preconditioner's contract, and I'd want it benchmarked on the in-situ fit rather than reasoned about — the transform is per accepted step, not per HVP, so the win is smaller than the per-HVP argument suggests.

2. **The per-block scatter** (`preconditioner.py:352`). Confirmed and measured: at fixed n ≈ 240, one block is 0.216 ms per `apply_T` and 239 blocks is 1.755 ms — an 8× penalty from block count alone, ×2 per HVP. The batched gather/matvec/scatter you describe is clearly the right shape. I'd rather land it as its own change with a before/after on a real fit, since it touches the preconditioner rather than the minimizers.

3. **`--minimizerFtol`** (`fitter.py:2517`). Confirmed silently dropped — set at 2478, logged at 2479, and `grep ftol rabbit/minimizer/` finds nothing. Two different fixes and I don't have a preference yet: wire it into `base.py`'s termination test (a real feature, and it would then differ subtly from scipy's `ftol` semantics), or warn on unconsumed `sci_opts` keys (honest, cheap, and arguably the right thing for *any* unrecognised option on the native path). Which would you rather have?

4. **The `_NATIVE_MINIMIZERS` table** (`fitter.py:2511`). Agreed the argument order shouldn't live in the fitter. Happy to do it; it's pure churn in a PR that's already large, so I'd default to a follow-up unless you'd rather see it here.

Happy to fold any of these into this PR if you'd rather not split them.

rabbit/fitter.py
davidwalter2: Benchmarked it on a V100 (fp64), so the PCIe traffic is really measured rather than reasoned about. Three things came out that change what I'd do.

**Scope.** `native_closure` is used by `tf-trust-exact` only. `tf-trust-ncg` and `tf-trust-krylov` take `native_grad_closure` (an n-vector) plus `native_hessp` -> `pc.tf_transforms()`, which is already on device. So the round-trip costs nothing on the two Krylov methods, and the PR's headline benchmark (`tf-trust-krylov`, 1043 s vs 1435 s) is untouched by it.

**The transfer is not the dominant term.** Per accepted step, tensors pinned to `/GPU:0`:

| n | blocks | D2H | numpy transform | H2D | total |
|---|---|---|---|---|---|
| 3603 | 1 | 19.4 ms | 547.9 ms | 75.5 ms | **642.8 ms** |
| 3603 | 239 | 19.4 ms | 229.5 ms | 75.5 ms | **324.4 ms** |
| 6000 | 239 | 54.4 ms | 788.3 ms | 205.8 ms | **1048.5 ms** |

D2H + H2D is 95 ms of the 324 ms at 239 blocks -- 29%. The other 71% is `hess_to_internal` itself: at that block count it is almost entirely fancy-indexing overhead in numpy rather than flops. (Worth noting for anyone reproducing this: `EagerTensor` caches its host buffer, so timing `__array__()` on the same tensor twice reports 0.0 ms for the D2H. The numbers above use a fresh tensor per rep.)

**The obvious fix regresses the configuration we care about.** Porting `hess_to_internal` to device with the per-block gather/scatter structure:

| n | blocks | round-trip | per-block device | batched device |
|---|---|---|---|---|
| 3603 | 1 | 642.8 ms | 220.6 ms (2.9x) | **47.9 ms (13.4x)** |
| 3603 | 239 | 324.4 ms | 557.5 ms (**0.58x**) | **16.0 ms (20.3x)** |
| 6000 | 239 | 1048.5 ms | 926.7 ms (1.13x) | **30.7 ms (34.2x)** |

At 239 blocks the per-block version is **1.7x slower than the round-trip it replaces** -- for exactly the reason you give on `preconditioner.py:352`. So these two comments are one problem: a single gather of all block indices, a batched `[B,m,m] x [B,m,n]` matmul and a single scatter gets 20-34x. Both device variants verified equal to the numpy result to 1e-6 relative.

**The denominator, though.** On a real rabbit model (100k bins, 257 parameters, same GPU):

```
loss_val_grad_hess()                  191.4 ms   <- the accepted step's real work
--precondition Hessian round-trip       1.1 ms   (0.6% of the two)
one n x n Cholesky (lambda trial)       1.1 ms
```

0.6% of an accepted step there; extrapolating to n=3603 against the ~11 s outer iterations in the PR description puts 324 ms at roughly 3%. So I think "precisely the transfer the native minimizer exists to eliminate" overstates the wall-clock stake -- but your underlying point stands, it plainly belongs on device, and 20x for a batched rewrite is worth having.

**What I'd propose:** do it as one follow-up covering this and `preconditioner.py:352` together, in the batched form rather than the per-block one. Not in this PR, because it is preconditioner work that wants its own before/after on the in-situ fit, and the method it speeds up is not the one this PR is selling. Say if you'd rather see it here and I will fold it in.

Scripts and the worktree are at `/scratch/submit/cms/dwalter/rabbit_benchmark/260910_precondTransfer/` on submit if you want to re-run any of it.

rabbit/preconditioner.py
davidwalter2: Benchmarked this together with the `fitter.py:2413` round-trip, since the fix is the same one -- full numbers in that thread.

Short version: your per-block cost estimate is right, and it is large enough that a naive device port of `hess_to_internal` using this structure comes out **1.7x slower** than the host round-trip it was meant to replace (557 ms vs 324 ms at n=3603, 239 blocks). The concatenate-once / batched-matvec / scatter-once shape you describe is what makes it work -- 16 ms, i.e. 20x faster than the round-trip and 35x faster than the per-block form.

I also measured the vector path directly, which is what this comment is actually about. At fixed n ~ 240, one `apply_T` call costs 0.216 ms with 1 block and **1.755 ms with 239 blocks** -- an 8x penalty from block count alone, twice per HVP. So this matters for `tf-trust-ncg`/`tf-trust-krylov` inside the compiled loop quite apart from the dense-Hessian path.

Proposing to do both as one follow-up; details and the trade-off in the other thread.

rabbit/fitter.py
davidwalter2: Fixed in fd1ef57, taking the second of your two options.

I went with warning on unconsumed options rather than wiring `ftol` into `base.py`'s termination test. Two reasons: implementing it would be a new feature whose semantics would differ subtly from scipy's `ftol` (scipy's is a relative-reduction test inside its own loop), and a declaration of what each method consumes generalises to any future option instead of special-casing this one.

```python
NATIVE_MINIMIZER_OPTIONS = {
    "tf-trust-exact": {"gtol", "maxiter"},
    "tf-trust-ncg": {"gtol", "maxiter"},
    "tf-trust-krylov": {"gtol", "maxiter"},
}
```

and the fitter diffs that against `sci_opts` right after the `[minimize]` line:

```
INFO: [minimize] method=tf-trust-exact options={'ftol': 1e-08}
WARNING: tf-trust-exact does not implement ftol; ignoring --minimizerFtol.
         Use a scipy --minimizerMethod if you need it.
```

Verified it fires for all three `tf-` methods with `--minimizerFtol`, stays quiet without it, and stays quiet on the scipy methods so there is no duplicate alongside scipy's own `OptimizeWarning`. `test_unconsumed_minimizer_option_warns` pins both directions.

One correction while I was in there: `--minimizerFtol`'s help already said "(L-BFGS-B only) ... Ignored by methods that don't honor 'ftol'". So the documentation was fine; the gap was purely the missing runtime signal, which is what your comment was really about.

tests/test_native_minimizer.py
davidwalter2: Fixed in fd1ef57 -- both the tensor and the reference fit move to a module-scoped fixture, as you suggest.

On the size of it, though: I measured rather than estimated, three runs each of the six parametrizations.

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| as committed | 41.32 s | 40.42 s | 41.09 s |
| with the fixture | 30.35 s | 30.21 s | 31.26 s |

So ~10.7 s, or 26% of that subset -- worth having, but not "roughly doubling this file's CI runtime". The reference fit itself is 3.19 s and `make_test_tensor` is 0.02 s, so five redundant reference fits was the whole of it.

Worth knowing separately: this file's wall-clock is load-sensitive enough that single-run numbers do not mean much. I measured the *unmodified* branch at both 38.25 s and 100.6 s on the same machine within the same session, which is roughly the size of the effect you were attributing to the redundancy.

rabbit/fitter.py
davidwalter2: Agreed on all of it — the argument order for each minimizer should live next to the implementations, not in a three-branch chain in the fitter, and `native_closure` duplicating the `cond_number`/`edmval` block from `scipy_hess` is straightforwardly a copy-paste.

Not doing it here, though: it is pure churn in a PR that is already 2k lines, and it touches the same `fit()` block as the two performance items in the other threads (`fitter.py:2413` and `preconditioner.py:352`), which I have proposed doing as one follow-up. Collapsing the chain at the same time as moving the Hessian transform on device means one careful pass over that block instead of two.

Happy to fold it in here instead if you would rather not carry it — say the word and I will.

lucalavezzo:
OK sorry for the slop-review, looks like it caught a small bug on the L^-1, other things seem minor to me, or can be done in a separate PR, good for me to merge.

PR #154, opened 2026-08-28
Bins-sharded multi-device fits (--nDevices)

davidwalter2:
**Depends on #153** (native TensorFlow trust-region minimizers). This branch is
based on `260826_tfMinimizer`, so until #153 merges the diff below also shows
its commits; it reduces to the seven listed at the end once #153 lands. Review
from "Add multi-device fits" onward.

## What this adds

`--nDevices N` splits the likelihood's bins across N devices. Each shard is a
duck-typed evaluator exposing the attributes the existing `Fitter` methods
read, and the Fitter's own unbound methods are bound onto it, so both
interpolation forms, both systematic types and the bin-by-bin-stat machinery
are shared with the single-device path by construction rather than duplicated.

Shard graphs are jit-compiled individually (XLA clusters are single-device) and
the thin combiner stays a plain graph. `--devices` picks the GPUs explicitly;
otherwise the least-occupied ones are chosen, so a fit started on a shared node
lands next to nobody.

On a 92144-bin, 6538-parameter W+Z fit across 4 V100s this preconditions all 71
parameter blocks in under 7 minutes and then iterates in seconds.

## The dense Hessian

Both the preconditioner's reference matrix and the postfit covariance need a
dense `[npar, npar]` Hessian. It is assembled a batch of columns at a time from
Hessian-vector products, which are the same HVPs the minimiser already
evaluates each iteration, so the work stays distributed across the shards.

`tf.vectorized_map` is a pfor of width k, so XLA compiles it without the
unrolling that a `while_loop` of the same shape produces, and peak memory is
one `[k, npar]` batch rather than the `[npar, nbins, 9]` that vectorising
`tape.jacobian` over every parameter at once would need (43 GB for this model,
~70 GB per device once sharded). `--hvpBatch` (default 256) sets k, and the
batch halves itself on `ResourceExhaustedError`, so it is an upper bound rather
than a number that has to be tuned.

## Regularizers

`Regularizer.needs_observables` declares whether a penalty reads the predicted
yields. Penalties that depend only on the parameters are supported here: every
device holds the whole parameter vector, so the penalty is evaluated once in
the global term `gnll_local`, which `gvg` and `gvgp` differentiate — gradient
and HVP come for free. Penalties that read the yields are rejected in
`arm_regularizers`, which runs when regularizers are present and before any
loss is evaluated; the graphs are rebuilt there, since `__init__` traces them
before `--regularizer` has been parsed.

The flag also pays off single-device: `_compute_nll_components` only builds
`full=True` yields when some regularizer actually needs them, and each
regularizer is handed `None` rather than a truncated vector when it has
declared it does not.

## Not supported in sharded mode

Checked at construction: sparse tensors, `--covarianceFit`.

Rejected at the point of use, before the fit starts: regularizers with
`needs_observables=True`, `--fullNll`, global impacts (`--doImpacts`), and toys
(`-t > 0`). These are global over all bins on one device, which is what the
sharding exists to avoid; each raises `NotImplementedError` naming the flag and
the single-device remedy. The fwdrev HVP falls back to revrev.

## Testing

`tests/test_sharded_fit.py` checks a 2-device fit against the single-device one
on parameters, uncertainties, NLL and EDM; that a parameter-only penalty shifts
the sharded loss by exactly the amount it shifts the single-device loss, and
reaches the gradient; and that each unsupported postfit step raises rather than
running. 114 tests pass on this branch.

## Commits

- Add multi-device fits: bins-sharded likelihood via `--nDevices`
- Only occupy the GPUs a fit actually uses, preferring unoccupied ones
- Move multi-device support into a `MultiDeviceFitter` subclass
- Forward all Fitter kwargs through `make_fitter`
- Let a regularizer declare whether its penalty reads the yields
- Assemble the dense Hessian from batched Hessian-vector products
- Stop multi-device fits silently computing the wrong thing

https://claude.ai/code/session_01HrsrmERvJF2Vafn7wE1B3v

### DIFF
--- a/bin/rabbit_fit.py
+++ b/bin/rabbit_fit.py
@@ -840,6 +840,21 @@ def main():
     if args.eager:
         tf.config.run_functions_eagerly(True)
 
+    # Restrict TF to the requested GPUs *before* the runtime context
+    # initializes: by default TF creates a CUDA context on -- and reserves
+    # the memory of -- every visible GPU, so a single-device fit on a
+    # 4-GPU machine would block all four while using one. On shared
+    # interactive nodes the selection also avoids GPUs another process is
+    # occupying (least-memory-used first); --devices picks explicitly.
+    # Under slurm with --gres this is all moot (CUDA_VISIBLE_DEVICES
+    # already hides other jobs' GPUs). Must run before any op touches the
+    # GPUs; list_physical_devices itself is safe.
+    from rabbit.sharding import pick_physical_gpus
+
+    chosen = pick_physical_gpus(args.nDevices, explicit=args.devices)
+    if chosen is not None:
+        tf.config.set_visible_devices(chosen, "GPU")
+
     # --noHessian skips computing the postfit Hessian, so the dense
     # parameter covariance matrix is never available. Any feature that
     # needs the covariance is incompatible.
--- a/bin/rabbit_fit.py
+++ b/bin/rabbit_fit.py
@@ -873,7 +873,9 @@ def main():
     )
     blinded_fits = [f == 0 or (f > 0 and args.toysDataMode == "observed") for f in fits]
 
-    indata = inputdata.FitInputData(args.filename, args.pseudoData)
+    indata = inputdata.FitInputData(
+        args.filename, args.pseudoData, host_memory=args.nDevices > 1
+    )
 
     model_specs = args.paramModel or [["Mu"]]
     param_model = ph.load_models(model_specs, indata, **vars(args))
--- a/bin/rabbit_fit.py
+++ b/bin/rabbit_fit.py
@@ -895,7 +895,7 @@ def main():
     model_specs = args.paramModel or [["Mu"]]
     param_model = ph.load_models(model_specs, indata, **vars(args))
 
-    ifitter = fitter.Fitter(
+    ifitter = fitter.make_fitter(
         indata,
         param_model,
         args,
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -12,6 +12,7 @@ from wums import logging
 
 from rabbit import external_likelihood, io_tools
 from rabbit import preconditioner as precond
+from rabbit import sharding
 from rabbit import tfhelpers as tfh
 from rabbit.bbstat.bbstat import BinByBinStat
 from rabbit.callbacks import RESTART_MIN_IMPROVEMENT, FitterCallback, merge_callbacks
@@ -194,6 +195,12 @@ class Fitter:
         # All BBB state is owned by the BinByBinStat helper. Constructed
         # here, before init_fit_parms, because init_fit_parms's is_linear
         # computation reads self.bbstat.enabled.
+        # Retained for constructing per-shard BinByBinStat instances in
+        # multi-device mode (see _build_shards).
+        self._options = options
+        self.n_devices = int(getattr(options, "nDevices", 1) or 1)
+        self.shards = []
+
         self.bbstat = BinByBinStat(
             indata,
             options,
@@ -449,6 +456,13 @@ class Fitter:
                     tf.function(val.python_function.__get__(self, type(self))),
                 )
 
+        # Multi-device mode: (re)build the per-shard evaluators against the
+        # current parameter layout before the tf.functions that close over
+        # them. Everything the shards borrow (logk, bbstat options, blinding
+        # offsets, frozen mask) exists by this point.
+        if self.n_devices > 1:
+            self._build_shards()
+
         # (re)build instance-level tf.function wrappers for loss/grad/HVP, which
         # are constructed dynamically so that jit_compile and the HVP autodiff
         # mode can be controlled via fit options.
@@ -473,6 +487,12 @@ class Fitter:
             "loss_val_grad_hessp",
             "loss_val_grad_hessp_fwdrev",
             "loss_val_grad_hessp_revrev",
+            # multi-device machinery: shard evaluators hold bound methods and
+            # traced functions; rebuilt from scratch on the copy
+            "shards",
+            "_shard_graph_fns",
+            "_global_view",
+            "_global_graph_fns",
         }
         skip = jit_overrides | dynamic_tf_funcs
         state = {k: v for k, v in self.__dict__.items() if k not in skip}
@@ -481,6 +501,9 @@ class Fitter:
         memo[id(self)] = obj
         for k, v in state.items():
             setattr(obj, k, copy.deepcopy(v, memo))
+        obj.shards = []
+        if obj.n_devices > 1:
+            obj._build_shards()
         obj._make_tf_functions()
         return obj
 
@@ -767,6 +790,13 @@ class Fitter:
         # the offset is chosen to give the saturated likelihood
         nobssafe = tf.where(values == 0.0, tf.constant(1.0, dtype=values.dtype), values)
         self.lognobs.assign(tf.math.log(nobssafe))
+
+        for shard in self.shards:
+            a, b = shard.indata.start, shard.indata.stop
+            shard.nobs.assign(values[a:b])
+            shard.lognobs.assign(tf.math.log(nobssafe[a:b]))
+            if shard.varnobs is not None:
+                shard.varnobs.assign(self.varnobs[a:b])
 
     def x0defaultassign(self):
         # reset all constraint centers
@@ -2133,7 +2163,365 @@ class Fitter:
     def _compute_loss(self, profile=True):
         return self._compute_nll(profile=profile)
 
+    # Fitter methods that also run per shard: the shard evaluators are
+    # duck-typed stand-ins (see rabbit.sharding), and binding the class's own
+    # unbound methods onto them shares the yields/NLL mathematics -- both
+    # interpolation forms, both systematic types, the BBB machinery -- with
+    # the single-device path by construction.
+    _SHARD_BORROWED_METHODS = (
+        "get_poi",
+        "get_model_nui",
+        "get_theta",
+        "get_x",
+        "_compute_yields_noBBB",
+        "_compute_yields_with_beta",
+        "_compute_ln",
+        "_compute_lbeta",
+        "_compute_lc",
+    )
+
+    def _build_shards(self):
+        """Construct the per-device shard evaluators (multi-device mode).
+
+        Shards partition the observed bins contiguously; each holds its
+        slice of logk / norm / nobs / sumw(2) and its own BinByBinStat
+        (beta and kstat are per-bin, so BBB shards along with everything
+        else). Parameter-level state (x, frozen mask, blinding offsets,
+        param model) is shared -- those tensors are [nparams]-sized and
+        cross devices for kilobytes per evaluation.
+        """
+        if self.indata.sparse:
+            raise NotImplementedError(
+                "Multi-device fits (--nDevices > 1) are not supported in "
+                "sparse mode."
+            )
+        if self.covarianceFit:
+            raise NotImplementedError(
+                "Multi-device fits are not supported with --covarianceFit "
+                "(the dense data covariance couples all bins)."
+            )
+
+        devices = sharding.select_devices(self.n_devices)
+        edges = sharding.shard_edges(self.indata.nbins, self.n_devices)
+        logger.info(
+            f"Sharding {self.indata.nbins} bins over {self.n_devices} "
+            f"device(s): {[f'{d}:[{a},{b})' for d, (a, b) in zip(devices, edges)]}"
+        )
+
+        self.shards = []
+        for device, (a, b) in zip(devices, edges):
+            view = sharding.ShardIndataView(self.indata, a, b, device)
+            shard = sharding.ShardEvaluator(device, view)
+            with tf.device(device):
+                shard.logk = tf.identity(self.logk[a:b])
+                shard.nobs = tf.Variable(
+                    tf.zeros([b - a], dtype=self.indata.dtype),
+                    trainable=False,
+                    name=f"nobs_shard{a}",
+                )
+                shard.lognobs = tf.Variable(
+                    tf.zeros([b - a], dtype=self.indata.dtype),
+                    trainable=False,
+                    name=f"lognobs_shard{a}",
+                )
+                shard.varnobs = (
+                    tf.Variable(
+                        tf.ones([b - a], dtype=self.indata.dtype),
+                        trainable=False,
+                        name=f"varnobs_shard{a}",
+                    )
+                    if self.chisqFit
+                    else None
+                )
+                shard.bbstat = BinByBinStat(
+                    view,
+                    self._options,
+                    chisqFit=self.chisqFit,
+                    covarianceFit=False,
+                    data_cov_inv=None,
+                    nobs_template=shard.nobs,
+                )
+            # seed from the current observation state: init_fit_parms (and
+            # thus shard construction) can re-run after set_nobs
+            shard.nobs.assign(self.nobs[a:b])
+            shard.lognobs.assign(self.lognobs[a:b])
+            if shard.varnobs is not None:
+                shard.varnobs.assign(self.varnobs[a:b])
+            shard.param_model = sharding.ShardParamModel(self.param_model, a, b)
+            # Parameter-level state. NB the fitter's tf.Variables
+            # (frozen_params_mask, blinding offsets) must NOT be captured
+            # here: XLA-compiled functions cannot read a Variable resident
+            # on a different device, so their *values* are threaded into the
+            # per-shard functions as arguments instead (see
+            # _make_sharded_tf_functions), the same way x is.
+            shard.do_blinding = self.do_blinding
+            shard.chisqFit = self.chisqFit
+            shard.covarianceFit = False
+            shard.data_cov_inv = None
+            for name in self._SHARD_BORROWED_METHODS:
+                setattr(shard, name, getattr(Fitter, name).__get__(shard))
+            self.shards.append(shard)
+
+        # the parameter-level NLL terms (constraints + external likelihood
+        # terms), evaluated once per call on the first device via the same
+        # borrowed-method pattern
+        gview = sharding.ShardEvaluator(devices[0], self.indata)
+        gview.param_model = self.param_model
+        gview.frozen_params_mask = self.frozen_params_mask
+        gview.do_blinding = self.do_blinding
+        if self.do_blinding:
+            gview._blinding_offsets_poi = self._blinding_offsets_poi
+            gview._blinding_offsets_theta = self._blinding_offsets_theta
+        gview.cw = self.cw
+        gview.x0 = self.x0
+        for name in ("get_poi", "get_model_nui", "get_theta", "get_x", "_compute_lc"):
+            setattr(gview, name, getattr(Fitter, name).__get__(gview))
+        self._global_view = gview
+
+    def _make_sharded_tf_functions(self):
+        """Multi-device counterparts of the loss/grad/HVP/Hessian wrappers.
+
+        The per-shard functions are jit-compiled on their own device and
+        differentiate with respect to their local x copy (one tape per
+        shard); the combiners below are plain graphs that broadcast x,
+        collect the [nparams]-sized partials and sum them. See
+        rabbit/sharding.py for why both choices are load-bearing.
+        """
+        jit = self.jit_compile
+
+        def make_shard_fns(shard):
+            # aux = (frozen_params_mask value[, blinding offsets]) -- Variable
+            # values threaded in as tensors because XLA cannot read a
+            # Variable on another device
+            def _pin(x, aux):
+                shard.x = x
+                shard.frozen_params_mask = aux[0]
+                if shard.do_blinding:
+                    shard._blinding_offsets_poi = aux[1]
+                    shard._blinding_offsets_theta = aux[2]
+
+            def nll_local(x, aux):
+                _pin(x, aux)
+                nexp, _, beta = shard._compute_yields_with_beta(
+                    profile=True, compute_norm=False, full=False
+                )
+                ln = shard._compute_ln(nexp[: shard.indata.nbins], full_nll=False)
+                lbeta = shard._compute_lbeta(beta, full_nll=False)
+                return ln + lbeta if lbeta is not None else ln
+
+            def vg(x, aux):
+                with tf.GradientTape() as t:
+                    t.watch(x)
+                    v = nll_local(x, aux)
+                g = t.gradient(v, x, unconnected_gradients=tf.UnconnectedGradients.ZERO)
+                return v, g
+
+            def vgp(x, aux, p):
+                p = tf.stop_gradient(p)
+                with tf.GradientTape() as t2:
+                    t2.watch(x)
+                    with tf.GradientTape() as t1:
+                        t1.watch(x)
+                        v = nll_local(x, aux)
+                    g = t1.gradient(
+                        v, x, unconnected_gradients=tf.UnconnectedGradients.ZERO
+                    )
+                hp = t2.gradient(
+                    g,
+                    x,
+                    output_gradients=p,
+                    unconnected_gradients=tf.UnconnectedGradients.ZERO,
+                )
+                return v, g, hp
+
+            def vgh(x, aux):
+                with tf.GradientTape() as t2:
+                    t2.watch(x)
+                    with tf.GradientTape() as t1:
+                        t1.watch(x)
+                        v = nll_local(x, aux)
+                    g = t1.gradient(
+                        v, x, unconnected_gradients=tf.UnconnectedGradients.ZERO
+                    )
+                h = t2.jacobian(g, x)
+                return v, g, h
+
+            def beta_local(x, aux):
+                _pin(x, aux)
+                _, _, beta = shard._compute_yields_with_beta(
+                    profile=True, compute_norm=False, full=False
+                )
+                return beta
+
+            return (
+                tf.function(nll_local, jit_compile=jit),
+                tf.function(vg, jit_compile=jit),
+                tf.function(vgp, jit_compile=jit),
+                tf.function(vgh, jit_compile=jit),
+                tf.function(beta_local, jit_compile=jit),
+            )
+
+        self._shard_graph_fns = [make_shard_fns(shard) for shard in self.shards]
+
+        gview = self._global_view
+
+        def gnll_local(x):
+            gview.x = x
+            l = gview._compute_lc(full_nll=False)
+            lext = external_likelihood.compute_external_nll(
+                self.external_terms, x, self.indata.dtype, full_nll=False
+            )
+            return l + lext if lext is not None else l
+
+        def gvg(x):
+            with tf.GradientTape() as t:
+                t.watch(x)
+                v = gnll_local(x)
+            g = t.gradient(v, x, unconnected_gradients=tf.UnconnectedGradients.ZERO)
+            return v, g
+
+        def gvgp(x, p):
+            p = tf.stop_gradient(p)
+            with tf.GradientTape() as t2:
+                t2.watch(x)
+                with tf.GradientTape() as t1:
+                    t1.watch(x)
+                    v = gnll_local(x)
+                g = t1.gradient(
+                    v, x, unconnected_gradients=tf.UnconnectedGradients.ZERO
+                )
+            hp = t2.gradient(
+                g,
+                x,
+                output_gradients=p,
+                unconnected_gradients=tf.UnconnectedGradients.ZERO,
+            )
+            return v, g, hp
+
+        def gvgh(x):
+            with tf.GradientTape() as t2:
+                t2.watch(x)
+                with tf.GradientTape() as t1:
+                    t1.watch(x)
+                    v = gnll_local(x)
+                g = t1.gradient(
+                    v, x, unconnected_gradients=tf.UnconnectedGradients.ZERO
+                )
+            h = t2.jacobian(g, x)
+            return v, g, h
+
+        self._global_graph_fns = tuple(
+            tf.function(f, jit_compile=jit) for f in (gnll_local, gvg, gvgp, gvgh)
+        )
+
+        def _read_aux():
+            aux = [tf.identity(self.frozen_params_mask)]
+            if self.do_blinding:
+                aux.append(tf.identity(self._blinding_offsets_poi))
+                aux.append(tf.identity(self._blinding_offsets_theta))
+            return tuple(aux)
+
+        def _to_device(aux):
+            return tuple(tf.identity(a) for a in aux)
+
+        def _loss_val():
+            x0 = tf.identity(self.x)
+            aux = _read_aux()
+            parts = []
+            for shard, fns in zip(self.shards, self._shard_graph_fns):
+                with tf.device(shard.device):
+                    parts.append(fns[0](tf.identity(x0), _to_device(aux)))
+            with tf.device(self.shards[0].device):
+                parts.append(self._global_graph_fns[0](tf.identity(x0)))
+            return tf.add_n(parts)
+
+        def _loss_val_grad():
+            x0 = tf.identity(self.x)
+            aux = _read_aux()
+            vs, gs = [], []
+            for shard, fns in zip(self.shards, self._shard_graph_fns):
+                with tf.device(shard.device):
+                    v, g = fns[1](tf.identity(x0), _to_device(aux))
+                vs.append(v)
+                gs.append(g)
+            with tf.device(self.shards[0].device):
+                v, g = self._global_graph_fns[1](tf.identity(x0))
+            vs.append(v)
+            gs.append(g)
+            return tf.add_n(vs), tf.add_n(gs)
+
+        def _loss_val_grad_hessp(p):
+            x0 = tf.identity(self.x)
+            aux = _read_aux()
+            vs, gs, hs = [], [], []
+            for shard, fns in zip(self.shards, self._shard_graph_fns):
+                with tf.device(shard.device):
+                    v, g, hp = fns[2](tf.identity(x0), _to_device(aux), tf.identity(p))
+                vs.append(v)
+                gs.append(g)
+                hs.append(hp)
+            with tf.device(self.shards[0].device):
+                v, g, hp = self._global_graph_fns[2](tf.identity(x0), tf.identity(p))
+            vs.append(v)
+            gs.append(g)
+            hs.append(hp)
+            return tf.add_n(vs), tf.add_n(gs), tf.add_n(hs)
+
+        def _loss_val_grad_hess(profile=True):
+            if not profile:
+                raise NotImplementedError(
+                    "profile=False Hessians are not supported in multi-device "
+                    "mode (--nDevices > 1)."
+                )
+            x0 = tf.identity(self.x)
+            aux = _read_aux()
+            vs, gs, hs = [], [], []
+            for shard, fns in zip(self.shards, self._shard_graph_fns):
+                with tf.device(shard.device):
+                    v, g, h = fns[3](tf.identity(x0), _to_device(aux))
+                vs.append(v)
+                gs.append(g)
+                hs.append(h)
+            with tf.device(self.shards[0].device):
+                v, g, h = self._global_graph_fns[3](tf.identity(x0))
+            vs.append(v)
+            gs.append(g)
+            hs.append(h)
+            return tf.add_n(vs), tf.add_n(gs), tf.add_n(hs)
+
+        def _profile_beta_sharded():
+            x0 = tf.identity(self.x)
+            aux = _read_aux()
+            betas = []
+            for shard, fns in zip(self.shards, self._shard_graph_fns):
+                with tf.device(shard.device):
+                    beta = fns[4](tf.identity(x0), _to_device(aux))
+                shard.bbstat.beta.assign(beta)
+                betas.append(beta)
+            # mirror into the fitter-level beta variable, whose masked-bin
+            # tail (never touched by the sharded loss) keeps its prior value
+            nbins = self.indata.nbins
+            full = tf.concat(betas + [self.bbstat.beta[nbins:]], axis=0)
+            self.bbstat.beta.assign(full)
+
+        self.loss_val = tf.function(_loss_val)
+        self.loss_val_grad = tf.function(_loss_val_grad)
+        self.loss_val_grad_hessp_revrev = tf.function(_loss_val_grad_hessp)
+        if self.hvp_method == "fwdrev":
+            logger.warning(
+                "fwdrev HVP is not supported in multi-device mode; "
+                "falling back to revrev."
+            )
+        self.loss_val_grad_hessp_fwdrev = self.loss_val_grad_hessp_revrev
+        self.loss_val_grad_hessp = self.loss_val_grad_hessp_revrev
+        # instance override of the class-level @tf.function
+        self.loss_val_grad_hess = _loss_val_grad_hess
+        if self.bbstat.enabled:
+            self._profile_beta = tf.function(_profile_beta_sharded)
+
     def _make_tf_functions(self):
+        if self.shards:
+            return self._make_sharded_tf_functions()
         # Build tf.function wrappers at instance construction time so that
         # jit_compile and the HVP autodiff mode can be controlled via fit
         # options without redefining the class. self.jit_compile has
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -66,20 +66,22 @@ def match_regexp_params(regular_expressions, parameter_names):
     return matched
 
 
-def make_fitter(indata, param_model, options, do_blinding=False):
+def make_fitter(indata, param_model, options, **kwargs):
     """Construct the Fitter appropriate for the requested device count.
 
     The device layout is fixed at initialization, so the choice is a
     static one: --nDevices > 1 returns the bins-sharded
     :class:`rabbit.sharding.MultiDeviceFitter` subclass, anything else the
     plain single-device Fitter. Imported lazily to avoid a module cycle
-    (sharding subclasses Fitter).
+    (sharding subclasses Fitter). All keyword arguments are forwarded
+    verbatim, so this wrapper cannot drift from Fitter.__init__'s
+    signature.
     """
     if int(getattr(options, "nDevices", 1) or 1) > 1:
         from rabbit.sharding import MultiDeviceFitter
 
-        return MultiDeviceFitter(indata, param_model, options, do_blinding=do_blinding)
-    return Fitter(indata, param_model, options, do_blinding=do_blinding)
+        return MultiDeviceFitter(indata, param_model, options, **kwargs)
+    return Fitter(indata, param_model, options, **kwargs)
 
 
 class Fitter:
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -124,6 +124,7 @@ class Fitter:
         # outcome can be written to the output. None if the minimizer raised.
         self.minimizer_result = None
         self.hvp_method = getattr(options, "hvpMethod", "revrev")
+        self.hvp_batch = int(getattr(options, "hvpBatch", 256) or 256)
         # Optional parameter preconditioning (see rabbit/preconditioner.py).
         # getattr so callers that build options objects by hand keep working.
         self.precondition = getattr(options, "precondition", False)
@@ -817,6 +818,14 @@ class Fitter:
 
         self.arm_regularizers()
 
+    def _regularizers_need_observables(self):
+        """Whether any attached regularizer reads the predicted yields.
+
+        Penalties on the parameters alone (see Regularizer.needs_observables)
+        do not, and then the full yields never have to be built for them.
+        """
+        return any(getattr(reg, "needs_observables", True) for reg in self.regularizers)
+
     def arm_regularizers(self):
         """Tell every regularizer about the current parameter layout.
 
@@ -824,9 +833,17 @@ class Fitter:
         ``init_fit_parms``.
         """
         xinit = self.get_x()
-        nexp0 = self.expected_yield(full=True)
+        nexp0 = (
+            self.expected_yield(full=True)
+            if self._regularizers_need_observables()
+            else None
+        )
         for reg in self.regularizers:
-            reg.set_expectations(xinit, nexp0, parms=self.parms)
+            reg.set_expectations(
+                xinit,
+                nexp0 if getattr(reg, "needs_observables", True) else None,
+                parms=self.parms,
+            )
         self._regularizers_armed = True
 
     def bayesassign(self):
@@ -2092,7 +2109,9 @@ class Fitter:
         nexpfullcentral, _, beta = self._compute_yields_with_beta(
             profile=profile,
             compute_norm=False,
-            full=len(self.regularizers),
+            # only build the full (masked-channel-inclusive) yields if some
+            # regularizer actually reads them
+            full=self._regularizers_need_observables(),
         )
 
         nexp = nexpfullcentral[: self.indata.nbins]
@@ -2111,7 +2130,15 @@ class Fitter:
                 )
             x = self.get_x()
             penalties = [
-                reg.compute_nll_penalty(x, nexpfullcentral) * tf.exp(2 * self.tau)
+                reg.compute_nll_penalty(
+                    x,
+                    (
+                        nexpfullcentral
+                        if getattr(reg, "needs_observables", True)
+                        else None
+                    ),
+                )
+                * tf.exp(2 * self.tau)
                 for reg in self.regularizers
             ]
             lpenalty = tf.add_n(penalties)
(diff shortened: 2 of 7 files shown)


rabbit/sharding.py
```diff
@@ -0,0 +1,784 @@
...
+        gview.cw = self.cw
+        gview.x0 = self.x0
+        for name in ("get_poi", "get_model_nui", "get_theta", "get_x", "_compute_lc"):
+            setattr(gview, name, getattr(Fitter, name).__get__(gview))
+        self._global_view = gview
+
+    def _hessp_batch_dispatch(self, P):
+        return self._hessp_batch_sharded(P)
+
+    def _reference_matrix(self):
```
lucalavezzo, line 379: **Blocking — `--preconditionFrom gaussnewton` is silently ignored when sharded.**

This returns `self.hessian_from_hvps()` unconditionally and drops the Gauss-Newton branch entirely. Measured on the unit-test model: the sharded reference matrix vs the single-device GN matrix is **max|Δ| = 1124.8**; vs the exact Hessian it's **1.8e-12**. So a sharded fit preconditions from the exact Hessian no matter what the flag says, with no warning.

Given the argument in `Fitter._reference_matrix`'s own docstring — that swapping GN for exact took a 2112-parameter block from 145 min to 4.6 min, and that GN "cannot represent negative curvature at all" — this is the same class of silent-wrong-thing the branch set out to kill.

**The fix is to delete the override.** The base `_reference_matrix` calls `self.loss_val_grad_hess()`, which `MultiDeviceFitter` has *already* replaced with the HVP assembler, so both branches work unchanged. Verified: `Fitter._reference_matrix(sharded_fitter)` reproduces the single-device GN matrix to **2.7e-15**. Net −25 lines and the bug goes away.

**Also:** the docstring here describes the *previous* implementation. The summary line says "built unsharded on the host" and the body says "bypass the shard machinery … running those on the CPU needs ~43 GB of host RAM", while the code does the opposite (the inline comment below it is the accurate one). It reads as a correctness claim about placement, so it should go with the override.

rabbit/sharding.py
```diff
@@ -0,0 +1,784 @@
...
+            self.sumw2 = tf.identity(indata.sumw2[self.start : self.stop])
+        self.nbins = nb
+        self.nbinsfull = nb
+        self.nbinsmasked = 0
+        self.nproc = indata.nproc
+        self.nsyst = indata.nsyst
+        self.symmetric_tensor = indata.symmetric_tensor
+        self.systematic_type = indata.systematic_type
+        self.sparse = False
+        self.dtype = indata.dtype
```
lucalavezzo, line 167: **Blocking — `--binByBinStatMode full` crashes the sharded fit.**

`bbstat.py:846` evaluates `self.indata.betavar is not None` *before* the `and full` short-circuits, and `ShardIndataView` doesn't define `betavar`:

```
AttributeError: 'ShardIndataView' object has no attribute 'betavar'
  bbstat.py:846, in profile_and_apply: if self.indata.betavar is not None and full:
```

raised on the first `loss_val()` with `nDevices=2, binByBinStatMode="full"` (the default `lite` is unaffected, so this is opt-in — but it's an opaque crash rather than one of the deliberate `NotImplementedError`s).

Suggest `self.betavar = indata.betavar` here so the common `betavar is None` case just works — and since `betavar` is bin×bin (`einsum("ijk,jk->ik")` contracts over the bin axis) it is genuinely unshardable, so also raise a clean `NotImplementedError` from `_build_shards` when it is non-`None` in full mode.

rabbit/sharding.py
```diff
@@ -0,0 +1,784 @@
...
+
+    def full_nll(self):
+        raise NotImplementedError(
+            "--fullNll is not supported in multi-device mode (--nDevices > 1): "
+            "the sharded likelihood is built with full_nll=False, so the "
+            "normalisation terms are not available per shard. Rerun the "
+            "postfit single-device with --externalPostfit if it is needed."
+        )
+
+    def expected_yield(self, profile=False, full=False):
```
lucalavezzo, line 494: **Significant — the refuse-at-point-of-use list misses `--saveHists`.**

This docstring spells out exactly why `expected_yield` has to be pinned: on the default device it materialises the full `[nbins, nproc, nsyst]` logk on GPU:0, 30.6 GB for this model, "precisely what sharding exists to avoid".

But `expected_events` → `expected_with_variance` / `_compute_expected` / `chi2` / `expected_variations` are the *same* all-bins computation over the *same* tensor, are neither pinned nor refused, and `save_hists` calls them for every mapping immediately after the minimiser. That is the identical end-of-a-multi-hour-fit failure the `_unsharded` guards exist to prevent — the reasoning here just wasn't carried across to them.

Either pin them the same way, or refuse `--saveHists` / `--computeHistErrors` / `--computeVariations` through `_unsharded()`.

rabbit/inputdata.py
```diff
@@ -7,7 +7,17 @@
 
 
 class FitInputData:
-    def __init__(self, filename, pseudodata=None):
+    def __init__(self, filename, pseudodata=None, host_memory=False):
+        # In multi-device (sharded) fits the full tensors must not
+        # materialize on one GPU -- only per-device slices are copied over
+        # later -- so all loading is pinned to host memory.
+        if host_memory:
+            with tf.device("/CPU:0"):
```
lucalavezzo, line 15: **Significant — this pinning doesn't cover the `systematic_type == "normal"` logk copy.**

`fitter.py:1716` (`_build_logk`) does `self.indata.logk * rnorm_init[..., None]` eagerly with no device scope, so for `systematic_type == "normal"` with a non-trivial param model the full tensor is built on GPU:0 before any shard exists — defeating both this `host_memory=True` pinning and the careful host-side slicing in `_build_shards`.

Only bites that one systematic type, but it's a one-line `tf.device("/CPU:0")`.

lucalavezzo (commented):
### Summary of what this does

`--nDevices N` splits the likelihood's **bins** across N devices. The NLL is a sum over bins plus parameter-level terms and every large tensor (`logk`, `norm`, `nobs`, `sumw/sumw2`, `beta`) is bins-proportional, so slicing the bin axis splits both memory and compute; only the parameter vector goes out and one `[npar]` partial comes back per shard.

Three choices carry it: (1) each shard is a duck-typed `ShardEvaluator` with `Fitter`'s own *unbound* methods bound onto it, so both interpolation forms, both systematic types and the BBB machinery are shared with the single-device path by construction rather than copied — this is the right call and it's what makes 784 new lines reviewable; (2) one tape per shard differentiating w.r.t. a device-local `x`, with parameter-level Variables threaded in as values because XLA can't read a Variable resident elsewhere; (3) the dense Hessian assembled `--hvpBatch` columns at a time from the HVPs the minimiser already computes, so it stays distributed and peak memory is one `[k, npar]` batch instead of `[npar, nbins, 9]`.

I ran the suite in the container — **130 passed, 142 s** (CPU-only) — and checked the numerics by hand: HVP-assembled Hessian max asymmetry **4.4e-16**, matches single-device to **1.8e-12**, `--hvpBatch 1` agrees with 256 to 1.8e-12, `chisqFit` sharded matches to 2e-15. The two tests aimed at invisible failures (a parameter-only penalty must shift the sharded loss by *exactly* what it shifts the single-device loss; the subprocess test across two genuinely distinct logical CPU devices) are the right tests.

The core is sound. But the last commit is called "Stop multi-device fits silently computing the wrong thing", and there is one more of those, one hard crash, and a gap in the refuse-at-point-of-use list. **Five findings are inline** — two blocking (`--preconditionFrom gaussnewton` silently ignored; `--binByBinStatMode full` raises `AttributeError`), one stale docstring, and two significant (`--saveHists` not covered by the guards; the `systematic_type == "normal"` logk copy escaping the host pinning). The rest is below.

---

### Significant

**`_build_shards` transiently doubles device memory on every rebuild.** `_make_tf_functions` calls `_build_shards()` at the top, but `self._shard_graph_fns` isn't reassigned until line 595 — so the old shard tensors are still reachable through the old closures while the new ones are being allocated. Rebuilds happen on `arm_regularizers` with a regularizer attached, on `deepcopy` (the saturated-model path in `save_hists`), and on any re-`init_fit_parms`. For a PR whose entire purpose is memory, a 2x peak on the largest tensors is worth an explicit `self._shard_graph_fns = []` before building.

**Two I'd rather ask about than assert.**

- `hessian_from_hvps` is O(npar) graph evaluations, so the postfit covariance now scales linearly in parameter count where `tape.jacobian` was a single pass. Fine at 6538 — worth saying in `--hvpBatch`'s help so nobody is surprised at 50k.
- `_loss_val_grad_hessp_batch` (`sharding.py:704`) is the only combiner not wrapped in `tf.function`; the run emits `5 out of the last 5 calls to pfor.<locals>.f triggered tf.function retracing`. Deliberate, so `ResourceExhaustedError` surfaces and `k` can change without retracing? If so it deserves a comment, since every sibling is wrapped.

---

### Smaller, but each changes behaviour

- **`--hvpBatch` is a no-op on the single-device path.** `_hessp_batch` / `hessian_from_hvps` / `_hessp_batch_dispatch` live in `Fitter` but are only reachable through the subclass, while the help text reads "preconditioning and, on the multi-device path, the postfit Hessian" — i.e. as if preconditioning used it everywhere. Either move them into `sharding.py`, or wire them into the single-device preconditioner too — that would fix single-GPU OOM as well, which is arguably the larger win.
- **The OOM retry doesn't persist the reduced `k`** (`fitter.py:2412` is local to the call), so the postfit Hessian rediscovers the same OOM the preconditioner already solved for.
- **`--hvpBatch 0` silently becomes 256** — `int(getattr(options, "hvpBatch", 256) or 256)`.
- **The GPU-selection log line is emitted before `logging.setup_logger()`** (`rabbit_fit.py:854` vs `:883`), so `Selecting GPU(s) [...] by occupancy` is dropped at every verbosity — the one record of a consequential automatic choice. The block only has to precede TF *ops*, not the logger.
- **`pick_physical_gpus` now runs for every fit, including `--nDevices 1`**, so single-GPU behaviour on a multi-GPU node changes and an `nvidia-smi` subprocess joins startup. Probably an improvement, but it's a separate behaviour change riding in a multi-device PR, and the occupancy choice is racy — two fits launched together both pick the same idle GPU.
- **Nothing checks `len(--devices) == --nDevices`**; a mismatch falls through to `select_devices`'s round-robin warning, which is not what the help text promises.
- Guard coverage: the parametrised unsupported-step test uses `loss_val_valfull_grad_hess`, which has **no caller in `bin/`** — the real `--fullNll` entry point is `full_nll()`. That one *is* guarded, it's just the other one that's tested.

rabbit/sharding.py
davidwalter2: Confirmed and fixed in cc848a2 — your numbers reproduce exactly on the unit-test model:

```
sharded(gaussnewton) vs single(gaussnewton) : max|d| = 1125          (you: 1124.8)
sharded(gaussnewton) vs single(hessian)     : max|d| = 1.8e-12   <-- what it returned
```

Deleting the override is the right call and I took it verbatim. Checked why it works before doing it, since "the base method reaches the sharded machinery" is the load-bearing claim: the base calls `self.loss_val_grad_hess()`, which `_make_tf_functions` rebinds to the HVP assembler, and `self.expected_yield()`, which the subclass host-pins — so both branches stay distributed with no change. After deleting:

```
Fitter._reference_matrix(sharded, gaussnewton) vs single(gaussnewton): 2.7e-15
Fitter._reference_matrix(sharded, hessian)     vs single(hessian)    : 1.8e-12
```

Net -25 lines, and the stale docstring goes with it — you're right that it described the previous implementation and read as a placement claim.

`test_reference_matrix_honours_precondition_from` pins it over both values of the flag, mutation-checked: restoring the override fails it.

rabbit/sharding.py
davidwalter2: Confirmed and fixed. Reproduced the exact traceback — `bbstat.py:846`, `if self.indata.betavar is not None and full:`, evaluated before the `and` short-circuits:

```
nDevices=1 lite -> OK 20.1857      nDevices=2 lite -> OK 20.1857
nDevices=1 full -> OK 20.5098      nDevices=2 full -> AttributeError
```

Took both halves of your suggestion. `ShardIndataView` now exposes `betavar`, so the common `None` case just works, and `_build_shards` raises a clean `NotImplementedError` when it is non-`None` in full mode — agreed it is genuinely unshardable, the full-mode profile contracts over the bin axis with `einsum("ijk,jk->ik")`.

Sharded full mode now matches single-device exactly (20.5098 both ways), which is a better check than "does not crash". `test_bin_by_bin_stat_modes_match_single_device` pins both modes against the single-device loss; dropping the `betavar` line fails it.

rabbit/sharding.py
davidwalter2: Confirmed — `expected_events` is neither pinned nor refused, and `save_hists` calls it at three sites (`rabbit_fit.py:362, 514, 541`). Same tensor, same all-bins path, same end-of-a-multi-hour-fit failure.

I went with **pinning rather than refusing**, for `expected_events`, `expected_with_variance`, `expected_variations` and `chi2`. Refusing `--saveHists` would leave a sharded fit unable to write histograms at all, which seems like a large usability loss for a feature whose whole point is making big fits runnable — and the cost of pinning is one slow host pass per mapping, outside the minimiser loop, which is exactly the trade `expected_yield` already makes. Say if you would rather they were refusals; the two are a one-line swap.

You are right that the reasoning in `expected_yield`'s docstring simply was not carried across, so the new ones point back at it rather than restating it.

rabbit/inputdata.py
davidwalter2: Confirmed and fixed with the one-line `tf.device("/CPU:0")`. `_init_logk_scaled` had **zero** device scopes, and `fitter.py:1728` builds the full dense copy eagerly.

Worth recording why it is narrow, since I had to check: the `log_normal` branch aliases `indata.logk` rather than copying, so it inherits the `host_memory=True` placement — only `systematic_type == "normal"` with a non-trivial param model materialises anything, exactly as you say.

davidwalter2:
Thanks — this was a careful review and every load-bearing claim held up when I checked it. All of it is addressed in `cc848a2`; the four inline threads have the details and the numbers.

Both blocking findings reproduced exactly as you described, and the two fixes are pinned by tests that I mutation-checked rather than just watched pass — restoring the `_reference_matrix` override fails one, dropping `betavar` fails the other. Sharded `--binByBinStatMode full` now agrees with single-device to the digit (20.5098 both ways), which is a better assertion than "does not crash".

### The rest of the list

- **`_build_shards` 2x peak** — fixed, and you were right that it is worth an explicit reset: `self._shard_graph_fns = []` and `self.shards = []` now happen immediately before the new generation is built, after the refusals.
- **`systematic_type == "normal"` logk copy** — fixed, one line, details in that thread.
- **`--saveHists`** — pinned rather than refused; reasoning in that thread, and happy to swap it if you disagree.
- **`--hvpBatch 0` -> 256** — fixed to test `None`. Same falsy-zero shape as the `kmax` / `cg_maxiter` defaults you caught in #153, which is a sign the pattern is worth grepping for repo-wide rather than fixing one at a time.
- **GPU-selection log dropped** — fixed properly rather than by moving the block: `pick_physical_gpus` has to run before TF may touch a GPU, which is genuinely before the logger exists, so it now buffers its messages and the driver drains them after `setup_logger`. That also keeps the `nvidia-smi`-failure warning, which was being dropped too.
- **`len(--devices) != --nDevices`** — now raises instead of falling through to the round robin.
- **The guard test** — retargeted at `full_nll`. You are right that the guard itself was fine and only the test pointed at the wrong method; `loss_val_valfull_grad_hess` has no caller in `bin/` at all.
- **`--hvpBatch` help** — corrected to say only the multi-device path assembles the Hessian from HVPs, and to note the O(nparams) graph evaluations against the jacobian's single pass, as you suggest.
- **`_loss_val_grad_hessp_batch` unwrapped** — deliberate, and now commented: the halving retry needs `ResourceExhaustedError` to surface from the eager call, and `k` changes between calls so a wrapper would retrace per batch width anyway. The retracing warning is expected.

### Two I have not changed

**`--hvpBatch` being a no-op single-device.** Confirmed — `hessian_from_hvps` has callers only in `sharding.py`. I have only corrected the help for now. Wiring it into the single-device preconditioner is, as you say, arguably the larger win (it would address single-GPU OOM), but it is a behaviour change for every existing preconditioned fit and wants its own benchmark. Would rather do it separately than smuggle it in here.

**`pick_physical_gpus` running at `--nDevices 1`.** You are right that it is a separate behaviour change riding along, and right that the occupancy choice is racy — two fits launched together both pick the same idle GPU. I have left it, because gating it on `nDevices > 1` would mean a single-GPU fit on a shared node keeps landing on GPU:0 next to whoever is already there, which is the problem it was added for. But it should be argued on its own terms rather than inherited from a multi-device PR, so: happy to split it out if you would prefer, and the race deserves a note in the help either way.

### One thing to flag about the suite

`test_snapshot.py::test_sigterm_writes_a_snapshot_before_the_process_dies` cost me a while. It failed 3/3 on this branch while passing 3/3 on `main` at a similar load, which looked like a real interaction — until I interleaved the runs instead of running them in sequence, at which point both passed 4/4 at a *higher* load. Sequential comparison on a shared machine was simply not a controlled experiment.

So your original read is right: it is the hardcoded `time.sleep(2.0)` before the signal, not anything in this branch. But it fails often enough to have sent me chasing a phantom regression twice, so it is worth replacing the sleep with a poll on the snapshotter being ready. Happy to do that as a separate one-liner PR since it is not this branch's problem.

Full suite green apart from that one (`170 passed`), black/isort clean, no new flake8 `F` findings.

rabbit/sharding.py
```diff
@@ -0,0 +1,854 @@
...
+            )
+
+        devices = select_devices(self.n_devices)
+        edges = shard_edges(self.indata.nbins, self.n_devices)
+        logger.info(
+            f"Sharding {self.indata.nbins} bins over {self.n_devices} "
+            f"device(s): {[f'{d}:[{a},{b})' for d, (a, b) in zip(devices, edges)]}"
+        )
+
+        self._shard_graph_fns = []
```
lucalavezzo, line 370: **The 2x peak is still there — the holder isn't the graph fns.**

Thanks for taking this, but resetting `shards` / `_shard_graph_fns` isn't what was keeping the old generation alive. I weakref'd the per-shard `logk` tensors and checked liveness at the moment the new generation exists, after exercising the graphs (the only state the rebuild paths actually occur in — `loss_val` / `loss_val_grad` / `loss_val_grad_hessp` / `_profile_beta` all called first):

```
as-merged (shards/_shard_graph_fns reset)    old alive when new exists: [True, True]
+ pop _DYNAMIC_TF_FUNCS                      old alive when new exists: [True, True]
+ pop _DYNAMIC_TF_FUNCS and _profile_beta    old alive when new exists: [False, False]
```

The holder is **`self._profile_beta`** — the instance-level `tf.function(_profile_beta_sharded)` set at line 854, whose traced graph captures every shard's `logk`. It is not in `_DYNAMIC_TF_FUNCS` (line 273), so clearing that set changes nothing on its own; and `loss_val` and friends turn out not to hold the tensors at all, which is why the reset at line 348 reads as if it should work.

One line at the top of `_make_tf_functions`, before `self._build_shards()` at line 595:

```python
for name in self._DYNAMIC_TF_FUNCS | {"_profile_beta"}:
    self.__dict__.pop(name, None)
```

gives `[False, False]`, and the loss is bit-identical afterwards (20.185670 either way). The `shards = []` / `_shard_graph_fns = []` reset then becomes redundant, though harmless.

Two things this suggests beyond the one-liner:

- `_profile_beta` arguably belongs *in* `_DYNAMIC_TF_FUNCS` — it is instance-level dynamic machinery capturing shard state, exactly what that set is for. Right now `__deepcopy__` only handles it by accident, via the separate `jit_overrides` scan that happens to catch it because the base class attribute is a `tf.function`.
- Worth asserting rather than reasoning about: a test that weakrefs `shards[0].logk`, forces a rebuild, and asserts the old tensor is dead would have failed on the as-merged version. Happy to hand over the probe script if useful.

davidwalter2: You're right, and the reset I added did nothing for the peak. Fixed in `067d1b7`.

I reproduced your table once I sampled at the right moment, which is what I had wrong the first time: I was checking liveness *after* `_make_tf_functions` returned, where the old generation is unreachable either way and the check passes vacuously. Sampling inside `_build_shards`, before the new tensors are allocated:

```
as merged (shards/_shard_graph_fns reset)   old alive: [True, True]
pop _DYNAMIC_TF_FUNCS only                  old alive: [True, True]
pop _DYNAMIC_TF_FUNCS + _profile_beta       old alive: [False, False]
```

Identical to yours. `self._profile_beta` is the holder, exactly as you say — a live reference across the whole of `_build_shards`, since it is only replaced at the end of the method.

**One refinement**: the pop alone was not enough here. Splitting your one-liner into its two effects:

```
pop attributes only        old alive: [True, True]
pop attributes + collect   old alive: [False, False]
```

What remains after dropping the last strong reference is cyclic — a `tf.function` graph references its captures and its siblings — so refcounting cannot reclaim it and the device memory is held until the generational collector happens to run. Possibly your probe collected, or a TF version difference; either way both steps are in now.

Took your second suggestion too: `_profile_beta` joins `_DYNAMIC_TF_FUNCS` rather than being special-cased at the call site. Agreed that is where it belonged for `__deepcopy__` as well.

And your third: `test_rebuild_frees_the_previous_shard_generation` asserts it rather than leaving it to reasoning. It samples inside `_build_shards` and deliberately does **not** collect in the probe — my first version did, which let it pass with the `gc.collect()` removed from the production path, i.e. exactly the vacuous-test shape. Mutation-checked both ways now: dropping either the collect or the `_profile_beta` registration gives `assert [True, True] == [False, False]`.

Loss bit-identical across the rebuild (20.185670 either way), `172 passed`, lint clean. No need for the probe script, but thanks for offering — the weakref-at-the-right-moment framing was the useful part.

Separately, `a5c4e1a` strips the development narrative out of the comments and test docstrings on this branch: which override was removed, which mode used to raise, what each intermediate probe measured. A superseded explanation of the shard release had also been left in `_build_shards` contradicting the accurate one. They describe the current invariant now.

davidwalter2:
Taken in `9e93ade` — the pin is now gated on `n_devices > 1`, and you're right that `MultiDeviceFitter` sets `n_devices` before `super().__init__` precisely so the base constructor can read it.

The placement half reproduces exactly as described:

```
as committed   self.logk on CPU:0   while indata.logk is on GPU:0
gated          self.logk on GPU:0
```

**The consequence does not, though.** Measured on a normal-systematic model (10000 bins, 2048 systs, 0.82 GB logk) on a V100, synchronising the device rather than timing the enqueue — eager GPU ops are async, and without a `.numpy()` both variants read 0.3 ms, which is just dispatch:

```
                 loss_val_grad    hessp
as committed     2.8 ms           3.5 ms
gated            2.8 ms           3.5 ms
```

Identical. There is no per-evaluation transfer: TF copies the captured host tensor to the device once and reuses it inside the compiled function. If it were paid per call, 0.82 GB over PCIe would be ~80 ms and would dominate the 2.8 ms completely — it isn't there at all.

So the fix is right, for a narrower reason than stated: the unconditional scope is an unneeded change to the default path, and it leaves the tensor resident on both host and device, which matters at the sizes this class exists for. Not a throughput regression on every likelihood evaluation.

Worth flagging since the same reasoning shape will recur: "captured host tensor inside a `tf.function`" does not imply a per-call copy, and the size figure in a claim like this is worth checking against a synchronised measurement before it sets the severity.

davidwalter2:
Confirmed and fixed in `028adff`.

`rabbit_limit.py` does build on `parsing.common_parser()` (line 25) and constructs `fitter.Fitter` directly (line 270), with no `make_fitter` and no `pick_physical_gpus` — so `--nDevices 4` parsed and ran single-device with nothing said.

I took the **reject** option rather than wiring it through `make_fitter`, because sharding is not merely unimplemented there: limit scans generate toys (`rabbit_limit.py:333` calls `toyassign`), and toy generation is one of the all-bins paths multi-device refuses. Routing it through `make_fitter` would construct a `MultiDeviceFitter` that then declined at the first toy — later, and less clear, than declining at the flag. It now raises up front naming both the flag and the reason.

One piece of feedback on the review output rather than its content. This run reported **0 critical, 0 important, 1 minor** and still concluded:

> **Status**: Patch has blocking issues

A lone P3 — a flag being inert in a sibling script — is not blocking, and the same verdict wording appeared on the run that carried a genuine P2. If the verdict maps "any findings" to "blocking", the line carries no information and readers will start ignoring it, which is a shame because the finding severities themselves have been well calibrated. Worth a threshold, e.g. blocking only at P0/P1/P2.

For the record, the diff-embedding fix is visibly working on this PR: `headSha 9e93ade`, `diffFiles 7`, `diffBytes 81773`, and the token count has come down from 3.18M on the empty-diff run to 1.81M here for a larger reviewed surface.

davidwalter2:
Both correct, fixed in `428ec84`.

**P2** — `loss_val_grad_hess_beta` was indeed the one post-fit all-bins path still inherited: `rabbit_fit.py:635` calls it under `bbstat.enabled and diagnostics`, and `grep -c loss_val_grad_hess_beta rabbit/sharding.py` was 0. It refuses through `_unsharded` now, and `--diagnostics` (with BBB on) joins the up-front list so the run stops in seconds.

One correction on the evidence: it does **not** raise an allocation failure on a small model — I ran it sharded and it completes. That is true of every entry in that refusal list, though; they all succeed on a toy and fail at the sizes `--nDevices` exists for, which is exactly why they refuse at the point of use instead of being left to fail naturally. So the finding stands, the stated symptom just is not reproducible at test scale.

**P3** — confirmed, and slightly worse than described. Tracing both layers:

```
--nDevices  0: visible GPUs -> []          make_fitter n_devices -> 1
--nDevices -1: visible GPUs -> [0, 1, 2]   make_fitter n_devices -> -1
```

You flagged that 0 hides every GPU while the fitter builds single-device, which is right. `-1` also fails to normalize: `int(-1 or 1)` is `-1`, not `1`, so the class carries a negative device count while one GPU goes missing from the visible set. `pick_physical_gpus` raises for `n < 1` now.

Both pinned by tests, mutation-checked — removing either guard gives `DID NOT RAISE`. 178 passing.

Following up on the verdict point from the previous round: this run is a fair "blocking" (a genuine P2), so the label is right here. The calibration concern was the run before, where a lone P3 got the same wording.

davidwalter2:
Confirmed, and fixed in 53f9d69 — thanks, this one was against my own new test and it was right.

Reproduced: `global_asym_impacts.py:134` calls `fitter._dxdvars()` under `linear_warmstart`; `_dxdvars` evaluates `self._compute_loss()` and then takes `t2.jacobian(grad, [x0, nobs, beta0])`; and `MultiDeviceFitter` overrides none of `_dxdvars`, `_compute_loss` or `global_asym_impacts_parms`. The helper's own comment already says what it costs — *"the same cost as one `--gaussianGlobalImpacts` call"* — and that call is exactly what `sharding.py:518` refuses.

It is not wrong, it is unsharded: on the test workspace `_dxdvars()` agrees to 1.1e-16 between 1 and 2 devices. So it is the usual shape — succeeds on a small model, dies at the sizes `--nDevices` exists for, after the minimiser has finished.

Fixed with both halves of the pattern already used here:

- `MultiDeviceFitter._dxdvars` refuses via `_unsharded`, so the library is safe from any entry point;
- `--globalAsymImpacts` **with** `--globalAsymImpactsLinearWarmstart` joins the up-front `_md` list in `rabbit_fit.py`, so it fails at second zero instead of after the fit.

One deviation from your suggestion: I kept `global_asym_impacts_parms` in `_SHARDED_SAFE` rather than dropping it. Without the warm start it is only repeated `minimize()` calls and does run sharded, so refusing the whole method would break a working combination. The refusal sits on the primitive, and the safe-list entry now carries that as its stated reason.

The more useful part of this finding is the hole it located in the test itself. `test_every_driver_called_fitter_method_is_classified_for_sharding` walks driver → `Fitter`; it cannot see driver → helper module → `Fitter`, which is precisely the edge `_dxdvars` sits on. My hand-check of the helpers grepped for `_compute_nll`, `_compute_yields_with_beta` and `.jacobian(`, so a `Fitter` method reaching `_compute_loss` slipped through. Added `test_every_fitter_method_reached_from_the_impacts_helpers_is_classified` — the same enumeration one level down, so the safe list is checked rather than asserted. Enumerating that layer turns up exactly five `Fitter` methods (`minimize`, `contour_scan`, `freeze_params`, `defreeze_params`, `_dxdvars`), of which only `_dxdvars` was unclassified.

Mutation-checked three ways: removing the `_dxdvars` override fails both the refusal test and the new enumeration test; adding a `fitter._compute_nll()` call to a helper fails it; emptying `_HELPER_SAFE` fails it. 183 passed, lint clean.

davidwalter2:
You're right, and this one was my regression — fixed in 8b12b2f.

Verified the chain: `_dndvars` (fitter.py:1448) calls `self._dxdvars()` **unconditionally** at line 1461, and `chi2(profile=True)` → `_residuals_profiled` (1902) → `_dndvars` (1918). `MultiDeviceFitter.chi2` host-pins but still dispatches through the override, so `tf.device("/CPU:0")` does nothing to save it. Same for `_expected_with_variance` at 1606. So every sharded `--saveHists` would have raised `NotImplementedError` while writing output, naming a flag the user never set.

Fixed the way you suggested: `_dxdvars` is host-pinned like its callers rather than refused. The `--globalAsymImpacts` + `--globalAsymImpactsLinearWarmstart` entry in the up-front `_md` list stays, which is where an entry-point refusal belongs — the mistake was putting one on a shared primitive as well.

What I think is the real lesson here: **the suite had no test that a supported sharded path still runs.** Everything in `test_sharded_fit.py` asserted that unsupported paths *refuse*; nothing asserted that the kept-working ones still work. That asymmetry is why a regression this blunt passed all 183 tests. Two tests added:

- `test_profiled_chi2_still_works_when_sharded` — `_dxdvars` must run sharded and agree with single-device. (Honest limitation, stated in the docstring: CI has no GPU, so dropping the `tf.device("/CPU:0")` is invisible to it, same as for the other pinned methods.)
- `test_no_refusal_is_reachable_from_a_path_that_is_kept_working` — walks `self.*` calls in `fitter.py` transitively from each host-pinned entry point and asserts none reaches a method `MultiDeviceFitter` refuses. Restoring the bad refusal makes it report exactly your three: `['chi2 -> ... -> _dxdvars', 'expected_events -> ... -> _dxdvars', 'expected_with_variance -> ... -> _dxdvars']`.

That pairs with the enumeration tests from the last two rounds: those catch a *missing* refusal, this one catches an *over-broad* one. Mutation-checked (restore the refusal → both new tests fail; drop the pin → the numeric one passes, as documented). 184 passed, lint clean.

davidwalter2:
Both confirmed and fixed in c79e014.

**P3-1 — wrong flags in the refusal messages.** Verified by building the real parser: `make_parser()` in `rabbit_fit.py` yields 105 options and `--impactType` is not one of them. It comes from `parsing.add_impact_args`, which only `rabbit_print_impacts.py` and `rabbit_plot_pulls_and_impacts.py` call — `rabbit_fit.py` uses `parsing.common_parser()` alone. And `gaussian_global_impacts_parms` is gated by `args.gaussianGlobalImpacts`, not `--doImpacts`. Both messages now name the option that actually reaches the path.

Worth adding: **the flag string was decorative.** `test_unsharded_postfit_steps_fail_before_the_fit_not_after` carried it as a parametrize argument but matched only on `"multi-device mode"`, so no wrong name could ever fail. It now matches the flag, and `test_refusal_messages_name_flags_the_fit_script_has` checks every flag named in an `_unsharded` call against the parser `rabbit_fit.py` actually builds — introspecting it rather than grepping for quoted strings, since a grep over `parsing.py` would have counted `--impactType` as known and missed exactly this. Restoring the old string makes it report ``['--impactType']``.

**P3-2 — `--devices` ignored by `rabbit_limit.py`.** Right, and my own comment above the guard said both flags were inert while the guard only covered one. `--devices 2` would have run on the default GPU, colliding with whatever the user was stepping around, which is the only reason to pass it. Now refused, pointing at `CUDA_VISIBLE_DEVICES` as the working alternative. The script had no tests at all, so both halves are now covered.

187 passed, lint clean, each guard mutation-checked.

One note on the verdict rather than the findings: this came back **"Patch has blocking issues"** on two P3s, with the summary itself confirming every prior-round finding is addressed. Minor-only rounds reading as blocking makes the status field less useful for deciding when a PR is done — the issue list was accurate and worth acting on, but I'd read the severity, not the verdict.

davidwalter2:
Confirmed and fixed in 519d014. This is a real P1 — reproduced the crash, and it would have hit the headline use case.

`get_poi()` reads `_blinding_offsets_poi_add` after the merge; the three sites that thread offsets to the duck-typed evaluators knew about only two. A sharded fitter with blinding armed dies at the first `loss_val_grad`:

```
AttributeError: 'ShardEvaluator' object has no attribute '_blinding_offsets_poi_add'
```

And the reachability claim is right: blinding is armed by default for an observed-data fit and is not in the up-front `_md` refusal list, so `--nDevices N` on data — the reason this PR exists — was broken. The offsets are read whatever their value, so an all-identity run would not have dodged it either.

**The coverage gap is the part worth recording.** Every sharded test built through `_make_fitter`, which hardcoded `do_blinding=False`, so all 35 of them ran with `get_poi`'s blinded branch never executed. That is why the merge looked clean. `_make_fitter` can arm it now, and the sharded loss and gradient are compared against single-device with the offsets armed, over 2 and 3 devices.

**Fixed as a list rather than a third assignment.** The three sites each carried their own copy of the offset names, which is exactly what let one fall behind without a sound. They now iterate a single `_BLINDING_OFFSET_ATTRS`, and `test_every_blinding_offset_is_threaded_to_the_shards` pins that tuple against what a blinded `Fitter` actually creates — so adding an offset, *or dropping one*, fails in the suite instead of in a fit. That second direction matters here: #174 removes `_blinding_offsets_poi` entirely, and this test will catch the tuple going stale when that lands.

Mutation-checked: removing the add offset from the tuple reproduces the original `AttributeError` verbatim, and stripping the global-view threading fails 2 of the 3 blinding tests.

277 pass. One unrelated failure, `test_snapshot.py::test_sigterm_writes_a_snapshot_before_the_process_dies` — it fails reproducibly in isolation here, so not a load flake, but it fails the same way on unmodified `main` (`8898336`), so it is pre-existing and not from this branch or the merge.

rabbit/inputdata.py
lucalavezzo, line 15: **The #152 merge dropped the pin that backs this, and a new full-size tensor took its place.**

`9e93ade` pinned the rnorm-scaled `logk` copy so a sharded fit could not materialise it on the default device before any shard exists — the finding this `host_memory=True` pinning is half of. The merge replaced `_init_logk_scaled` with main's version and the pin went with it: `grep -c 'tf.device' rabbit/fitter.py` is now **0**.

Main no longer copies `logk`, which is better — but it materialises a new full-size tensor in its place (`fitter.py:2063`):

```python
rnorm_init = tf.broadcast_to(rnorm_init, [self.indata.nbinsfull, self.indata.nproc])
```

`broadcast_to` is not lazy, so for `systematic_type == "normal"` that is a real `[nbinsfull, nproc]` allocation made in `Fitter.__init__`, on the default device, before any shard exists. `_build_shards` then slices it under `/CPU:0` — a device→host copy of the whole thing — and `self.rnorm_init` stays resident on GPU:0 for the life of the fit. Smaller than `logk` by a factor `nsyst`, so not fatal on its own, but it is the same failure mode and it breaks exactly the invariant this pinning and the host-side slicing in `_build_shards` were established to hold.

Fix is the gate that was already there, moved onto the broadcast: `tf.device("/CPU:0")` when `self.n_devices > 1` — `n_devices` is set before `super().__init__` precisely so the base constructor can read it — and only on this branch, since the `log_normal` path returns before it.

rabbit/sharding.py
```diff
@@ -0,0 +1,988 @@
...
+    """
+
+    def __init__(self, indata, start, stop, device):
+        self.start = int(start)
+        self.stop = int(stop)
+        nb = self.stop - self.start
+        with tf.device(device):
+            self.norm = tf.identity(indata.norm[self.start : self.stop])
+            self.sumw = tf.identity(indata.sumw[self.start : self.stop])
+            self.sumw2 = tf.identity(indata.sumw2[self.start : self.stop])
```
lucalavezzo: **These slices execute on the shard's device, so each one transiently copies the whole tensor there.**

`indata.norm[start:stop]` is evaluated inside `with tf.device(device)`, so the StridedSlice is placed on that device and TF has to copy its input — the entire `[nbinsfull, nproc]` `norm` — there first. Same for `sumw` and `sumw2`, and again for every shard. The `logk` / `rnorm_init` block at line 409 and the `nobs` / `lognobs` / `varnobs` block at line 450 both deliberately slice under `/CPU:0` for this exact reason, and the comment there says so ("defeating both that pinning and the host-side slicing").

Confirmed the mechanism rather than asserting it, with two logical devices so it needs no GPU:

```
big on            /device:CPU:0
slice-in-scope    /device:CPU:1   <- StridedSlice ran here, so `big` had to be copied here
slice-then-move   /device:CPU:0 -> /device:CPU:1
```

On CPU→CPU the copy is free, which is why the suite cannot see this and why it only bites on the hardware the feature is for. The fix is the pattern already used below: slice under `/CPU:0`, then `tf.identity` the slice inside the device scope.

bin/rabbit_fit.py
```diff
@@ -940,6 +940,66 @@ def main():
...
+    chosen = pick_physical_gpus(args.nDevices, explicit=args.devices)
+    if chosen is not None:
+        tf.config.set_visible_devices(chosen, "GPU")
+
+    # Multi-device incompatibilities that the sharded Fitter would otherwise
+    # only discover at the point of use -- i.e. after the minimiser and the
+    # postfit Hessian have already run, taking the completed fit down with
+    # them. Checked here so the run ends in seconds with the flags named.
+    if args.nDevices > 1:
+        _md = []
```
lucalavezzo, line 963: **Question: should the Hessian-per-iteration minimizers be on this list?**

`--minimizerMethod trust-exact`, `dogleg` and `tf-trust-exact` hand the minimizer `scipy_hess` (`fitter.py:2907`) / `native_closure` (`:2927`), and both call `self.loss_val_grad_hess()` **once per iteration**. On `MultiDeviceFitter` that is `_loss_val_grad_hess_hvp` → `hessian_from_hvps()`: all `npar` HVPs, i.e. 26 batched graph calls at 6538 parameters and the default `--hvpBatch 256`, *per iteration* — where `trust-krylov` (the default) pays one HVP per CG step. `--asymImpacts` has the same shape: `contour_scan`'s `hess_mode` defaults to `"exact"` (`rabbit_fit.py:251`), which rebuilds the dense Hessian once per contour iterate (`fitter.py:3462`).

I am genuinely unsure this belongs here, which is why it is a question rather than a finding. Nothing is *wrong* — it converges to the same answer — so it is not an incompatibility in the sense of the other entries. But it is only discoverable by watching a fit fail to finish, the ratio is large enough to look like a hang, and it is invisible from the flags, which is the situation this block exists to remove. Refusal, a warning at `fit()` time, or a sentence in `--nDevices`'s help would all cover it; do you have a preference? Single-device is unaffected either way.

bin/rabbit_fit.py
```diff
@@ -940,6 +940,66 @@ def main():
...
+            # compute_curvature calls fitter._compute_nll and
+            # _compute_yields_with_beta directly and takes a dense full-bins
+            # jacobian; MultiDeviceFitter overrides none of those, so the
+            # curvature would run unsharded -- and for --lCurveScan it runs
+            # after each per-tau minimize, discarding finished fits mid-scan.
+            _md.append("--lCurveScan / --lCurveOptimize")
+        if args.diagnostics and not args.noBinByBinStat:
+            # loss_val_grad_hess_beta takes a jacobian over the full-length
+            # ubeta on one device; see MultiDeviceFitter.
+            _md.append("--diagnostics (unless --noBinByBinStat)")
```
lucalavezzo, line 1000: **`--covarianceFit` and sparse inputs are the two entries missing here, and they refuse after the expensive part.**

They are the two combinations the PR description lists as "checked at construction", and construction turns out to be late. `Fitter.__init__:232` builds `tf.linalg.diag(1.0 / data_var)` when `--covarianceFit` is set with no `data_cov_inv` in the workspace — a dense `[nbins, nbins]`, ~68 GB at the 92144 bins this PR is sized for — while `_build_shards` does not raise until `sharding.py:375`, reached from `init_fit_parms` at `:258`. So the user gets an allocation failure instead of the refusal that names the flag. A sparse input is milder but the same ordering: the whole file loads first.

Two more `_md.append` lines and both fail in seconds with the rest.

rabbit/regularization/regularizer.py
```diff
@@ -21,6 +38,10 @@ def set_expectations(self, initial_params, initial_observables, parms=None):
     def compute_nll_penalty(self, params, observables):
         """
         Compute the penalty term that gets added to -ln(L), this function should be called in each step of the minimization
+
+        ``observables`` is None when ``needs_observables`` is False, so that a
+        regularizer which declares it does not need them cannot quietly start
+        using them (it would raise rather than read a truncated vector).
```
lucalavezzo, line 44: **This contract isn't honoured by the L-curve path.**

`Fitter._compute_nll` and `arm_regularizers` both pass `None` now, but `regularization/lcurve.py` was not updated: `:40` still builds the yields with `full=len(fitter.regularizers)`, and `:52` hands `nexpfullcentral` to every penalty regardless of what it declared.

Multi-device refuses `--lCurveScan` / `--lCurveOptimize` up front, so this is a single-device-only inconsistency — but it is the same contract, introduced in the same change, and `tests/test_sharded_fit.py::_ParamOnlyPenalty` is exactly the class of regularizer it concerns. One written to this docstring either raises inside the L-curve (if it checks `observables is None`, which the docstring invites) or quietly starts reading a vector the fit path never gives it, at which point the scan objective and the fit objective are no longer the same function and the chosen tau is optimising something else.

Same two-line shape as the `_compute_nll` fix: `full=self._regularizers_need_observables()` at `:40`, and the `None`-or-yields conditional at `:52`.

lucalavezzo (commented):
Re-reviewed at `f1ac790`, in a clone rather than the prod checkout. #153 has landed, so the diff is sharding-only now. All 32 CI jobs green; I also ran `tests/test_sharded_fit.py` in the container here — **41 passed, 87 s**.

Everything from my last round is addressed, and I checked the load-bearing ones rather than taking them on the commit message: the `_reference_matrix` override is gone (so `--preconditionFrom gaussnewton` reaches the sharded HVP assembler), `betavar` is exposed, the four all-bins postfit paths are host-pinned, and the shard generation is released before the next is built. The three enumeration tests added since (driver → `Fitter`, impacts-helper → `Fitter`, and "no refusal is reachable from a path that is kept working") are the right answer to this PR's failure mode — the last one in particular turns "did we over-refuse?" from a review question into a test.

**Six findings inline.** The first is a regression the #152 merge introduced: it replaced `_init_logk_scaled` wholesale and the host pin from `9e93ade` went with it, while main's version materialises a new full-size tensor on the default device in its place. The rest are a device-scope slice that copies what it is trying to avoid copying, two gaps in the up-front refusal list (one of them posed as a question, since it is a performance cliff rather than an incompatibility), the `needs_observables` contract not reaching the L-curve path, and the HVP batch width still not persisting across calls.

One thing I checked and found clean, since it looked like the most likely place for a silent wrong answer: gamma + `--binByBinStatMode full` is reachable sharded, and that mode profiles β with a stateful Newton loop (`self.nbeta.assign_sub` inside a `tf.while_loop`, `bbstat.py:949`) which `_loss_val_grad_hessp_batch` runs k directions of inside a single `tf.vectorized_map` against one shared variable. It is correct: sharded vs single-device dense Hessian max|Δ| = 1.8e-12 at `--hvpBatch 256`, 1.4e-14 at `--hvpBatch 1`, asymmetry ≤ 7e-15, and no pfor while-loop fallback. The reason it is safe is worth writing down — the Newton state depends only on `x`, which is identical across the k directions, so the k solves cannot disagree. But `test_bin_by_bin_stat_modes_match_single_device` compares `loss_val` only, so nothing pins it; that test is one line from covering the Hessian too, and that is where the sharing actually happens.

And one correction to my own last round: the `betavar` + full-mode refusal I asked for is stricter than it needs to be. `bbstat.py:846` reads `betavar` only under `and full`, the sharded loss only ever evaluates `full=False`, and the `full=True` paths run host-pinned on the intact `self.indata` — so the term never enters the sharded likelihood and the refusal removes a combination that would have worked. Safe as it stands, and the stated reason ("it cannot be split along with the rest") is true of `betavar`; it just is not something the sharded loss needs. Your call whether to keep the caution.

rabbit/fitter.py
```diff
@@ -2657,6 +2714,90 @@ def _reference_matrix(self):
...
+                    # The per-batch memory is k * nbins * 9 * 8 in principle,
+                    # but the loss graph holds several bins-shaped tensors at
+                    # once -- measured at ~6.5x that on one model -- so the
+                    # right k is model dependent. Halve and retry rather than
+                    # making the caller guess: this runs once per build, so a
+                    # couple of wasted attempts cost far less than falling back
+                    # to an unpreconditioned fit or losing the covariance.
+                    if k == 1:
+                        raise
+                    k = max(1, k // 2)
```
lucalavezzo, line 2795: **The reduced `k` doesn't survive the call, so the next assembly re-discovers the same OOM.**

`k` is a local. After the preconditioner's `_reference_matrix` has walked 256 → 128 → 64 (two failed allocations, each one a compile + allocate + unwind), the postfit `loss_val_grad_hess` calls `hessian_from_hvps` again with `k = self.hvp_batch = 256` and pays for the same two failures, as does every preconditioner rebuild in the restart loop. The warning even tells the user to pass `--hvpBatch 64` next time, which is information the object has just derived and thrown away.

`self.hvp_batch = k` after a successful shrink, and the auto-tuning becomes a property of the fit rather than of each call.

(Raised last round; the reply covered the `--hvpBatch 0` falsy-default but not this one, so I assume it was missed rather than declined.)

davidwalter2:
All six addressed in `2f0c4af`, each reproduced before fixing. Thanks for re-reviewing in a clean clone — finding 1 is a regression I introduced and did not notice.

**1. The host pin, lost in the #152 merge.** Confirmed: `grep -c 'tf.device' rabbit/fitter.py` was **0**. I resolved that conflict by taking main's `_init_logk_scaled` verbatim and reasoned that with no copy left there was nothing to place — which was right about `logk` and wrong about what replaced it. Measured `broadcast_to`: **6.4 MB materialised from a 32-byte source**, so it is a real allocation, in `__init__`, on the default device. Gated on `n_devices > 1` again.

**2. Slicing inside the device scope.** Confirmed with two logical devices, same setup you used:

```
big on            /device:CPU:0
slice-in-scope    /device:CPU:1   <- StridedSlice ran here, so `big` was copied here
slice-then-move   /device:CPU:0 -> /device:CPU:1
```

`norm`, `sumw`, `sumw2` now slice under `/CPU:0` and `tf.identity` into the device scope, matching the `logk` and `nobs` blocks. Your point that CPU→CPU makes this free — and therefore invisible to the suite — is the uncomfortable part: no test I can write on this runner would have caught either of these.

**3. Hessian-per-iteration minimizers — a warning, not a refusal.** They converge to the same answer, so an entry in `_md` would put something in a list whose whole value is that everything in it is a hard no. Help text alone does not cover your actual objection, that it is invisible from the flags. So it warns before the minimiser starts, with the numbers for the model in hand: parameter count, batch count, current `--hvpBatch`, and the alternatives. Covers `trust-exact`, `dogleg`, `tf-trust-exact` and `--asymImpactsHess exact`.

**4. Late refusals.** `--covarianceFit` joins the up-front list. The sparse one **cannot** — `indata` does not exist at that point in `rabbit_fit.py` (the block is at :962, the load at :1064), and sparseness is a property of the file. It now refuses immediately after the load, which is still ahead of the Fitter and the dense diag. Two lines, but not two `_md.append` lines.

**5. The L-curve contract.** Fixed at both sites. Worth recording: **nothing in the suite executed `lcurve.py` at all** — I checked before assuming the fix was covered, and it was not. `test_lcurve_honours_the_needs_observables_contract` now runs `_compute_curvature` with the existing `_ParamOnlyPenalty`, which asserts `observables is None` on every call. It fails against the pre-fix code.

**6. `hvp_batch` persistence.** Missed, not declined — you were right. `self.hvp_batch = min(self.hvp_batch, k)`, so an explicit oversized `batch=` that shrinks cannot raise a lower standing default.

**On the two extras:**

`test_bin_by_bin_stat_modes_match_single_device` now compares the dense Hessian and its symmetry, not just the loss. You are right that this is where the sharing happens and that the loss comparison left it unpinned. A 1e-6 perturbation of the sharded Hessian fails both modes; asymmetry is checked because interference between the k directions would be direction-dependent and show up there first.

**The `betavar` refusal I have kept**, with the comment rewritten to say what is actually true rather than implying necessity. Your analysis is right — the term does not enter the sharded likelihood, so this refuses something that would work. I kept it because `ShardIndataView` hands every shard the whole unsliced bin × bin tensor, which is correct only while nothing reads it, and if that changes the failure is a wrong answer rather than an error. Happy to drop it if you would rather have the combination; it is one `if`.

Full suite 286 passed, lint clean.

davidwalter2:
Confirmed and fixed in the latest push — this one is a real gap in my own fix from `2f0c4af`.

I gated the L-curve pass-down on `fitter._regularizers_need_observables()`, which is an `any()` over the whole list. The contract is **per regularizer**: `Fitter._compute_nll` and `arm_regularizers` both test each reg's own `getattr(reg, "needs_observables", True)`. So a mixed list — one yield-reading penalty, one parameter-only — still handed the parameter-only one the vector. The same bug I was fixing, one list longer.

The `any()` stays where it belongs, on `full=`: that decides whether the full yields are *built*, and one reader is enough to require it. Only the pass-down is per-regularizer now, mirroring the fit path exactly.

**Worth naming why my test missed it.** It attached a single `_ParamOnlyPenalty`, so `any()` was False and a global flag behaved identically to a per-regularizer one — a one-element list cannot distinguish them. It now attaches a `_YieldReadingPenalty` alongside, which asserts it *is* given observables, making the list mixed:

| lcurve version | mixed-list test |
|---|---|
| original (yields to everyone) | fails |
| `2f0c4af` global `any()` | **fails** |
| per-regularizer (now) | passes |

That middle row is the one the previous test could not produce.

Full suite 286 passed, lint clean. CI was already 32/32 green on `2f0c4af`; this is on top.

PR #156, opened 2026-09-09
preconditioner: spectral transform as an option, block-membership logging, and a ridge fix

lucalavezzo:
Three changes from using `--precondition` on a 3720-parameter fit. The new
behaviour is **opt-in** — the parsed default is `preconditionTransform='ridge'`,
so nothing changes for existing callers, and the default is pinned in all three
places it lives so a future flip cannot pass silently.

> **Updated twice after @davidwalter2's review.** `--stallRelTol` is split out to
> #157. Two of his comments changed the substance here rather than the prose: the
> reported conditioning is now the **true** condition number (§2), and a block is
> **dropped when whitening would make it worse** (§3) — which retracts a claim I
> had made twice.

### 1. Log which parameters are in each block

Blocks were logged by size only ("block of 14 parameters"), so there was no way
to tell from a log which directions a transform had helped — the first question
anyone asks when preconditioning changes a fit's behaviour.

Now on both paths. It was initially wired only into the spectral logging, which
left the default-transform user looking at the very line this PR opens by
complaining about.

### 2. Report the conditioning actually achieved

The summary printed a hardcoded `-> 1`: the list it aggregates held
`cond_before` only, and `Block.cond_after` was computed and never used.

It now reports the **true** condition number at both ends of the arrow, with the
scale-free degeneracy as a separate clause for the incoming block only:

```
ridge     condition number 3.3e+09 -> 1.44e+08, of which degeneracy 1.23
spectral  condition number 3.3e+09 -> 1, of which degeneracy 1.23
```

The correlation number that was there before is scale-invariant by construction,
so it could not see a scale collapse — and the ridge's failure mode *is* a scale
collapse. It printed `1.225 -> 1` for **both** transforms, i.e. it could not
distinguish them at all on the case that motivates having both. On the spectral
path it is worse than blind: `B` and `|B|` share eigenvectors, so with
`|B| = LLᵀ` the congruence sends `Λ` to its own signature, making `tb` symmetric
*and* orthogonal — `tb² = I`, true κ ≡ 1 at any block size. But `L` is a Cholesky
factor, not the eigenbasis, so `diag(tb)` picks up the arbitrary orientation
between the two and the `1/√|diag|` normalisation turns it into an artefact that
*grows* with block size: 1.7 / 7.7 / 21 / 34 at m = 5 / 12 / 30 / 60, worst
exactly where `--preconditionBlocks none` sends you, against a true value of 1
throughout.

That identity is now the spec test, at four block sizes — a better statement of
what the transform does than "reaches the identity up to sign" on one hand-built
3×3. It holds only where nothing was floored (a floored direction is one the
congruence no longer preserves), so κ is still measured per block rather than
asserted, and a second test pins the floored case.

### 3. `--preconditionTransform {ridge,spectral}`, and a ridge fix

Two separate problems, which the first version of this description conflated:

**A scalar ridge cannot serve a heterogeneous block** — it must be at least
`|λ_min|` to restore definiteness, so it swamps every softer direction and the
whitening leaves them near-null. Measured on curvatures `(3.3e9, -8.7e6, 1)`:
the O(1) direction landed at `7e-08` and the block went `3.3e+09 -> 1.44e+08`,
where spectral whitening of `|H| = Q|Λ|Qᵀ` gives `1`. This is a genuine
limitation of one scalar number and it is what spectral is for.

**All-negative blocks were a ridge bug**, not an argument for spectral. `scale`
is only the unit the ridge is expressed in; nothing in the schedule requires it
to come from a positive curvature, so taking it from `max(diag)` rather than
`max|diag|` rejected those blocks before any factorisation was attempted. That
change is provably inert for every block `main` already handled — `chol`
identical to 6.7e-16 relative over 4000 blocks, and necessarily so, since `scale`
cancels between `need/scale` and `eps*scale` from try 2 onward.

**But ridging such a block makes it worse, so it is now dropped.** I had claimed
"a usable transform beats none" on the strength of `cond_after = 320` where the
block used to be skipped. That was wrong: doing nothing leaves it at **30**, so
the transform was 10.7× worse than the skip it replaced. Over 7094 all-negative
blocks at four off-diagonal strengths, ridging degraded the true condition number
in **100%** of them, median 10.8×, and improved **none**.

So the "a preconditioner must never break a fit" principle — previously applied
only to factorisation *failing* — now also covers it *succeeding and making
things worse*: a block whose true κ degrades by more than `DEGRADE_TOLERANCE` is
left unpreconditioned, which restores exactly the `--precondition off` behaviour
for that block. The guard covers the pre-existing ridge path too, not just the
blocks `max|diag|` newly reaches. It cannot key off the correlation number, which
is blind to the clean case: a diagonal block has the identity as its correlation
matrix at both ends, so `diag(-7.5e3, -6.3e4, -2.1e3)` reports `1 -> 1` where the
truth is `30 -> 320`.

It deliberately stands down above `SINGULAR_COND`. A rank-deficient block reads
`5.28e+16 -> 7.32e+16` — a 1.39× "degradation" between two values past `1/eps`,
where the true condition number has stopped being a measurement — while the
*correlation* number shows the ridge genuinely helping it, `3.1e+16 -> 1.4e+12`.
Ridging a rank-deficient block into shape is what the ridge was for, so that is
left alone. It is the mirror image of the all-negative case, and the reason both
metrics are kept rather than one being declared correct.

**Net effect on all-negative blocks, stated plainly since this description
previously oversold it:** they end up unpreconditioned, as before. What changed
is that it is now reached by measurement with a logged reason rather than by a
scale test that was mislabelled (it always said `max|diag|`), and that the same
protection now covers every other block. Spectral is what actually rescues them —
it reaches 1 on the same block.

Spectral agrees **exactly** with the plain Cholesky on a positive-definite block
(both return the identity, verified to 1e-8 on a 40×40), so nothing is lost on
the easy blocks. It costs an eigendecomposition rather than a Cholesky: measured
11× at m=200 and 44× at m=2112, i.e. ~1.7 s once per preconditioner build.
Taking `|Λ|` preserves each direction's sign, which matters because
`trust-krylov` exploits negative curvature to escape saddles.

The cross-option guidance now lives in the `preconditioner.py` module docstring
(`WHITENING EACH BLOCK`, `CHOOSING THE OPTIONS`, `WHAT THE NUMBERS MEAN`,
`NEVER MAKE IT WORSE`) rather than in `--help`, with a pointer from each option —
it is guidance rather than option semantics, and the help strings had grown to
591 characters on a single source line against a file maximum of 186. The short
version is that the advice **inverts**: under ridge, blocking is a correctness
tool; under spectral a larger block is always better or equal, and
`--preconditionBlockThreshold` works against you.

Also times the reference-matrix evaluation, since that is the whole cost of a
rebuild. Measured across three rebuilds on one fit: 130, 424, 139 s — variable
enough with node load that it should be read per-run rather than assumed.

### Tests

97 passing (86 before). `tests/test_preconditioner_spectral.py` pins the ridge
failure mode, the equivalence with Cholesky when positive definite, the
`true κ ≡ 1` spec at four block sizes, the near-null floor, the guard declining
an all-negative block *while `_factorise_ridge` alone succeeds* (so the two
rejection reasons cannot be confused), a block each transform genuinely helps
being kept, the already-singular carve-out, and the ridge default.

Three additions came out of review:

- `test_fit_is_invariant_for_either_transform` in `test_preconditioner.py`. For a
  reparameterisation the decisive property is that the minimum does not move, and
  nothing here ran a fit.
- the near-null floor test asserted on `blk.chol`, whose entries are `√|λ|` and so
  get *smaller* for a near-null direction — it passed with the flooring removed.
  It now asserts on `L⁻¹` and is mutation-checked. The protection is a factor of
  ~21, not "1e9 → bounded", since the floor is `eps·m·wmax`; the comment used to
  promise more than the code delivers.
- the all-negative test now pins the guard rather than a rescue that was not one.

### DIFF
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -105,9 +105,7 @@ class Fitter:
         self.precondition_params = getattr(options, "preconditionParams", None)
         self.precondition_from = getattr(options, "preconditionFrom", "hessian")
         self.precondition_blocks = getattr(options, "preconditionBlocks", "auto")
-        self.precondition_transform = getattr(
-            options, "preconditionTransform", "ridge"
-        )
+        self.precondition_transform = getattr(options, "preconditionTransform", "ridge")
         self.precondition_block_threshold = getattr(
             options, "preconditionBlockThreshold", 0.1
         )
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -2316,7 +2316,16 @@ class Fitter:
         # formed (memory, or a tracing failure on a large model) fall back to an
         # unpreconditioned fit rather than taking the whole job down.
         try:
+            _t_ref = time.time()
             hess_np = self._reference_matrix()
+            # Time it: this is a FULL Hessian at the current point, and it is
+            # the entire cost of a preconditioner rebuild. Whether restarting
+            # aggressively is worth it is exactly this number against the
+            # remaining iterations, so it should not have to be guessed.
+            logger.debug(
+                f"Preconditioner reference matrix ({self.precondition_from}) "
+                f"took {time.time() - _t_ref:.1f} s"
+            )
         except Exception as ex:
             logger.warning(
                 f"Could not compute the reference Hessian for preconditioning ({ex}); "
@@ -2351,8 +2360,10 @@ class Fitter:
             transform=self.precondition_transform,
             # names so the per-block log says WHICH parameters each block holds;
             # "block of 14 parameters" alone leaves no way to tell from a log
-            # which directions the transform actually helped.
-            names=[str(p) for p in self.parms],
+            # which directions the transform actually helped. astype(str), not
+            # str() per element: indata.systs comes out of h5py as an object
+            # array of bytes, so str() would render every name as b'...'.
+            names=self.parms.astype(str),
         )
 
     def fit(self):
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -2347,6 +2347,10 @@ class Fitter:
             theta_ref,
             index_blocks,
             ridge=self.precondition_ridge,
+            # names so the per-block log says WHICH parameters each block holds;
+            # "block of 14 parameters" alone leaves no way to tell from a log
+            # which directions the transform actually helped.
+            names=[str(p) for p in self.parms],
         )
 
     def fit(self):
--- a/rabbit/fitter.py
+++ b/rabbit/fitter.py
@@ -105,6 +105,9 @@ class Fitter:
         self.precondition_params = getattr(options, "preconditionParams", None)
         self.precondition_from = getattr(options, "preconditionFrom", "hessian")
         self.precondition_blocks = getattr(options, "preconditionBlocks", "auto")
+        self.precondition_transform = getattr(
+            options, "preconditionTransform", "ridge"
+        )
         self.precondition_block_threshold = getattr(
             options, "preconditionBlockThreshold", 0.1
         )
@@ -2347,6 +2350,7 @@ class Fitter:
             theta_ref,
             index_blocks,
             ridge=self.precondition_ridge,
+            transform=self.precondition_transform,
             # names so the per-block log says WHICH parameters each block holds;
             # "block of 14 parameters" alone leaves no way to tell from a log
             # which directions the transform actually helped.
--- a/rabbit/parsing.py
+++ b/rabbit/parsing.py
@@ -257,7 +257,7 @@ def common_parser():
         "'expressions' makes one block per --preconditionParams entry. 'none' does "
         "no grouping at all and factorises the whole selected scope as a single "
         "block, which keeps every cross-correlation but is the most expensive and "
-        "fails entirely if any part of the scope is singular.",
+        "fails entirely if any part of the scope is singular. NOTE the purpose of blocking depends on the transform. Under 'ridge' it is a correctness tool: one scalar ridge cannot serve a block whose curvatures span orders of magnitude, so splitting protects the soft directions. Under 'spectral' a larger block is always better or equal -- it whitens the cross-terms exactly where splitting discards them (measured: one block of 8 gives condition number 1, the same matrix in two blocks gives 4.0) -- so blocking there only limits the O(m^3) cost, and 'none' is usually the right choice.",
     )
     parser.add_argument(
         "--preconditionBlockThreshold",
@@ -266,7 +266,7 @@ def common_parser():
         help="Correlation threshold for --preconditionBlocks auto. Correlations "
         "below it are left unpreconditioned. Too low and every parameter "
         "percolates into a single block; too high and genuinely coupled "
-        "parameters are split apart.",
+        "parameters are split apart. Only used by --preconditionBlocks auto. Under 'spectral' this threshold works against you: it splits away exactly the cross-correlations the spectral transform would have handled exactly, so 'every parameter percolates into a single block' is the IDEAL there rather than the warning it is under 'ridge'.",
     )
     parser.add_argument(
         "--preconditionFrom",
@@ -281,7 +281,7 @@ def common_parser():
         "exact Hessian is indefinite at the starting point, whitening with it leaves "
         "those directions negative and the fit stalls, so 'hessian' is usually the "
         "better choice; prefer 'gaussnewton' only where the Hessian is positive "
-        "definite anyway, e.g. near a minimum.",
+        "definite anyway, e.g. near a minimum. Under --preconditionTransform spectral there is no reason to choose gaussnewton: its PSD-ness exists to guarantee the Cholesky succeeds, which spectral does not need, and it still cannot represent negative curvature. Prefer 'hessian' there.",
     )
     parser.add_argument(
         "--preconditionRidge",
@@ -290,7 +290,28 @@ def common_parser():
         help="Ridge added to the preconditioning block diagonal, relative to its largest "
         "diagonal entry, to keep near-degenerate blocks factorisable. Escalated "
         "automatically if the Cholesky still fails; a block that cannot be factorised "
-        "falls back to no preconditioning.",
+        "falls back to no preconditioning. ONLY applies to --preconditionTransform ridge; the spectral transform derives its floor from the numerical rank instead.",
+    )
+    parser.add_argument(
+        "--preconditionTransform",
+        default="ridge",
+        type=str,
+        choices=["ridge", "spectral"],
+        help="How to whiten each block. 'ridge' (the default) factorises "
+        "H + eps*I, raising eps until the Cholesky succeeds. 'spectral' "
+        "factorises |H| = Q |Lambda| Q^T instead, giving every eigendirection "
+        "its own scale. They agree exactly on a positive definite block -- both "
+        "return the identity -- and differ where H is indefinite, which on real "
+        "data it generally is away from the minimum. A single scalar ridge must "
+        "be at least |lam_min| to restore definiteness, so in a block whose "
+        "curvatures span orders of magnitude it swamps every softer direction "
+        "and the whitening leaves them near-null: measured on curvatures "
+        "(3.3e9, -8.7e6, 1) the O(1) direction landed at 7e-08 and the block "
+        "went from condition number 3.3e+09 to 1.44e+08, where spectral gives "
+        "1. Spectral also handles a block whose whole diagonal is negative, "
+        "which the ridge cannot start on at all. It costs an eigendecomposition "
+        "rather than a Cholesky: measured 11x at m=200 and 44x at m=2112, i.e. "
+        "~1.7 s once per preconditioner build.",
     )
     parser.add_argument(
         "--snapshotFile",
--- a/rabbit/parsing.py
+++ b/rabbit/parsing.py
@@ -257,7 +257,10 @@ def common_parser():
         "'expressions' makes one block per --preconditionParams entry. 'none' does "
         "no grouping at all and factorises the whole selected scope as a single "
         "block, which keeps every cross-correlation but is the most expensive and "
-        "fails entirely if any part of the scope is singular. NOTE the purpose of blocking depends on the transform. Under 'ridge' it is a correctness tool: one scalar ridge cannot serve a block whose curvatures span orders of magnitude, so splitting protects the soft directions. Under 'spectral' a larger block is always better or equal -- it whitens the cross-terms exactly where splitting discards them (measured: one block of 8 gives condition number 1, the same matrix in two blocks gives 4.0) -- so blocking there only limits the O(m^3) cost, and 'none' is usually the right choice.",
+        "fails entirely if any part of the scope is singular. What blocking is "
+        "FOR depends on --preconditionTransform (under 'ridge' it is a "
+        "correctness tool, under 'spectral' only a cost limit): see CHOOSING "
+        "THE OPTIONS in rabbit/preconditioner.py.",
     )
     parser.add_argument(
         "--preconditionBlockThreshold",
@@ -266,7 +269,9 @@ def common_parser():
         help="Correlation threshold for --preconditionBlocks auto. Correlations "
         "below it are left unpreconditioned. Too low and every parameter "
         "percolates into a single block; too high and genuinely coupled "
-        "parameters are split apart. Only used by --preconditionBlocks auto. Under 'spectral' this threshold works against you: it splits away exactly the cross-correlations the spectral transform would have handled exactly, so 'every parameter percolates into a single block' is the IDEAL there rather than the warning it is under 'ridge'.",
+        "parameters are split apart. Only used by --preconditionBlocks auto, "
+        "and under --preconditionTransform spectral it works against you: see "
+        "CHOOSING THE OPTIONS in rabbit/preconditioner.py.",
     )
     parser.add_argument(
         "--preconditionFrom",
@@ -281,7 +286,9 @@ def common_parser():
         "exact Hessian is indefinite at the starting point, whitening with it leaves "
         "those directions negative and the fit stalls, so 'hessian' is usually the "
         "better choice; prefer 'gaussnewton' only where the Hessian is positive "
-        "definite anyway, e.g. near a minimum. Under --preconditionTransform spectral there is no reason to choose gaussnewton: its PSD-ness exists to guarantee the Cholesky succeeds, which spectral does not need, and it still cannot represent negative curvature. Prefer 'hessian' there.",
+        "definite anyway, e.g. near a minimum. Under --preconditionTransform "
+        "spectral there is no reason to choose 'gaussnewton' at all: see "
+        "CHOOSING THE OPTIONS in rabbit/preconditioner.py.",
     )
     parser.add_argument(
         "--preconditionRidge",
@@ -290,7 +297,9 @@ def common_parser():
         help="Ridge added to the preconditioning block diagonal, relative to its largest "
         "diagonal entry, to keep near-degenerate blocks factorisable. Escalated "
         "automatically if the Cholesky still fails; a block that cannot be factorised "
-        "falls back to no preconditioning. ONLY applies to --preconditionTransform ridge; the spectral transform derives its floor from the numerical rank instead.",
+        "falls back to no preconditioning. Applies to "
+        "--preconditionTransform ridge only; spectral derives its floor from "
+        "the numerical rank of the block instead.",
     )
     parser.add_argument(
         "--preconditionTransform",
@@ -301,17 +310,13 @@ def common_parser():
         "H + eps*I, raising eps until the Cholesky succeeds. 'spectral' "
         "factorises |H| = Q |Lambda| Q^T instead, giving every eigendirection "
         "its own scale. They agree exactly on a positive definite block -- both "
-        "return the identity -- and differ where H is indefinite, which on real "
-        "data it generally is away from the minimum. A single scalar ridge must "
-        "be at least |lam_min| to restore definiteness, so in a block whose "
-        "curvatures span orders of magnitude it swamps every softer direction "
-        "and the whitening leaves them near-null: measured on curvatures "
-        "(3.3e9, -8.7e6, 1) the O(1) direction landed at 7e-08 and the block "
-        "went from condition number 3.3e+09 to 1.44e+08, where spectral gives "
-        "1. Spectral also handles a block whose whole diagonal is negative, "
-        "which the ridge cannot start on at all. It costs an eigendecomposition "
-        "rather than a Cholesky: measured 11x at m=200 and 44x at m=2112, i.e. "
-        "~1.7 s once per preconditioner build.",
+        "return the identity -- and differ where H is indefinite, which away "
+        "from the minimum it generally is: one scalar ridge swamps every "
+        "direction softer than |lam_min| and leaves them near-null, where "
+        "spectral does not. Costs an eigendecomposition rather than a Cholesky, "
+        "~1.7 s once per build at m=2112. For the measurements, and for how "
+        "this interacts with the other --precondition* options, see WHITENING "
+        "EACH BLOCK in rabbit/preconditioner.py.",
     )
     parser.add_argument(
         "--snapshotFile",
--- a/rabbit/preconditioner.py
+++ b/rabbit/preconditioner.py
@@ -156,7 +156,9 @@ class Preconditioner:
         return max(c) if c else None
 
     @classmethod
-    def from_hessian(cls, hess, theta_ref, index_blocks, ridge=1e-8, max_tries=4):
+    def from_hessian(
+        cls, hess, theta_ref, index_blocks, ridge=1e-8, max_tries=4, names=None
+    ):
         """Build from a reference Hessian, one factorisation per index block.
 
         ``index_blocks`` is a list of index arrays (a single array is accepted
@@ -173,7 +175,12 @@ class Preconditioner:
         for spec in index_blocks:
             label, idx = ("", spec) if not isinstance(spec, tuple) else spec
             blk = cls._factorise(
-                hess, np.asarray(idx, dtype=np.int64), ridge, max_tries, label
+                hess,
+                np.asarray(idx, dtype=np.int64),
+                ridge,
+                max_tries,
+                label,
+                names=names,
             )
             if blk is not None:
                 blocks.append(blk)
@@ -189,19 +196,35 @@ class Preconditioner:
         n_req = len(index_blocks)
         npar = sum(b.idx.size for b in blocks)
         conds = [b.cond_before for b in blocks if b.cond_before is not None]
+        # Report the conditioning ACHIEVED, not a hardcoded 1. The literal "-> 1"
+        # this used to print claimed the transform had whitened every block
+        # perfectly, which is only true of the ridged matrix; measured on the
+        # un-ridged block (Block.cond_after) it can be worse than where it
+        # started -- seen at 1.8e+03 -> 1e+04 on a block whose ridge was forced
+        # up to |lam_min| by a large negative eigenvalue. A summary that cannot
+        # express that is worse than none, because it is read as success.
+        conds_after = [b.cond_after for b in blocks if b.cond_after is not None]
         summary = f"Preconditioned {len(blocks)} of {n_req} block(s), {npar} parameters"
         if conds:
             summary += (
                 f"; correlation condition number median {np.median(conds):.3g}, "
-                f"worst {max(conds):.3g} -> 1 at the reference point"
+                f"worst {max(conds):.3g}"
             )
+            if conds_after:
+                summary += (
+                    f" -> median {np.median(conds_after):.3g}, "
+                    f"worst {max(conds_after):.3g}"
+                )
+                if max(conds_after) > max(conds):
+                    summary += " (WORSE than unpreconditioned)"
+            summary += " at the reference point"
         if len(blocks) < n_req:
             summary += f" ({n_req - len(blocks)} block(s) not factorisable, skipped)"
         logger.info(summary)
         return cls(theta_ref, blocks)
 
     @staticmethod
-    def _factorise(hess, idx, ridge, max_tries, label=""):
+    def _factorise(hess, idx, ridge, max_tries, label="", names=None):
         """One block -> a :class:`Block`, or None if it is unusable."""
         tag = f"{label} " if label else ""
         if idx.size == 0:
@@ -217,7 +240,7 @@ class Preconditioner:
         if not np.isfinite(scale) or scale <= 0.0:
             logger.warning(
                 f"Preconditioning block {tag}has no positive diagonal "
-                f"(max diag = {scale}); skipping."
+                f"(max diag = {scale}); skipping.  [{_describe(idx, names)}]"
             )
             return None
 
@@ -269,11 +292,16 @@ class Preconditioner:
             tb = scipy.linalg.solve_triangular(chol, block, lower=True, trans="N")
             tb = scipy.linalg.solve_triangular(chol, tb.T, lower=True, trans="N").T
             cond_after = _cond_corr(tb)
+            # Log the MEMBERS, not just the size: "block of 14 parameters" does
+            # not say which 14, so there is no way to tell from a log which
+            # parameters the transform actually helped. Names come from the
+            # optional `names` argument; without it this degrades to indices.
+            who = _describe(idx, names)
             logger.debug(
                 f"Preconditioning {tag}block of {idx.size} parameters from the "
                 f"reference Hessian (ridge {eps:.3g} x max|diag|): correlation "
                 f"condition number {cond_before:.3g} -> {cond_after:.3g} at the "
-                "reference point"
+                f"reference point  [{who}]"
             )
             return Block(idx, chol, cond_before, cond_after, label)
 
@@ -370,6 +398,21 @@ class Preconditioner:
             f"preconditioning: {self.n_blocks} block(s) covering "
             f"{self.nblock} of {self.n} parameters"
         )
+
+
+def _describe(idx, names=None, limit=8):
+    """Readable membership for a block: names when available, else indices."""
+    idx = np.asarray(idx).ravel()
+    if names is None:
+        out = [str(int(i)) for i in idx[:limit]]
+    else:
+        out = [
+            str(names[int(i)]) if int(i) < len(names) else str(int(i))
+            for i in idx[:limit]
+        ]
+    if idx.size > limit:
+        out.append(f"... +{idx.size - limit} more")
+    return ", ".join(out)
 
 
 def _cond_corr(mat):
--- a/rabbit/preconditioner.py
+++ b/rabbit/preconditioner.py
@@ -11,7 +11,9 @@ cost minutes and return a bit-identical loss.
 
 WHAT. We reparameterise. With theta the physical parameters, y the internal
 ones, theta_ref the point the transform was built at, and a reference Hessian
-H0 = L L^T restricted to a selected block:
+H0 = L L^T restricted to a selected block (away from a minimum H0 is
+generally indefinite and has no Cholesky -- see WHITENING EACH BLOCK below for
+the two ways to get an L):
 
     theta = theta_ref + T y,     T = L^-T   ->   T^T H0 T = I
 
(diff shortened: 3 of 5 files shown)


tests/test_preconditioner_spectral.py
```diff
@@ -0,0 +1,115 @@
...
+    assert np.all(np.diag(t) < 0)  # still a maximum in every direction
+
+
+def test_near_null_direction_is_floored_not_amplified():
+    """A genuinely flat direction has no scale to whiten to."""
+    m = np.diag([1.0, 1e-18]).astype(float)
+    blk = _factorise(m, ridge=1e-8)
+    assert blk is not None
+    # the floor keeps L^-1 bounded; without it 1/sqrt(1e-18) = 1e9
+    assert np.max(np.abs(blk.chol)) < 1e3
```
davidwalter2: This test is vacuous — it passes with the flooring removed.

`blk.chol` is `L`, not `L⁻¹`. Its entries are `√|λ|`, so a near-null direction makes them *small*, not large. Ran it both ways on this exact block:

```
no-floor  chol max: 1.0        -> assert < 1e3 passes?  True
no-floor  L^-1 max: 1.0e+09
floored   chol max: 1.0
floored   L^-1 max: 4.7e+07
```

The comment on the line above names the right quantity ("the floor keeps L^-1 bounded; without it 1/sqrt(1e-18) = 1e9") — the assert just doesn't check it. Asserting on `np.linalg.inv(blk.chol)` (or on the floored eigenvalue) would pin the mechanism.

Worth noting while you're here that the real protection is a factor of ~21 (1e9 → 4.7e7), not "1e9 → bounded" — the floor is `eps·m·wmax`, so `1/√floor` is still ~5e7. That is fine, but the comment currently promises more than the code delivers.

tests/test_preconditioner_spectral.py
```diff
@@ -0,0 +1,115 @@
+"""The spectral transform, and the ridge failure it exists to fix.
+
+These pin the mechanism with concrete numbers so a future change back to a
+scalar ridge cannot pass silently.
```
davidwalter2, line 4: All five tests here are on `_factorise` in isolation — none of them runs a fit. For a reparameterisation the decisive property is that the minimum doesn't move, and `test_preconditioner.py` already has exactly that pattern in `test_fit_is_invariant_for_either_reference_source`.

I wrote the counterpart and it passes for both transforms (`2 passed in 22.01s`):

```python
@pytest.mark.parametrize("transform", ["ridge", "spectral"])
def test_fit_is_invariant_for_either_transform(transform):
    with tempfile.TemporaryDirectory() as tmp:
        filename = make_polynomial_tensor(tmp, order=6)
        plain, _ = _run(filename, "trust-krylov")
        pre, pc = _run(
            filename, "trust-krylov", precondition=True,
            preconditionTransform=transform,
        )
        assert pc.enabled and pc.nblock > 1
        assert check_results("plain", plain, f"precond[{transform}]", pre)
```

Six lines reusing the existing `_run`/`check_results`, and it's the test that justifies the whole feature. Could go in `test_preconditioner.py` next to the reference-source version rather than here.

rabbit/preconditioner.py
```diff
@@ -156,14 +156,24 @@ def cond_after(self):
...
+        transform="ridge",
+    ):
         """Build from a reference Hessian, one factorisation per index block.
 
         ``index_blocks`` is a list of index arrays (a single array is accepted
-        and treated as one block). Each block is symmetrised, then a ridge
-        proportional to its largest diagonal entry is added until the Cholesky
-        succeeds. A block that cannot be factorised is dropped -- the others
-        still apply, and a preconditioner must never break a fit.
+        and treated as one block). Each block is symmetrised and then whitened
```
davidwalter2, line 324: This now documents the non-default path as the behaviour. `transform` defaults to `"ridge"`, so for most callers each block is *not* "whitened by the Cholesky of `|H| = Q |Lambda| Q^T`" — and the ridge description that was here has been deleted rather than kept alongside.

Three docs to reconcile:

- this docstring, which should describe the dispatch and both paths;
- `_factorise_spectral`'s docstring, which kept the old generic one-liner ("One block -> a `Block`, or None if it is unusable") instead of describing spectral — the substance is all in the inline comment below it;
- the module docstring, which is the actual documentation for this file and isn't touched at all. It still says `H0 = L L^T` with no mention that there are now two ways to get `L`.

The module docstring also seems like the natural home for the cross-option guidance that currently sits in the `--help` strings.

rabbit/preconditioner.py
```diff
@@ -217,7 +270,7 @@ def _factorise(hess, idx, ridge, max_tries, label=""):
         if not np.isfinite(scale) or scale <= 0.0:
             logger.warning(
                 f"Preconditioning block {tag}has no positive diagonal "
-                f"(max diag = {scale}); skipping."
+                f"(max diag = {scale}); skipping.  [{_describe(idx, names)}]"
```
davidwalter2: This guard is a ridge-path bug that the PR routes around rather than fixes.

`scale` is only the unit the ridge is expressed in. Nothing about the ridge schedule requires it to be positive — `np.max(np.abs(diag))` would let an all-negative block through and the existing spectrum-based escalation (`itry == 1`, `need = |lam_min| + margin`) would then size the ridge correctly.

So if all-negative blocks really cost 21 of 34 on a real fit, the **default** path still loses them after this PR. That's independent of the spectral transform and looks like a one-line change; worth deciding whether it belongs here, because right now the description sells it as a reason to prefer spectral when it's really a reason to fix ridge.

(It wouldn't fix the heterogeneous-curvature failure, which is a genuine limitation of one scalar ridge and does need spectral. Those two arguments are currently conflated in the description.)

rabbit/preconditioner.py
```diff
@@ -372,6 +536,21 @@ def summary(self):
         )
 
 
+def _describe(idx, names=None, limit=8):
```
davidwalter2, line 805: `_describe` is wired into all five log lines on the spectral path but into exactly one warning on the ridge path (the `no positive diagonal` one above). The ridge **success** log is untouched:

```
"Preconditioning {tag}block of {idx.size} parameters from the reference Hessian "
"(ridge {eps:.3g} x max|diag|): correlation condition number ..."
```

That is literally the "block of 14 parameters" line the PR opens by complaining about, and it's the line every default-transform user sees. Under `--preconditionTransform ridge` you still can't tell from a log which parameters were in a block — so change #1 doesn't yet deliver its stated motivation.

Adding `[{_describe(idx, names)}]` to that debug line and to the `is not factorisable` warning would finish it. The other ridge-path warnings (`is empty`) would be nice too for symmetry with spectral.

rabbit/fitter.py
```diff
@@ -2347,6 +2358,11 @@ def _build_preconditioner(self):
             theta_ref,
             index_blocks,
             ridge=self.precondition_ridge,
+            transform=self.precondition_transform,
+            # names so the per-block log says WHICH parameters each block holds;
+            # "block of 14 parameters" alone leaves no way to tell from a log
+            # which directions the transform actually helped.
+            names=[str(p) for p in self.parms],
```
davidwalter2: `str(p)` will render every name with a `b'...'` wrapper. `self.parms` is `np.concatenate([param_model.params, indata.systs])` and `indata.systs` comes straight out of h5py as an object array of bytes. Checked on `tests/test_tensor.hdf5`:

```
dtype: object   repr(s[0]): b'slope_background'
str(s[0])      -> b'slope_background'
s.astype(str)[0] -> slope_background
```

So the membership log becomes `[b'slope_background', b'...', ...]`. `fitter.py` already uses `self.parms.astype(str)` in about ten places for exactly this reason — `names=self.parms.astype(str)` (or `list(...)`) would match the surrounding idiom and `_describe` handles an ndarray fine since it indexes and `str()`s each element anyway.

rabbit/callbacks.py
```diff
@@ -38,6 +38,16 @@ def __init__(self, xv, early_stopping=-1, snapshotter=None):
         self.t0 = time.time()
 
         self.early_stopping = early_stopping
+        # Relative improvement over the window below which the fit counts as
+        # stalled. 0.0 keeps the original test, which fires only on LITERALLY no
+        # improvement -- and a fit that crawls never satisfies that. Measured on
```
davidwalter2: This comment and the PR description say opposite things, and the comment is the one that survives into the codebase.

Here: a 3720-parameter fit "improved by ~1e-4 relative per iteration for hundreds of iterations at 100+ s each, so the stall detector never fired, no restart happened".

Description: "our own fits do **not** exhibit that crawl — replayed against real trajectories they keep gaining 0.2–4% per 30 iterations, so no threshold up to 1e-3 fires on them, and the exact test was correctly reporting 'not stalled'. … it was not the fix for what we were chasing."

If the second is what actually held up, this comment should be rewritten to argue the generalisation on its own terms rather than citing a measurement that was withdrawn.

Two more things on this change specifically, which is why I'd rather see it as its own PR:

**No test.** Not even that `stall_rel_tol=0.0` reproduces the old test — which is trivially checkable and does hold, since `gained <= 0` ⟺ `ref - loss <= 0` ⟺ `ref <= loss`. And nothing exercises `stall_rel_tol > 0` detecting a crawl. This is the only change in the PR that touches a default-on path (`--earlyStopping` defaults to 20), so it's the one that most wants a regression test pinning the equivalence.

**The criterion conflates converged with stalled.** "Flat to 1e-4 over 20 iterations" is also what approaching a minimum looks like, whereas the exact test only fires when the trust radius has actually collapsed. So a non-zero `--stallRelTol` will tend to fire near a good minimum and trigger restarts costing a full reference Hessian each — 130/424/139 s by your own measurement above. `RESTART_MIN_IMPROVEMENT = 1e-9` bounds the loop and scipy's `gtol` usually exits first, so it's not unsafe, but it's a real behaviour change that the help text ("Something like 1e-4") doesn't warn about.

rabbit/parsing.py
```diff
@@ -257,7 +271,7 @@ def common_parser():
         "'expressions' makes one block per --preconditionParams entry. 'none' does "
         "no grouping at all and factorises the whole selected scope as a single "
         "block, which keeps every cross-correlation but is the most expensive and "
-        "fails entirely if any part of the scope is singular.",
+        "fails entirely if any part of the scope is singular. NOTE the purpose of blocking depends on the transform. Under 'ridge' it is a correctness tool: one scalar ridge cannot serve a block whose curvatures span orders of magnitude, so splitting protects the soft directions. Under 'spectral' a larger block is always better or equal -- it whitens the cross-terms exactly where splitting discards them (measured: one block of 8 gives condition number 1, the same matrix in two blocks gives 4.0) -- so blocking there only limits the O(m^3) cost, and 'none' is usually the right choice.",
```
davidwalter2: Formatting: the four edited help strings are single source lines of 591 / 342 / 289 / 165 characters. The longest line anywhere in `parsing.py` on main is 186. black doesn't reformat string literals so this lints clean, but it breaks the file's own convention of wrapping help text at ~88 — every other `help=` here is a stack of short adjacent literals.

Separately, this is a lot of prose to put in `--help`. The three "how this option interacts with the transform" paragraphs (`--preconditionBlocks`, `--preconditionBlockThreshold`, `--preconditionFrom`) are genuinely useful, but they're guidance rather than option semantics — the `preconditioner.py` module docstring already carries that kind of explanation (FINDING THE BLOCKS, SEVERAL BLOCKS) and would be a better home, with a one-line pointer from each help string.

davidwalter2 (commented):
Reviewed by fetching the branch, running the suite, and checking the claims numerically.

The core numerics of the spectral transform are right: whitening by `|H| = Q|Λ|Qᵀ` is a valid change of variables for any invertible `T`, and keeping the sign so trust-krylov can still exploit negative curvature is the correct call. Shipping it opt-in with the ridge default is the right decision. `86 passed` on the branch (I get 86 rather than 60 with the tests dir on the path — collection difference, not a problem), black clean, the isort failures are pre-existing on main.

Taking the four changes separately:

| # | Change | View |
|---|---|---|
| 2 | Summary reports `cond_after` instead of a hardcoded `-> 1` | Correct and uncontroversial — the list genuinely held `cond_before` only and `Block.cond_after` was genuinely computed and dropped |
| 3 | `--preconditionTransform spectral` | Sound and well argued, but one test is vacuous, the decisive test is missing, and the docstring documents the non-default path as the behaviour |
| 1 | Block-membership logging | Only wired into the spectral path — the default path still prints the "block of 14 parameters" line this PR opens by complaining about |
| 4 | `--stallRelTol` | I'd split this out. The PR description retracts its motivation but the code comment still asserts it, and it is the one change touching a default-on path with no test at all |

Comments inline. The blocking ones for me are the vacuous floor test and the missing end-to-end invariance test; the rest are fixes I'd be happy to see in a follow-up.

**Smaller things not worth their own comment:**

- `"ridge"` as the default now lives in three places (parser default, `Fitter.__init__` getattr, `_factorise` signature) and the test pins only the third, though the description says it pins "the ridge default".
- `[timing]` is a new log prefix with exactly one occurrence in `fitter.py`.
- `_factorise_spectral` duplicates ~15 lines of preamble from `_factorise_ridge` (empty check, symmetrise, finite check, `cond_before`) that could be hoisted into the dispatcher.
- In the summary, `conds` and `conds_after` are filtered for `None` independently, so in principle the two medians could be over different block sets. In practice both are set together, so this is cosmetic.
- The Cholesky of `|H|` is a second O(m³) that can itself fail on a floored spectrum (cond ≈ 1/(eps·m) ≈ 4e14), when `Q` and `√|Λ|` are already in hand. I assume the triangular-solve fallback in `_apply_T` is why `L` has to be triangular — worth a comment saying so, since it is otherwise an obvious "why not just use the symmetric square root".

tests/test_preconditioner_spectral.py
lucalavezzo: You're right, and it was worse than vacuous — I had asserted on the wrong side of the factorisation. `blk.chol` holds `√|λ|`, so a near-null direction makes its entries *smaller*; the assert could not have failed. Reproduced your table exactly (`chol max = 1.0` either way, `L⁻¹` 1e9 → 4.75e7).

Now asserts on `np.linalg.inv(blk.chol)` against `1/√(eps·m·wmax)`, and I mutation-checked it rather than just watching it pass — with the flooring commented out it fails with `999999999.9999999 == 47453132.8 ± 4.7e+06`.

And you're right that the comment promised more than the code delivers. The test now pins the factor as a factor:

```python
    # so the protection is a factor of ~21 here (1e9 -> 4.7e7), not unbounded
    # -> bounded.
    assert 10.0 < gain < 100.0
```

with the comment in `_factorise_spectral` rewritten to say "the protection is finite, not absolute … a factor of ~21. That is a step the trust region can still reject rather than one that overflows it."

tests/test_preconditioner_spectral.py
lucalavezzo: Agreed, and added — in `test_preconditioner.py` next to the reference-source version as you suggest, not in the spectral file. Substantively yours; I only reordered the kwargs for black. Passes for both transforms.

You're right that it is the test that justifies the feature: the five unit tests pin the *algebra* of the whitening, and none of them would have caught a transform that whitened correctly and still moved the minimum.

While there: `test_preconditioner_spectral.py` is deliberately kept free of the TensorFlow import chain (it is numpy/scipy only, ~1.4 s), so the parser and `Fitter` defaults you note below are pinned in `test_preconditioner.py` instead, which already needs the fitter stack.

rabbit/preconditioner.py
lucalavezzo: All three reconciled, and you're right that the module docstring was the real omission — it is the documentation for the file and I had left it describing a world with one transform.

- `from_hessian` now describes the dispatch and both paths.
- `_factorise_spectral` has a real docstring instead of the inherited one-liner.
- The module docstring gets a `WHITENING EACH BLOCK` section, and the `H0 = L L^T` line now points forward to it ("away from a minimum H0 is generally indefinite and has no Cholesky").
- The shared preamble is hoisted into `_factorise`, so its docstring documents the dispatch and the two leaf docstrings only have to cover what differs.

Also took your suggestion on the home for the cross-option guidance: it is now a `CHOOSING THE OPTIONS` section there, with a one-line pointer from each of the four help strings. See the reply on `parsing.py`.

rabbit/preconditioner.py
lucalavezzo: You're right, and this was the most useful comment in the review — it is a bug on the **default** path, and framing it as a reason to prefer spectral was letting the ridge off.

Fixed on the ridge: `scale = float(np.max(np.abs(diag)))`. Confirmed your reasoning end to end before changing it — try 1 fails, then the existing `itry == 1` spectrum branch sizes the ridge from `|λ_min|` and it succeeds:

```
diag(-7.5e3, -6.3e4, -2.1e3)
  before:  _factorise(..., "ridge") -> None          (skipped)
  after:   try 1 eps=1e-08 FAILED
           try 2 eps=1.1   OK, true cond_after = 320, diag signs all negative
```

Not as good as spectral, which reaches 1 on the same block, but a usable transform beats none — and the guard was also mislabelling itself, since the log line already called the quantity `max|diag|`. `_cond_corr` already takes `|diag|`, so nothing downstream needed changing. `test_ridge_can_start_on_an_all_negative_block` pins it, and asserts spectral still does better.

The description is rewritten to keep the two arguments apart: the all-negative skip was a ridge bug and is fixed on the ridge; the heterogeneous-curvature failure is the genuine limitation of one scalar number and remains the case for spectral.

rabbit/preconditioner.py
lucalavezzo: Right — change #1 was advertising something it only delivered on the path almost nobody is on. `_describe` is now on the ridge **success** debug line and on the `is not factorisable` warning, and the `is empty` / non-finite warnings are shared with the spectral path via the hoisted preamble, so they carry it for both transforms.

rabbit/fitter.py
lucalavezzo: Confirmed and fixed — `names=self.parms.astype(str)`. Reproduced on `tests/test_tensor.hdf5` exactly as you describe (`repr(s[0])` → `b'slope_background'`), so every name in the membership log would have carried the wrapper. Matching the surrounding idiom is the right call; there are ~10 `astype(str)` call sites in `fitter.py` for precisely this. `_describe` handles the ndarray unchanged.

rabbit/callbacks.py
lucalavezzo: Split out to #157, and you were right on all three counts.

**On the contradiction.** The comment was the wrong one to keep, and this repo's own record settles it in your favour rather than mine: the 3720-parameter run that stopped short of a minimum did so on **100 bit-identical iterations** with `--earlyStopping 100` firing — the exact test working as intended, not the crawl the comment described. The comment now argues the generalisation on its own terms and says explicitly "Do not read this option as a fix for a known stall."

**On the test.** Added, including the equivalence you point out is trivially checkable — over five trajectories (healthy descent, hard flat, crawl, worsening, all-zero), the last of which is the one input where the relative form would divide by zero and has to fall back:

```python
@pytest.mark.parametrize("losses", TRAJECTORIES)
def test_stall_rel_tol_zero_is_exactly_the_original_test(losses):
```

It re-derives the old predicate inline and asserts the two agree, rather than restating the algebra. Plus a crawl detected at 1e-3 and invisible at 0.0, a healthy descent surviving every threshold to 1e-2, and the default pinned in all three places it lives.

**On converged-vs-stalled.** That is the sharpest point and it is now in the code and the `--help`, not just in a PR thread — the comment says a non-zero value "will tend to fire near a good minimum and spend a full reference-Hessian evaluation (measured 130-424 s) on each restart", and `--help` carries a `WARNING:` and ends "Leave it at 0 unless a fit is demonstrably crawling" rather than suggesting 1e-4 unprompted.

rabbit/parsing.py
lucalavezzo: Both fixed, and the second is the better observation — it was guidance masquerading as option semantics.

**Formatting.** Wrapped. `parsing.py`'s longest line is back to 186, i.e. exactly `main`'s maximum, so this PR no longer adds the longest line in the file. (For the record: 591 / 342 / 289 / 165 before.)

**Home for the prose.** Moved to the `preconditioner.py` module docstring as you suggest, as a `CHOOSING THE OPTIONS` section next to the existing `FINDING THE BLOCKS` / `SEVERAL BLOCKS`, with a one-line pointer from each of the four options. Keeping the four paragraphs together also reads better than it did split across four `--help` strings, since the whole point is that the guidance *inverts* with the transform — that is much easier to see as a table of four options than one option at a time.

lucalavezzo:
Thanks — this was a genuinely useful review, and the two you called blocking were both real. I reproduced every numeric claim before changing anything and none of them was overstated; the `chol` vs `L⁻¹` one in particular I had simply asserted on the wrong side of the factorisation.

`--stallRelTol` is split out to #157. You were right that it did not belong here: it was the only change touching a default-on path, and the run I had cited as motivation actually stalled *literally* (100 bit-identical iterations, the exact test firing), so the comment was arguing from a measurement the description had already withdrawn.

The one I'd flag back is that a comment you filed as non-blocking was the most consequential: `scale = max(diag)` is a bug on the **default** ridge path, and describing the all-negative skip as a reason to prefer spectral was letting the ridge off for something one `np.abs` fixes. That is now a ridge fix, with the heterogeneous-curvature argument left standing on its own.

### Smaller things

- **`"ridge"` in three places, one pinned.** Fair — the description claimed more than the test did. All three now: `_factorise`'s signature and `from_hessian`'s in `test_preconditioner_spectral.py`, and the parser plus `Fitter`'s `getattr` in `test_preconditioner.py` (which has the fitter stack; the spectral file is deliberately TF-free). `parser.get_default(...)` rather than `parse_known_args([])`, since `common_parser` has a required positional. Same for `--stallRelTol` in #157.
- **`[timing]`.** Dropped; it now reads `Preconditioner reference matrix (hessian) took 130.4 s`.
- **Duplicated preamble.** Hoisted into `_factorise` — the empty check, symmetrise, finite check and `cond_before`. Two side effects worth noting: the ridge path picks up the non-finite guard it did not have, and `ridge`/`max_tries` are now dropped from `_factorise_spectral`'s signature, since the "shared so the dispatcher needs no special-casing" rationale disappears once the dispatcher does the work. Also pulled out `_whiten()` for the `L⁻¹ B L⁻ᵀ` that both paths and the tests were each spelling out.
- **`conds` / `conds_after` filtered independently.** Paired, with a comment saying why — you're right that it is cosmetic today, but nothing enforced it.
- **Cholesky of `|H|` vs the symmetric square root.** Your assumption is exactly right, and it deserved to be written down; it now is:

  ```
  # Cholesky of |B|, rather than the symmetric square root Q sqrt(|Lambda|)
  # which is already in hand: Block caches an explicit L^-1 but falls back
  # to a triangular solve when it cannot be formed (see _apply_T), so L
  # has to be lower triangular. The cost is a second O(m^3), and it can
  # itself fail on a floored spectrum -- cond(|B|) is then ~1/(eps*m),
  # i.e. ~4e14 -- which is why this is caught and the block skipped rather
  # than assumed to succeed.
  ```

### One thing I found while doing this, which limits change #2

`_cond_corr` is scale-invariant by construction, so the before/after pair still cannot see a *scale* failure — and the ridge's failure mode **is** a scale collapse. On the very block that motivates the spectral transform:

| | summary prints | true conditioning |
|---|---|---|
| ridge | `1.225 -> 1` | 3.3e9 → 1.4e8 |
| spectral | `1.225 -> 1` | 3.3e9 → 1 |

So it cannot distinguish the two transforms at all on the case that justifies having both. Scale invariance is right for `cond_before`, where parameter units are arbitrary, and wrong for `cond_after`, where the units are precisely what the transform chose and what governs the Krylov iteration count is `√κ` of the *true* Hessian in `y`. Your `1.8e+03 -> 1e+04` observation is the metric catching the *other* half — the degeneracy — which is why it is worth keeping rather than replacing.

I have left the code alone: reporting the true `κ` alongside changes the `Block` API and the log format, which is more than a review response should carry, and if the two numbers go in one `X -> Y` printout a reader will misread `1.225 -> 1.4e8` as preconditioning having made things worse. It is called out as a known limitation in §2 of the description. Happy to do it here if you'd rather it not ship half-blind.

90 passing (86 before) on this branch, 89 on #157, black and isort clean, no new flake8 findings. The `test_snapshot` SIGTERM case failed once for me under full-suite load and passes 3/3 alone — it signals after a hardcoded `time.sleep(2.0)`, so that looks like pre-existing load sensitivity rather than anything here.

davidwalter2:
Re-checked `7ae95f9` rather than taking the replies on trust; all of it holds.

- `90 passed`, black clean, `parsing.py`'s longest line back to 186 (= main's maximum).
- The floor test now genuinely bites. I mutation-checked it by disabling the flooring rather than watching it pass: `assert 999999999.9999999 == 47453132.81212577 ± 4.7e+06` — exactly your number.
- Ridge fix confirmed end to end: `diag(-7.5e3, -6.3e4, -2.1e3)` now factorises (true `cond_after` 320, all diagonal signs negative) where it used to return `None`, and the ridge **success** debug line carries `_describe` now, which was the half that was missing.
- `names=self.parms.astype(str)`, `test_fit_is_invariant_for_either_transform` in `test_preconditioner.py`, `_whiten` hoisted, `[timing]` gone.

And you're right that I filed the `scale = max(diag)` comment too low. It went in as a non-blocking one because it is a one-line change; what mattered was the framing, and that is the part you caught rather than me.

## On `_cond_corr`: do it here

I reproduced your table, and the case is stronger than "scale-invariant, so it cannot see a scale collapse".

On the spectral path `tb = L⁻¹BL⁻ᵀ` with `|B| = LLᵀ` is symmetric **and** orthogonal — `B` and `|B|` share eigenvectors, so the congruence sends `Λ` to its own signature — hence `tb² = I` and **true κ_after is identically 1 by construction**, at every block size. Verified to ~1e-9:

```
   n  ||tb^2 - I||  true k(tb)  k_corr(tb)  min|diag tb|
   5       9.4e-09           1        1.01          0.99
  12       1.5e-09           1       1.035         0.966
  30      2.88e-09           1       5.832         0.145
  60      4.63e-09           1       4.455         0.113
```

But `tb` is orthogonal, not diagonal: `L` is a Cholesky factor, not the eigenbasis, so `diag(tb)` can be anything in `[-1, 1]` and `_cond_corr`'s `1/√|diag|` normalisation picks up the arbitrary orientation between the two. So on the spectral path the reported number is not merely blind to the ridge's scale failure — it is reporting an artefact, and the artefact grows with block size, i.e. precisely where `--preconditionBlocks none` now sends people. At n=60 the summary would say the transform left κ at 4.5 when it provably left it at 1.

That is also what tips it for me. A documented blind spot would be defensible; a number that drifts away from the truth as the block grows, on the configuration this PR newly recommends, is not something I would want to ship and then have someone tune against.

It gives the change a clean spec and a free assertion too: true κ_after ≡ 1 on the spectral path, so the test can pin it at several sizes instead of only on the 3×3 — which is a much better statement of what the transform does than "reaches the identity up to sign" on one hand-built matrix.

Two cautions, both from trying it:

**1. Do not report it as a decomposition.** I tried `κ_corr(tb) × (max|diag tb| / min|diag tb|)`, on the theory that it separates residual degeneracy from scale collapse and reuses the already-computed `tb`. It is exact when `tb` comes out diagonal — 4 significant figures on both 3×3 blocks — and badly wrong otherwise, overestimating true κ by 13x at n=12 and 1500x at n=30. Report the true κ directly: one SVD of an `m × m` matrix you have already formed, negligible against the eigendecomposition on that path.

**2. Scope the correlation number, don't just keep it.** Your instinct to keep it rather than replace it is right, but it is right for the *ridge* path — my `1.8e+03 -> 1e+04` was a ridge observation, where `tb` is not orthogonal and residual degeneracy is a real thing to measure. As the "after" number on the spectral path it is the artefact above. Either report it for `cond_before` only, where scale invariance is unambiguously the right choice because the parameter units are arbitrary, or make the log say which path it is meaningful on.

On the misreading risk you raise — `1.225 -> 1.4e8` read as preconditioning having made things worse — agreed, and it is the reason to do this as two labelled clauses rather than one arrow. `cond_before` is the only place a single arrow was ever honest.

Haven't looked at #157 yet; will review it separately.

davidwalter2:
Follow-up on the ridge fix, since it is now the only functional change this PR makes to the default path and I wanted to know what to expect from it before running it.

## The fix is purely additive, and provably so

Good news first. Over 4000 random indefinite blocks (n=2–6, spectra spanning 1e-6 to 1e9, 45% negative eigenvalues), every block `main` already factorised comes out with a `chol` identical to **6.7e-16** relative — zero blocks differing by more than 1e-10, no change in true κ anywhere. 1468 were newly factorised.

And it cannot be otherwise, which is worth having in the commit message: `λ_min ≤ min(diag)`, so the only case where `scale` changes at all is `max|diag| = |min(diag)|`, and there try 1's ridge is `1e-8·|min diag| ≤ 1e-8·|λ_min|` — still far short of `|λ_min|`, so it fails either way. From try 2 the schedule is `need/scale` while the applied ridge is `eps·scale`, so `scale` cancels exactly. The 52/4000 blocks that differ at all differ only through `need/scale` then `eps·scale` in floating point.

Also checked the one thing that looked like a second default-path change: the hoisted `np.all(np.isfinite(block))` guard, which the ridge path did not have. Not a behaviour change — a block with a NaN off-diagonal returns `None` on `main` too, just via four failed Choleskys and the "not factorisable" warning rather than an upfront check.

## But the newly-factorised blocks come out worse

This is the part that gives me pause. On 961 synthetic blocks that `main` skipped and the PR now factorises:

| | |
|---|---|
| true κ improved | **0.0%** |
| true κ degraded | **100%** — median **11x**, range 4.3x to 14.4x |

The mechanism is the one this PR already argues, applied to the blocks that are entirely in that regime: an all-negative-diagonal block needs a ridge of at least `|λ_min| ≈ max|diag|`, so the ridge swamps every direction in the block. Your own pinned example is the clean case:

```
diag(-7.5e3, -6.3e4, -2.1e3):   true cond 30 -> 320   (ridge)
                                true cond 30 ->   1   (spectral)
```

So "not as good as spectral … but a usable transform beats none" is the claim I would want to check rather than assume. On this evidence the ridge transform on an all-negative block is not usable-but-worse, it is actively worse than the skip it replaces — and the skip was `main`'s behaviour, so the fix converts a no-op into a 10x degradation on exactly the blocks it newly reaches.

My ensemble is synthetic and deliberately adversarial, so I would not read the 11x as a number to expect. The direction is not an artefact of the ensemble, though; it follows from `ridge ≥ |λ_min| ≈ max|diag|`.

## The proposal

Line 237 already states the principle:

> A block that cannot be factorised is dropped: the others still apply, and a preconditioner must never break a fit.

That is currently applied only to factorisation **failing**, never to factorisation **making things worse**. Extending it — drop the block when the conditioning degrades, with a margin so ties do not churn — makes the ridge fix strictly safe:

- blocks where ridging helps: newly preconditioned, which is the gain the fix is for;
- blocks where it hurts: dropped, i.e. exactly `main`'s behaviour, so nothing is lost relative to today;
- and it applies to the pre-existing ridge path too, not just the blocks this fix newly reaches.

It has to key off the **true** κ, not `_cond_corr`, which is the second reason I think that change belongs in this PR rather than after it. On the random ensemble the correlation metric would have caught 99.7% of these degradations, but on your pinned `diag(-7.5e3, -6.3e4, -2.1e3)` it reports `1 -> 1` for both transforms — a diagonal block has the identity as its correlation matrix before and after, so the metric is blind to precisely the example you chose to pin the fix with. With true κ the guard is exact and needs no threshold tuning.

`cond_after` is already computed on both paths, so the guard is a comparison and a `return None`, and it shares its input with the reporting change. Happy to be wrong about the size of the effect on real blocks — but I would rather the default path not be able to make a fit worse than `main` on a block `main` declined to touch.

lucalavezzo:
Done in `7b57fac`. You convinced me — and having implemented it, your orthogonality argument is the part that made the case, not the blindness.

I re-derived it before changing anything: with `|B| = Q|Λ|Qᵀ = LLᵀ`,

```
tb·tbᵀ = L⁻¹Q Λ|Λ|⁻¹Λ QᵀL⁻ᵀ = L⁻¹Q|Λ|QᵀL⁻ᵀ = L⁻¹|B|L⁻ᵀ = I
```

since `Λ²/|Λ| = |Λ|`, and `tb` is symmetric because `B` is. So `tb² = I` and true κ ≡ 1, as you said.

On random indefinite blocks the artefact grows **faster** than your table showed — presumably because a random symmetric block is more generically oriented than the structured ones you tried:

| n | ‖tb²−I‖ | true κ | reported κ_corr |
|---|---|---|---|
| 5 | 6.7e-15 | 1 | 1.73 |
| 12 | 9.3e-15 | 1 | 7.67 |
| 30 | 5.0e-14 | 1 | 21.3 |
| 60 | 6.8e-13 | 1 | 33.6 |

And on the ridge path `‖tb²−I‖ ≈ 100`, i.e. `tb` is nowhere near orthogonal, which is the concrete version of your point that the correlation number is measuring something real *there* — so your `1.8e+03 -> 1e+04` was not an artefact.

**What I did**, following both cautions:

- the arrow is now the true κ at both ends (`_cond_true`, one SVD of the `tb` both paths already form);
- the degeneracy is kept for `cond_before` **only** — your first option rather than the per-path labelling, since scale invariance is unambiguously right for the incoming block and needs no caveat — and it is its own clause, not the far end of an arrow:

```
ridge     condition number 3.3e+09 -> 1.44e+08, of which degeneracy 1.23
spectral  condition number 3.3e+09 -> 1, of which degeneracy 1.23
```

- not as a decomposition. Reproduced your finding independently: `κ_corr(tb) × (max|diag|/min|diag|)` is exact when `tb` is diagonal and overestimates true κ by **1887×** at m=12 and 193× at m=30 otherwise.

### One refinement to the spec

`true κ_after ≡ 1` holds only where **nothing was floored** — a floored direction is precisely one the congruence no longer preserves, so the identity breaks by construction:

| block | floored | ‖tb²−I‖ | true κ_after |
|---|---|---|---|
| `diag(3, −2, 1)` | 0 | 6.3e-16 | **1** |
| `diag(1, −2, 1e-20)` | 1 | 1.0 | 1.3e5 |
| `diag(1, −2, 0)` | 1 | 1.0 | ∞ |

That is correct behaviour, not a defect — a null direction has no scale to whiten to — but it means the assertion has to be conditioned on `n_floored == 0`, which the code already computes and logs. So κ stays *measured* per block rather than asserted, and there is a second test (`test_flooring_is_why_true_kappa_is_measured_not_assumed`) pinning the broken case so nobody later "simplifies" the measurement into a constant.

The spec test is at m = 5, 12, 30, 60 with `tb² = I` asserted alongside, and it asserts `κ_corr > 5` at m ≥ 12 so the artefact itself is pinned rather than just avoided. You're right that this is a much better statement of what the transform does than "reaches the identity up to sign" on one hand-built 3×3.

95 passed (90 before), black and isort clean. `WHAT THE NUMBERS MEAN` in the module docstring carries the derivation, the measured artefact growth, and the flooring caveat.

lucalavezzo:
Implemented in `851e442`, and you were right to want the claim checked rather than assumed — it was wrong, and the way it was wrong is that I never checked it.

I reported `cond_after = 320` where the block used to be skipped and read that as a success. I never compared 320 to the alternative. Doing nothing leaves that block at **30**, so the fix I described as "a usable transform beats none" was 10.7× worse than the skip it replaced, on the default path. Reproduced your ensemble independently and extended it over off-diagonal strength, since near-diagonal blocks were the obvious way for mine to be unrepresentative:

| off-diag strength | newly factorised | improved | degraded | median ratio |
|---|---|---|---|---|
| 0.02 | 2431 | 0 | 2431 | 10.93 |
| 0.2 | 1646 | 0 | 1646 | 10.86 |
| 0.6 | 1189 | 0 | 1189 | 10.83 |
| 1.0 | 1012 | 0 | 1012 | 10.78 |
| 2.0 | 816 | 0 | 816 | 10.80 |

**0 improved out of 7094**, and the ratio is flat in off-diagonal strength — consistent with your point that the direction follows from `ridge ≥ |λ_min| ≈ max|diag|` rather than from the ensemble. So I've taken the 11× as a real direction and not as a number to expect, as you framed it.

Your proof that the change is inert for blocks `main` already factorised is in the commit message — `λ_min ≤ min(diag)`, so `scale` only changes when `max|diag| = |min(diag)|`, and there try 1 fails either way while from try 2 `scale` cancels between `need/scale` and `eps*scale`. That is worth having written down, because it is the half of the fix that survives.

**The guard**, in `_factorise` so it covers both transforms: drop when true κ degrades by more than `DEGRADE_TOLERANCE` (1%, so a tie doesn't churn a block in and out). And you're right that it has to key off true κ — the correlation number reports `1 -> 1` on the block I'd pinned the fix with, since a diagonal block has the identity as its correlation matrix at both ends. That's a second argument for the reporting change, independent of the first.

### One thing your proposal needed

**The guard must not act on an already-singular block.** My first version broke `test_rank_deficient_block_is_ridged_into_shape`, and correctly so from its own point of view — that block reads `5.28e+16 -> 7.32e+16`, a 1.39× "degradation" between two values past `1/eps`, where true κ has stopped being a measurement. Meanwhile the *correlation* number shows the ridge genuinely helping it: `3.1e+16 -> 1.4e+12`. Ridging a rank-deficient block into shape is the ridge's original purpose, so the guard stands down above `SINGULAR_COND = 1/eps`.

It's the mirror image of the all-negative case — there true κ is informative and correlation is blind; here correlation is informative and true κ is noise — which is a better reason to keep both numbers than the one I gave when I kept the correlation one.

### Net effect, since I oversold it twice

All-negative blocks end up **unpreconditioned, exactly as on `main`**. The `max|diag|` change makes them reachable, and the guard then declines them. What actually improved is that this is now reached by measurement with a logged reason instead of by a scale test that was mislabelled — it always printed `max|diag|` — and that the same protection now covers every other block, including the pre-existing ridge path. Spectral remains the thing that rescues them. The description says this plainly now rather than claiming 21 of 34 blocks were recovered.

Tests: the all-negative case now asserts both that the guard declines it *and* that `_factorise_ridge` alone would have succeeded, so a future change cannot confuse the two rejection reasons; plus a block each transform genuinely helps being kept, and the singular carve-out. 97 passed (95 before), isort/flake8/black clean under the exact CI invocations.

PR #157, opened 2026-09-10
callbacks: make the stall threshold relative, with --stallRelTol

lucalavezzo:
Split out of #156 at @davidwalter2's request. It was the only change there
touching a default-on path (`--earlyStopping` defaults to 20), so it should be
judged on its own rather than carried along with a preconditioner feature.

### What

The stall test was `loss_history[-N] <= loss`, which fires only on **literally**
no improvement over the window. A fit that crawls therefore never satisfies it,
and `--maxRestarts` never gets a chance to rebuild the preconditioner at the
point the fit has actually reached. `--stallRelTol` makes the threshold relative,
scaled by `|ref|` so it means the same thing at loss 1e7 and at loss 1e4.

The default `0.0` is **exactly** the old test — `gained <= 0` ⟺ `ref - loss <= 0`
⟺ `ref <= loss` — so existing callers are bit-identical. That equivalence is what
the whole option rests on, and it is now pinned over five trajectories including
`ref == 0.0`, the one input where the relative form would divide by zero and has
to fall back to the absolute test.

### In fairness about the motivation

Our own fits do **not** exhibit the crawl. Replayed against real trajectories
they keep gaining 0.2–4% per 30 iterations, no threshold up to 1e-3 fires on
them, and the exact test was correctly reporting "not stalled". The one
3720-parameter run that did stop short of a minimum did so on **100 bit-identical
iterations** — the exact test working as intended.

So this is a gap in the trigger's generality, not a fix for an observed failure.
An earlier version of the code comment asserted otherwise; it has been rewritten
to argue the generalisation on its own terms rather than citing a measurement
that does not hold.

### The caveat, since the option is easy to reach for

The criterion **cannot distinguish converged from stalled**. "Flat to 1e-4 over
20 iterations" is also what approaching a minimum looks like, whereas the exact
test only fires once the trust radius has genuinely collapsed. A non-zero value
will therefore also fire near a good minimum and spend a reference-Hessian
evaluation on each restart — measured 130/424/139 s across three rebuilds on one
fit. `RESTART_MIN_IMPROVEMENT = 1e-9` bounds the loop and scipy's `gtol` usually
exits first, so it is not unsafe, but it is a real behaviour change. Hence off by
default, and the `--help` now warns instead of suggesting a value unprompted.

### Tests

89 passing. New in `tests/test_restart.py`:

- `stall_rel_tol=0.0` is exactly the old predicate, over five trajectories
  (healthy descent, hard flat, crawl, worsening, and all-zero);
- a crawl is detected at 1e-3 and invisible at 0.0;
- a healthy descent survives every threshold up to 1e-2;
- the default is pinned in all three places it lives (parser, `Fitter` getattr,
  `FitterCallback` signature).

### DIFF
--- a/rabbit/callbacks.py
+++ b/rabbit/callbacks.py
@@ -23,7 +23,7 @@ logger = logging.child_logger(__name__)
 
 
 class FitterCallback:
-    def __init__(self, xv, early_stopping=-1, snapshotter=None):
+    def __init__(self, xv, early_stopping=-1, snapshotter=None, stall_rel_tol=0.0):
         self.iiter = 0
         self.xval = xv
         # Optional rabbit.snapshot.Snapshotter. The callback is the only place
@@ -38,6 +38,29 @@ class FitterCallback:
         self.t0 = time.time()
 
         self.early_stopping = early_stopping
+        # Relative improvement over the --earlyStopping window below which the
+        # fit counts as stalled. 0.0, the default, is exactly the original test:
+        # `gained <= 0` is `ref - loss <= 0` is `ref <= loss`, so the behaviour
+        # of every existing caller is bit-identical.
+        #
+        # Why make it relative at all. The exact test fires only on LITERALLY no
+        # improvement over the window, so a fit that crawls never satisfies it
+        # and --maxRestarts never gets a chance to rebuild the preconditioner at
+        # the point the fit has actually reached. That is a real gap in the
+        # trigger, but it is NOT a bug we have observed: replayed against our
+        # own trajectories they keep gaining 0.2-4% per 30 iterations, no
+        # threshold up to 1e-3 fires, and the exact test was correctly reporting
+        # "not stalled". Do not read this option as a fix for a known stall.
+        #
+        # And note what it cannot distinguish: "flat to 1e-4 over 20 iterations"
+        # is also what APPROACHING A MINIMUM looks like, whereas the exact test
+        # only fires once the trust radius has genuinely collapsed. So a
+        # non-zero value will tend to fire near a good minimum and spend a full
+        # reference-Hessian evaluation (measured 130-424 s) on each restart.
+        # RESTART_MIN_IMPROVEMENT bounds the loop and scipy's gtol usually exits
+        # first, so it is not unsafe -- but it is a behaviour change, which is
+        # why it is off by default.
+        self.stall_rel_tol = float(stall_rel_tol)
         # set just before raising, so fit() can tell a recoverable stall apart
         # from a genuine error and restart instead of giving up
         self.stopped_early = False
@@ -56,15 +79,23 @@ class FitterCallback:
         if np.isnan(loss):
             raise ValueError(f"Loss value is NaN at iteration {self.iiter}")
 
-        if (
-            self.early_stopping > 0
-            and len(self.loss_history) > self.early_stopping
-            and self.loss_history[-self.early_stopping] <= loss
-        ):
-            self.stopped_early = True
-            raise ValueError(
-                f"No reduction in loss after {self.early_stopping} iterations, early stopping."
-            )
+        if self.early_stopping > 0 and len(self.loss_history) > self.early_stopping:
+            ref = self.loss_history[-self.early_stopping]
+            gained = ref - loss
+            # scale by |ref| so the threshold means the same thing at loss 1e7
+            # and at loss 1e4; ref == 0 falls back to the absolute test
+            budget = self.stall_rel_tol * abs(ref)
+            if gained <= budget:
+                self.stopped_early = True
+                how = (
+                    f"only {gained / abs(ref):.3g} relative"
+                    if ref
+                    else f"only {gained:.3g} absolute"
+                )
+                raise ValueError(
+                    f"Loss improved {how} over {self.early_stopping} iterations "
+                    f"(threshold {self.stall_rel_tol:.3g}), early stopping."
+                )
 
         self.loss_history.append(loss)
         self.time_history.append(elapsed)
--- a/rabbit/callbacks.py
+++ b/rabbit/callbacks.py
@@ -60,6 +60,15 @@ class FitterCallback:
         # RESTART_MIN_IMPROVEMENT bounds the loop and scipy's gtol usually exits
         # first, so it is not unsafe -- but it is a behaviour change, which is
         # why it is off by default.
+        # Negative would silently WEAKEN the test -- it would require the loss
+        # to get worse before declaring a stall, i.e. a quieter trigger than
+        # the default rather than a louder one.
+        if float(stall_rel_tol) < 0.0:
+            raise ValueError(
+                f"stall_rel_tol must be >= 0, got {stall_rel_tol}. A negative "
+                "threshold would require the loss to WORSEN before the fit "
+                "counts as stalled."
+            )
         self.stall_rel_tol = float(stall_rel_tol)
         # set just before raising, so fit() can tell a recoverable stall apart
         # from a genuine error and restart instead of giving up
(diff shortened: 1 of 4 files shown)


rabbit/callbacks.py
```diff
@@ -56,15 +79,23 @@ def __call__(self, intermediate_result):
...
-            raise ValueError(
-                f"No reduction in loss after {self.early_stopping} iterations, early stopping."
-            )
+        if self.early_stopping > 0 and len(self.loss_history) > self.early_stopping:
+            ref = self.loss_history[-self.early_stopping]
+            gained = ref - loss
+            # scale by |ref| so the threshold means the same thing at loss 1e7
+            # and at loss 1e4; ref == 0 falls back to the absolute test
+            budget = self.stall_rel_tol * abs(ref)
+            if gained <= budget:
```
davidwalter2: This is not bit-identical to the old test, and the exception is reachable.

`gained = ref - loss` and `budget = tol * abs(ref)` are both NaN when the loss is pinned at an infinity, and every comparison with NaN is False — so the stall is never detected, where `main`'s `ref <= loss` sees `inf <= inf` and fires. Ran both predicates over nine trajectories:

```
case                       main   PR157
all finite flat            True    True
descent                   False   False
worsening                  True    True
all zero                   True    True
ref=+inf loss finite      False   False
STUCK at +inf              True   False   <-- differs
+inf then flat             True    True
ref=-inf                   True    True
STUCK at -inf              True   False   <-- differs
```

It matters because `+inf` is reachable and unguarded. `__call__` raises on a NaN loss but not on an infinite one, and a Poisson term with a non-positive prediction and non-zero data gives exactly `+inf`. The trust region then rejects every step from a point it cannot improve on, the iterate does not move, and the loss stays at `inf` — which is precisely the collapsed-trust-radius case `--earlyStopping` exists to catch. On `main` that early-stops (and with `--maxRestarts` restarts); here it runs to `maxiter = 200 * nparams`, i.e. ~744k iterations on the 3720-parameter model.

Two lines fix it, and keep the equivalence exact rather than approximate:

```python
budget = self.stall_rel_tol * abs(ref) if self.stall_rel_tol else 0.0
if loss >= ref - budget:
```

Short-circuiting `budget` avoids `0.0 * inf`, and keeping the comparison in `loss >= ref - budget` form rather than `ref - loss <= budget` avoids `inf - inf`. At `tol == 0` it reduces to `loss >= ref`, which *is* the original predicate, infinities included. Checked:

- agrees with `main` on all nine cases above, including both that currently differ;
- identical to this PR on 80,000 random finite trajectories at `tol` = 1e-6, 1e-4, 1e-2, 0.5 — zero disagreements;
- all 17 tests in `test_restart.py` still pass.

Worth deciding separately whether an infinite loss should raise the way NaN does. That's a different behaviour change and I wouldn't put it in this PR, but the asymmetry is what makes this edge reachable at all.

tests/test_restart.py
```diff
@@ -206,3 +207,88 @@ def counting():
...
+    return False, cb
+
+
+# healthy descent, a hard flat, a crawl, and one that gets worse
+TRAJECTORIES = [
+    [10.0, 9.0, 8.0, 7.0, 6.0, 5.0],
+    [10.0, 9.0, 9.0, 9.0, 9.0, 9.0],
+    [1e4, 1e4 * (1 - 1e-5), 1e4 * (1 - 2e-5), 1e4 * (1 - 3e-5), 1e4 * (1 - 4e-5)],
+    [10.0, 9.0, 9.5, 10.0, 10.5, 11.0],
+    [0.0, 0.0, 0.0, 0.0, 0.0],
```
davidwalter2, line 244: The five trajectories are well chosen — `ref == 0.0` in particular is the input I'd have gone looking for. But they're all finite, which is why the equivalence claim reads as proven when it isn't.

One line catches it:

```python
    [np.inf, np.inf, np.inf, np.inf, np.inf],
```

Against the current code that gives `assert False == True` on `test_stall_rel_tol_zero_is_exactly_the_original_test[losses5]`; with the two-line fix from the `callbacks.py` comment it passes. Since this test re-derives the old predicate inline rather than restating the algebra, it's already the right harness for this — it just needs the input where the algebra stops holding.

Also worth a `[-np.inf, ...]` row for symmetry: it differs too, though I'd call that one genuinely unreachable.

rabbit/parsing.py
```diff
@@ -192,6 +192,22 @@ def common_parser():
         "its maxiter defaults to 200*nparams, so a stalled fit spins for hours making "
         "no progress rather than exiting. Specify -1 to disable.",
     )
+    parser.add_argument(
+        "--stallRelTol",
+        default=0.0,
```
davidwalter2, line 213: Nit, but a footgun on a default-on path: nothing validates the value.

A negative value silently *weakens* the test — `gained <= -1e-4·|ref|` requires the loss to get **worse** before a stall is declared, so `--stallRelTol -1e-4` is a quieter trigger than the default rather than a louder one.

More reachable: the upper end has no safe bound documented. A geometric descent at factor `f` per iteration over an `N`-iteration window is flagged as stalled exactly when `tol >= 1 - f^N`, so the safe ceiling depends on `--earlyStopping` and on how fast the fit is actually converging:

```
f=0.5    N=3    boundary 1-f^N = 0.875
f=0.9    N=20   boundary          0.878
f=0.99   N=20   boundary          0.182
f=0.999  N=20   boundary          0.0198
```

So on a fit converging at 0.1% per iteration — slow but perfectly healthy — `--stallRelTol 0.05` declares it stalled. `test_stall_rel_tol_leaves_a_healthy_descent_alone` stops at 1e-2, which is below every boundary in that table, so the suite can't see this.

The help already says "leave it at 0 unless a fit is demonstrably crawling", which covers most of it. I'd add either a `< 0` rejection, or a sentence giving `1 - f^N` as the ceiling — it's the one number that tells a user what "too big" means for *their* window rather than in the abstract.

rabbit/callbacks.py
```diff
@@ -56,15 +79,23 @@ def __call__(self, intermediate_result):
...
+                self.stopped_early = True
+                how = (
+                    f"only {gained / abs(ref):.3g} relative"
+                    if ref
+                    else f"only {gained:.3g} absolute"
+                )
+                raise ValueError(
+                    f"Loss improved {how} over {self.early_stopping} iterations "
```
davidwalter2, line 126: Nit: this changes the default path's message. `main` raises

```
No reduction in loss after 20 iterations, early stopping.
```

and this raises

```
Loss improved only 0 relative over 20 iterations (threshold 0), early stopping.
```

Nothing parses it — I grepped the repo when reviewing #156 and there are no other references to the string. But "improved only 0 relative" is a slightly awkward way to say "did not improve", and it's what every existing user sees. Keeping the original wording when `stall_rel_tol == 0.0` and using the relative phrasing only when the threshold is actually in play would leave the default path unchanged in the logs too, which is the same standard the rest of the PR holds itself to.

davidwalter2 (commented):
Reviewed on its own terms, as split out. Verified rather than taken on trust: `89 passed`, black clean, no new flake8 `F` findings, and the equivalence argument checks out on every finite trajectory I could throw at it — 80,000 random ones across four thresholds.

Splitting it was right, and the version that came out is a better change than the one in #156: the motivation is now stated honestly (a gap in generality, not a fix for an observed stall), the converged-vs-stalled caveat is in the code and the `--help` rather than only in a review thread, and the default is pinned in all three places it lives — which is exactly what I'd asked for on the ridge default and didn't get there.

**One thing blocks it.** The headline claim is that `0.0` is bit-identical to the old test, and it isn't: a loss pinned at an infinity is detected as a stall on `main` and is not detected here. Details inline — it's a two-line fix, and one line added to `TRAJECTORIES` makes your own parametrised test catch it.

Everything else is a nit.

rabbit/callbacks.py
lucalavezzo: You're right, and the claim was the whole basis for calling the option opt-in, so it was the wrong one to get wrong. Fixed with your two lines verbatim in `f9bf2f3`.

Reproduced your table exactly, including both rows that differed:

```
case                       main   before   after
all finite flat            True     True    True
descent                   False    False   False
worsening                  True     True    True
all zero                   True     True    True
ref=+inf loss finite      False    False   False
STUCK at +inf              True    False    True   <-- was wrong
+inf then flat             True     True    True
ref=-inf                   True     True    True
STUCK at -inf              True    False    True   <-- was wrong
```

And confirmed it is a pure edge-case repair rather than a behaviour change: identical to the previous version on 80,000 random finite trajectories at `tol` = 1e-6, 1e-4, 1e-2, 0.5 — zero disagreements. Your framing of *why* to write it as `loss >= ref - budget` is now in the comment, since the two forms look interchangeable and the reason they aren't is not obvious from the code.

Agreed on leaving "should an infinite loss raise the way NaN does" out of this PR. It is a separate behaviour change, and the fix above makes the stall detectable rather than depending on that asymmetry being resolved.

tests/test_restart.py
lucalavezzo: Added, both rows, plus two mixed cases (`[inf, inf, 5, 5, 5, 5]` and `[inf, inf, inf, inf, 1, 1]`) so the boundary between "was infinite" and "is infinite" is covered rather than just the endpoints.

Mutation-checked rather than watched to pass — against the old predicate it gives exactly what you predicted:

```
FAILED test_stall_rel_tol_zero_is_exactly_the_original_test[losses5]  assert False == True
FAILED test_stall_rel_tol_zero_is_exactly_the_original_test[losses6]  assert False == True
2 failed, 21 passed
```

and 23 passed with the fix. You're right that this test was already the right harness — because it re-derives the old predicate inline rather than restating the algebra, it needed nothing but the input where the algebra stops holding.

I've kept `[-inf]*5` even though you'd call it unreachable: it costs one line, and the asymmetry between the two would have been a strange thing to leave in place deliberately.

rabbit/parsing.py
lucalavezzo: Both taken.

**Negative rejected**, at the argparse boundary (so it fails before a fit starts rather than 100 iterations in) and in `FitterCallback` for programmatic callers. You're right that it's the direction that makes it a footgun: it *weakens* the test, so someone reaching for it to catch a stall sooner would get a quieter trigger than the default.

**The ceiling is now the formula rather than a vibe**, which is the part I'd have got wrong on my own — I'd have written "don't go above ~1e-2" and that is exactly the bound your table shows to be window-dependent:

> a descent at factor f per iteration is declared stalled exactly when `tol >= 1 - f^N` for a window of `N = --earlyStopping`, so a healthy fit converging at 0.1% per iteration over a window of 20 is flagged by anything above 0.0198. Pick the value from your own f and N rather than in the abstract.

And your observation about the test is the more useful half: `test_stall_rel_tol_leaves_a_healthy_descent_alone` stops at 1e-2, which is below every boundary in your table, so the suite is structurally unable to catch this. I have not extended it to sample above the boundary, because a test asserting "this healthy descent IS flagged as stalled" would pin the behaviour the `--help` is warning against. If you'd rather have it pinned as documented-and-intended, say so and I'll add it.

rabbit/callbacks.py
lucalavezzo: Fixed — the default path keeps `main`'s wording exactly, and the relative phrasing appears only when a threshold is actually in play. Pinned by a test so it can't drift back.

You're right that this is the same standard the rest of the PR holds itself to, and I'd applied it to the numerics and then not to the string every existing user reads. Thanks for grepping for parsers of it; I'd have assumed rather than checked.

PR #158, opened 2026-09-10
Compute global impacts from an external postfit covariance

davidwalter2:
`--globalImpacts` with `--externalPostfit` silently computes nothing.

Global impacts need the parameter values and the covariance, never the Hessian:
`Fitter.global_impacts_parms()` takes `fitter.cov` and no `hess` argument, and
`gaussian_global_impacts_parms()` works from `_dxdvars()`. But both sat inside the
block gated on

```python
if (dofit or args.externalPostfit is None or not external_cov_loaded):
```

so pointing `--externalPostfit` at a result that already carries a covariance — the
natural thing after the two-pass `--noHessian` recipe — skipped the whole block and
wrote no impacts at all, while exiting 0 and logging `Results written in file`. The
only symptom is the absent key in the output.

This moves them out of the gate. On a 6538-parameter W+Z fit it turns a job that
silently did nothing, or ~9 min of recomputing a Hessian already on disk, into ~60 s.

Verified to agree with recomputation to **1.9e-11 MeV** (relative 1.6e-10) across all
74/76 impact groups on two independent models, so reusing the stored covariance is not
an approximation.

### `--doImpacts` is different

It genuinely needs the Hessian: `impacts_parms()` uses it for `_cov_stat_floating()`,
and computes a second, unprofiled Hessian for the no-`binByBinStat` variant. Neither is
recoverable from the covariance, so it now raises with a message pointing at the
pre-Hessian result instead of silently producing nothing.

### Also closed

- `--globalImpacts`/`--gaussianGlobalImpacts` with no covariance available at all
  (`--noHessian` and no external one) also skipped silently; now raises.
- `parms_variances` stayed NaN when the covariance came from `--externalPostfit`, even
  though `fitter.cov` was populated; now filled from its diagonal.

### Tests

On top of this branch (i.e. `main` + this commit):

- `--globalImpacts --externalPostfit <result with cov>` → `global_impacts (1, 6685)`,
  `global_impacts_grouped (1, 76)`, 58 s. Previously: no impacts, rc=0.
- `--doImpacts --externalPostfit <result with cov>` → raises with guidance.
  Previously: no impacts, rc=0.

https://claude.ai/code/session_012sxpuabhZaYxUN3erGezFa


bin/rabbit_fit.py
```diff
@@ -699,6 +696,64 @@ def fit(args, fitter, ws, dofit=True):
...
+        ext_var = tf.linalg.diag_part(fitter.cov).numpy()
+        ext_mask = getattr(fitter, "external_cov_mask", None)
+        if ext_mask is not None and not ext_mask.all():
+            n_missing = int((~ext_mask).sum())
+            logger.warning(
+                f"--externalPostfit covered {int(ext_mask.sum())} of "
+                f"{ext_mask.size} parameters; the remaining {n_missing} have no "
+                "postfit uncertainty and are written as NaN. Global impacts for "
+                "them are computed against a covariance block that is still the "
+                "prefit diagonal, so treat those rows as unreliable."
```
lucalavezzo: With partial coverage it isn't only the uncovered rows that are unreliable — the impacts of the **covered** POIs are silently wrong too, and they are not masked.

`self.cov` is initialized to the prefit diagonal (`fitter.py:813`) and `load_fitresult` only overwrites the `np.ix_(idxs, idxs)` block, so every covered↔uncovered cross-term stays exactly zero. In `global_impacts.py`:

```python
cov_dexpdx = tf.matmul(cov, dexpdx, transpose_b=True)  # column of cov per POI -> 0 on uncovered rows
impacts_x0 = sc @ cov_dexpdx                           # uncovered nuisances contribute exactly 0
var_nobs   = var_total - var_x0                        # their missing variance lands in the stat term
```

So a covered POI gets zero impact from every uncovered nuisance, and that variance is reassigned to the data-stat component. Nothing about the written output looks wrong, and the job exits 0.

Since `--globalImpacts` on a partial covariance can't produce trustworthy numbers for *any* parameter, I'd raise here when `not ext_mask.all()` and global impacts were requested, rather than warn. If you'd rather keep it a warning, the wording needs to cover the covered rows as well — "treat those rows as unreliable" points at the wrong set.

bin/rabbit_fit.py
```diff
@@ -699,6 +696,64 @@ def fit(args, fitter, ws, dofit=True):
...
+    # Global impacts need the covariance and the parameter values, not the
+    # Hessian, so they run whether it was recomputed above or loaded from
+    # --externalPostfit.
+    if args.globalImpacts or args.gaussianGlobalImpacts:
+        if not cov_available:
+            raise Exception(
+                "--globalImpacts/--gaussianGlobalImpacts need the parameter "
+                "covariance. Either let it be computed here (drop --noHessian) "
+                "or supply a --externalPostfit result that contains one."
+            )
```
lucalavezzo, line 933: This raise fires inside `fit()`, after `fitter.minimize()` (`:584`) and before `dump_and_flush`, so `--noHessian --globalImpacts` now throws away a completed fit. Previously it ran the fit and just omitted the impacts, which was the silent-failure you're fixing — but paying for the fix with the fit itself is a regression for anyone hitting it.

It belongs in the up-front `--noHessian` incompatibility list in `main()` (`:901`), alongside `--doImpacts`, `--computeVariations` etc., so it fails before any work is done.

One other thing: `--noEDM` without `--noHessian` also reaches this raise (neither the `not args.noEDM and not args.noHessian` branch nor the `elif not args.noEDM` branch runs, so `cov_available` stays `False`), and those users are told to "drop `--noHessian`", a flag they never passed.

lucalavezzo (commented):
Two comments on the new external-covariance branch. The core change looks right to me — `global_impacts_parms` and `gaussian_global_impacts_parms` both go through `self.cov` only (`fitter.py:1131`, `_dxdvars` at `:1409`), no Hessian anywhere, so gating them on the recompute was a real bug, and the `--doImpacts` split is correct since `impacts_parms` genuinely needs `hess` for `_cov_stat_floating` plus the second unprofiled Hessian.

PR #159, opened 2026-09-10
blinding: additive offsets for POIs a model declares physical

lucalavezzo:
rabbit blinds a POI **multiplicatively**, `poi * exp(N(0, 5))`. That is right for the default `Mu`: a signal strength centred at 1 which scales *yields*, so any value it is handed stays evaluable, positivity is preserved, and `Mu` and its uncertainty are scaled by the same factor — leaving the *relative* uncertainty, which is what such a measurement quotes, intact.

It is the wrong form for a POI that is a physical parameter fed into a calculation, in two ways.

**The calculation can be asked for a value it cannot evaluate.** `self.x` is initialised to the model's declared default in the *blinded* frame (`fitter.py:363-370`, `:782-785`), so iteration 0 hands the model `xparamdefault * offset`. In an α_s fit that meant SCETlib was evaluated far outside its physical range: starting loss 4.6e7 against a converged 361, with no usable gradient to recover from.

**And the reported uncertainty is destroyed.** The reported coordinate is `poi_true / offset`, so the curvature scales as `offset²` and the reported σ — along with the POI row of the covariance and every impact on that POI — is divided by the random factor. Only the relative uncertainty survives. For a parameter whose *absolute* uncertainty is the result, that is the number you needed.

## The change

`ParamModel.blind_additive = True` opts a model's POIs into additive blinding: the same deterministic `N(0, 5)` draw, applied in the parameter's own fit units. A translation has unit Jacobian, so the covariance, the uncertainties and the impacts come out **exactly** unblinded while the central value stays hidden. This is the form `get_theta` already uses for nuisances of interest — so a physical POI now gets the treatment a NOI would have had.

**Default `False`**, so an undeclared model keeps precisely its current arithmetic: `offset_poi = 1` with `offset_poi_add = 0` is the identity. `test_multiplicative_path_unchanged` pins that as a regression guard.

Note this is not about the central value being 0 rather than 1. Writing `alpha_s = anchor * r` with `r` a POI centred at 1 and blinding it gives `anchor * r * offset` — the original failure again. The distinction is *what the POI multiplies*.

## Three details worth review attention

- **`set_blinding_offsets` now holds the physical point fixed** across a change of frame, shifting `x` by `off_old - off_new` on the additive slots. Without it, additive blinding would move the fit's start to `xparamdefault + offset` — the same class of failure as above. The update is self-idempotent. Multiplicative slots are deliberately left alone: their physical start is preserved only when the declared default is 0, and changing that would alter results for existing users.
- **The constraint term needs no change.** `lc` compares `get_x()` against `x0` and both are in the model frame, so a Gaussian prior on an additively blinded POI stays exactly correct — the minimum simply sits at `mean - offset` internally. `test_x0_untouched_by_arming` pins `x0` against a well-meant "symmetrisation" later.
- **Additive blinding is refused with `allowNegativeParam=False`.** That branch squares `xpoi` to keep the model-frame value positive, and an additive offset on `x²` can hand `compute()` a negative POI, destroying the only guarantee the squared storage provides.

## Tests

`tests/test_blinding.py`, eight checks. There was no blinding coverage before this.

Every check asserts an **invariance**, so no test prints, returns or asserts on an offset value. Two guards make the Hessian test meaningful, both commented in place:

- the toy model's default is deliberately **non-zero** — a zero default satisfies start-invariance by accident when the offset is multiplicative, since `0 * offset = 0`;
- it assigns **Asimov data**, because with `nobs = 0` the Poisson term is linear in the yields, the POI curvature is exactly zero, and comparing Hessians would pass vacuously through `isclose(0, 0)`.

A positive control is included, so the suite demonstrates that the multiplicative path really does scale the POI Hessian rather than only that the additive path is inert.

## Verification

Full suite green (**216 passed** on the top of the stack), lint clean. `tests/test_blinding.py` is now registered in the `unit-tests` CI matrix — see the review thread; it previously ran only under `python tests/<file>` and errored under `pytest`.

**The evidence for exactness is the toy, not the analysis run.** On the toy the additive path agrees with unblinded to `0.000e+00` relative, while the multiplicative path suppresses the POI Hessian element by 2.1e-08 (σ inflated ~10⁴ at random). That is the measurement: a translation has unit Jacobian, so the difference should be, and is, exactly zero.

The real-analysis number is a **smoke test**, not a measurement of agreement. A blinded data fit and an unblinded Asimov fit on the same card give σ within 3.8% (0.001321 vs 0.001371), but the two Hessians sit at different parameter points — the Asimov arm converged to EDM 1.58e-27 with 0 of 3720 negative eigenvalues, the data arm stopped at EDM 1.4e-3. A residual dominated by one arm not having converged neither confirms nor contradicts the exactness claim; it only shows the machinery runs end to end on a real card. (Thanks @davidwalter2 — the original text presented this as supporting evidence, which read the wrong way.)

## Review changes

- **the tests now run** — converted from a hand-rolled `main()` to the repo's `pytest.main` + module-scoped fixture pattern, and registered in CI;
- **`ParamModel.blind_additive_scale`** (default 1.0, bit-identical to the previous draw) plus a startup warning when the offset is small against a prefit σ, because the additive draw — unlike `exp(N(0, 5))` — is not scale free;
- the composite's **blinding-form override is now announced** with a `logger.info` naming both sides.


rabbit/param_models/param_model.py
```diff
@@ -193,6 +220,14 @@ def __init__(
                 + [v[m.npoi :] for v, m in zip(means, param_models)]
             )
 
+        # Blinding form: a boolean, so unlike prior_sigmas there is no index
+        # permutation to track. Any submodel asking for additive blinding makes
+        # the composite additive -- the composite's POI block is the
+        # concatenation of the submodels' POIs, and mixing forms within one POI
+        # block is exactly what the Fitter refuses for a blinding group.
+        if any(getattr(m, "blind_additive", False) for m in param_models):
```
davidwalter2, line 228: A submodel that explicitly asked for multiplicative blinding gets silently switched to additive by a sibling. Verified:

```
composite of [ToyModel(blind_additive=True), ToyModel(blind_additive=False)]
  -> composite.blind_additive = True
```

The comment argues the composite's POI block is a concatenation and mixing forms within one block is what the Fitter refuses, which is fair — but the resolution currently favours one submodel's declaration over another's with no signal. Someone composing an α_s model with a signal-strength model gets the signal strength's blinding form changed and no way to notice short of reading this line.

The dangerous version of this is already blocked, and worth saying so: `CompositeParamModel` rejects mixed `allowNegativeParam` a few lines above, so an additive model composed with a default `Mu` fails there before it can reach the Fitter's positivity check. So this is about the surviving case — all submodels `allowNegativeParam=True`, only one additive — where the override is silent but harmless to correctness.

A `logger.info` naming the submodels whose form was overridden would be enough. Alternatively make it explicit and refuse a mixed declaration the way `allowNegativeParam` does, since "any one makes all" is a policy the caller may not want inferred.

rabbit/fitter.py
```diff
@@ -650,31 +671,77 @@ def deterministic_random_from_string(s, mean=0.0, std=5.0):
...
         for i in range(self.param_model.npoi):
             param = self.param_model.params[i]
             if param in unblind_parameters:
                 continue
             seed = param_to_seed.get(param, param)
             logger.debug(f"Blind parameter {param} (seed='{seed}')")
             value = deterministic_random_from_string(seed)
-            self._blinding_values_poi[i] = np.exp(value)
+            if self._blind_additive:
+                self._blinding_values_poi_add[i] = value
```
davidwalter2, line 700: The additive draw has a **fixed absolute scale**, so how well it blinds depends entirely on the POI's units — and unlike the multiplicative form, nothing bounds it.

`exp(N(0, 5))` is scale-free: it spans roughly `e^±10` whatever `poi` means, so a multiplicatively blinded value is hidden by orders of magnitude regardless of units. `+ N(0, 5)` is not. Measured on the toy model here, and extrapolated across plausible POI scales:

```
toy POI: offset=-8.86, sigma=0.135  ->  hides 65.7 sigma      fine

sigma=0.001   a typical |offset|~4 hides   4000 sigma         alpha_s: fine
sigma=0.05                                   80 sigma         fine
sigma=1.0                                     4 sigma         marginal
sigma=5.0                                   0.8 sigma   <--   effectively unblinded
sigma=20.0                                  0.2 sigma   <--   effectively unblinded
```

The motivating case is comfortably safe — α_s at σ ~ 1e-3 is hidden by thousands of σ. The problem is the next user. A POI in units where its uncertainty is O(1) or larger gets a blinded value from which the true one is recoverable to a fraction of σ, and the code, the docstring and the `ParamModel` comment block all present `blind_additive = True` as simply "the right form for a physical parameter" with no mention that its strength is unit-dependent.

Two ways out, either fine by me:

1. Scale the draw by the parameter's prefit uncertainty — `value * sigma_prefit[i]` — which restores the scale-freedom the multiplicative form has and makes "N(0, 5)" mean the same thing for every POI.
2. Keep the absolute draw but warn when it is small against the prefit σ, e.g. `|offset| < 5 * sigma_prefit`, so a model whose units make this ineffective says so at startup rather than silently.

I would not block on which, but I would not ship the fixed absolute draw with no guard — the failure is silent, and the thing it fails at is blinding.

tests/test_blinding.py
```diff
@@ -0,0 +1,285 @@
...
+
+    Required for the Hessian checks to mean anything: with nobs = 0 the Poisson
+    term ``sum(nexp - nobs*log nexp)`` collapses to ``sum(nexp)``, which is
+    LINEAR in the yields, so the POI curvature is exactly zero and a comparison
+    of Hessians passes vacuously via ``isclose(0, 0)``.
+    """
+    return f.expected_yield()
+
+
+def test_model_sees_physical_while_x_is_blinded(path):
```
davidwalter2, line 125: This is where the missing fixture bites: `path` is a bare parameter with no fixture behind it and no `conftest.py` in the PR, so all eight tests error at setup and none of them executes.

```
ERROR tests/test_blinding.py::test_model_sees_physical_while_x_is_blinded
E       fixture 'path' not found
...
81 passed, 8 errors
```

`make_tensor(path)` and `build(path, ...)` both want a written tensor, so the fixture has to create one:

```python
@pytest.fixture(scope="module")
def path(tmp_path_factory):
    p = str(tmp_path_factory.mktemp("blind") / "t.hdf5")
    make_tensor(p)
    return p
```

With that, all eight pass in 4.85 s. `scope="module"` because `make_tensor` is deterministic (`np.random.seed(1234)`) and every test reads the same tensor — building it eight times would be waste.

Worth a look at why this got missed locally, since the same gap would hide a genuine failure just as well as this one: if the fixture existed in an uncommitted `conftest.py`, `git status` would have shown it as untracked rather than as a modification, which is easy to skim past.

davidwalter2 (commented):
Reviewed the mechanism and then ran it. The physics is right and the motivation is real — a translation has unit Jacobian, so an additively blinded POI keeps its covariance, σ and impacts exactly, while the multiplicative form divides all three by the random factor. For a parameter whose *absolute* uncertainty is the result, that is the number you needed, and the α_s failure mode you describe (iteration 0 handing SCETlib `xparamdefault * offset`) is a genuine bug that this fixes. I traced the `defaultassign` ordering — `xdefaultassign` subtracts the armed offset, then `set_blinding_offsets(False)` adds it back, so `x` lands on `x0default` and the driver's later arming moves the physical start by zero. That composes correctly, and the idempotence argument holds.

Three things before merge, one of them blocking.

**Blocking: none of the eight tests run.** Every one takes a `path` argument, nothing supplies it, and there is no `tests/conftest.py` in the PR:

```
81 passed, 10429 warnings, 8 errors
E       fixture 'path' not found     (x8, one per test in test_blinding.py)
```

So the description's "Full suite green on this base (12 files)" cannot have been run against what was committed — and since setup errors fail pytest, this is red in CI rather than silently green.

The good news is that it is only the wiring. I added six lines:

```python
@pytest.fixture(scope="module")
def path(tmp_path_factory):
    p = str(tmp_path_factory.mktemp("blind") / "t.hdf5")
    make_tensor(p)
    return p
```

and all eight pass in 4.85 s. The tests themselves are well built — every one asserts an invariance rather than an offset value, and the two guards you call out (non-zero model default, Asimov data) are exactly the two ways this suite could have passed vacuously. Worth keeping the comments explaining them.

**The other two are inline**: the composite silently overriding a submodel's declared blinding form, and the additive draw having a fixed absolute scale so that how well it blinds depends on the POI's units.

**On the validation.** The real-analysis number is presented as support but reads the other way to me. The mechanism claim is that additive blinding is *exact* — unit Jacobian, zero difference. The observed 3.8% (0.001321 vs 0.001371) is then attributed to the data arm stopping at EDM 1.4e-3 while the Asimov arm reached 1.58e-27. That is a statement about one arm not having converged, so it neither confirms nor contradicts the exactness claim; the toy's `0.000e+00 relative` is the actual evidence. I would either re-run the data arm to convergence, or lead with the toy and present the analysis run as a smoke test rather than a measurement of agreement.

davidwalter2:
Related: opened #160 for moving this out of `fitter.py` into a `Blinding` collaborator, on the `BinByBinStat` / `Preconditioner` pattern.

Deliberately **after** this PR rather than folded into it — mixing a behaviour change with a file move makes the diff unreviewable, and this PR is what clarifies what the interface should be.

Two things from this review feed into it: the `set_blinding_offsets` shift of `self.x` would be cleaner as a returned shift the Fitter applies (making "arming blinding moves the fit's start" visible at the call site rather than a side effect in a setter), and the additive-offset scale question has no natural home today — a `Blinding` class would own that policy.

tests/test_blinding.py
lucalavezzo: Fixed, and you were right about how it got missed.

The cause wasn't an uncommitted `conftest.py` — it was that this file shipped a **hand-rolled runner**: a `main()` looping over a list of functions and passing `path` positionally. Run as `python tests/test_blinding.py` it works and prints `ALL CHECKS PASSED`, which is what I did. Under pytest the same functions are collected as tests with an unsatisfied `path` argument, so all eight error at setup. Green one way, red the other, from the same file.

So rather than add a `conftest.py`, I converted it to the pattern this repo already has (`test_bbstat.py`, and the two blinding suites added later in the stack):

```python
@pytest.fixture(scope="module")
def path():
    with tempfile.TemporaryDirectory() as d:
        p = os.path.join(d, "blinding_tensor.hdf5")
        make_tensor(p)
        yield p

if __name__ == "__main__":
    sys.exit(pytest.main([__file__, "-v"]))
```

Module-scoped for the reason you gave — `make_tensor` is seeded, every test reads the same tensor. Now correct under `pytest tests/` *and* under `python tests/test_blinding.py`.

The more important half: **`test_blinding.py` is now registered in the `unit-tests` CI matrix.** That job is the only thing that runs `python tests/<file>`, and it was never going to catch this because the file wasn't in it. (The hodor bot raised the same gap independently as a P2 on #171, for the two sibling files — all three are in the matrix now.)

`11 passed` — the original eight plus three new ones for the scale knob below.

One unrelated thing I hit while checking this, in case it explains your `81 passed`: `tests/test_external_nll_constant.py` fails to import under plain `pytest tests/` (`ModuleNotFoundError: No module named 'test_external_term'`) on unmodified `main` as well. Not from this PR, and I've left it alone.

rabbit/param_models/param_model.py
lucalavezzo: Agreed, and taken — with a `logger.info` rather than a refusal.

```
CompositeParamModel: blinding form set to ADDITIVE for the whole POI block
because ['ToyModel'] declared blind_additive=True; this overrides the
(default multiplicative) form of ['Mu'].
```

I went back and forth on refusing it the way `allowNegativeParam` does, and landed on not refusing, because the mix isn't the pathology here — it's the **motivating case**. An additively blinded physical POI composed with a default signal strength *is* a mixed declaration, and refusing would block exactly the composition the feature exists for, with no way out short of making `Mu` lie about its preference.

Your framing of the surviving case is what convinced me it's safe to override rather than refuse: the dangerous mix (an additive offset reaching a squared POI) is already refused upstream, so what's left is all-`allowNegativeParam=True`, where both forms hide the value and the override costs the caller nothing but surprise. Surprise is what the log fixes.

That reasoning is now in the comment at the call site too, so the next reader doesn't have to re-derive why this one infers and the one above it refuses.

rabbit/fitter.py
lucalavezzo: This is the one I want to push back on slightly, then agree with.

**Your diagnosis is right** — the additive draw has a fixed absolute scale and nothing bounds it, and the docstring presenting `blind_additive = True` as simply "the right form for a physical parameter" is a real omission. I've fixed that.

**But your option 1 is not safe in rabbit specifically**, and it's worth spelling out because it fails in the same direction as the bug it's meant to fix, only harder.

A POI has a prefit σ here only where the model declared a Gaussian prior on it. `indata.constraintweights` covers the *nuisances*; the POI block gets its widths from `param_model.prior_sigmas`, which is optional and which by convention most models leave unset for POIs (`fitter.py`, the `prior_sigmas` block — entries that are non-finite or 0 leave the parameter free). For an unconstrained POI — the common case, and the α_s case — `sigma_prefit` is 0, so:

```
offset = value * sigma_prefit[i] = value * 0 = 0
```

i.e. blinding switches off completely and silently. That's your failure mode with the dial at maximum.

So I've done **option 2 plus a declared scale**:

- `ParamModel.blind_additive_scale`, default `1.0`. The offset drawn is `blind_additive_scale * N(0, 5)`. The model is the only thing that knows its own units, and this is the same kind of declaration `blind_additive` already is. At the default it is bit-identical to the current draw — `test_default_scale_is_the_historical_draw` pins that.
- `Fitter._warn_if_additive_blinding_is_weak`, firing at startup wherever a prefit σ *does* exist and `|offset| < 5 * sigma_prefit`:

```
WARNING Additive blinding may be INEFFECTIVE for 1 of 1 POIs: the drawn offset
is smaller than 5 prefit sigma, so the true value is recoverable to within a
few sigma of the blinded one. alphaS (|offset|=8.856, 8.9e-06 sigma). Raise
param_model.blind_additive_scale to a few times the expected uncertainty, in
the POI's own fit units.
```

Three tests: the scale rescales the draw and nothing else (same seed, same sign — a rescale, not a different random number), the default is the historical draw exactly, and the warning fires on a wide prior and stays quiet on a tight one.

On your table: α_s at σ ~ 1e-3 is hidden by thousands of σ at the default scale and needs nothing. Your σ = 5 and σ = 20 rows are now either warned about or fixed by one declaration on the model.

Where I think this genuinely wants to end up is #160 — a `Blinding` collaborator owning the scale policy, with the check unit-testable without constructing a Fitter. The knob and the warning are the version of that which fits in this PR.

lucalavezzo:
## Restacked

| | PR | base |
|---|---|---|

davidwalter2:
Reviewed at `54a4cdc`, against `main`. The restack made this much easier to read — thanks.

The design is right and the reasoning comments are genuinely good. The affine form with `(1, 0)` as the identity keeps unaffected analyses on exactly their current arithmetic; `set_blinding_offsets` holding the physical point fixed and being self-idempotent is the correct invariant; the `xdefaultassign` fix closes a real latent trap ("today's driver only by accident of ordering" is exactly right); and I appreciate that `_warn_if_additive_blinding_is_weak` documents *why* scaling by the prior sigma would be unsafe rather than just not doing it.

Two findings.

---

### [P2] `CompositeParamModel` propagates `blind_additive` but drops `blind_additive_scale`

The composite sets `self.blind_additive = True`, and nothing sets `blind_additive_scale`. The Fitter reads it off the composite with `getattr(self.param_model, "blind_additive_scale", 1.0)`, so a declared scale is silently replaced by the default. Proven rather than inferred:

```
submodel   blind_additive=True  scale=7.0
composite  blind_additive=True  scale=1.0   <-- what Fitter reads
RESULT: SCALE LOST (7.0 -> 1.0); offsets for the SAME name differ by x7
```

(`CompositeParamModel([Physical(blind_additive=True, blind_additive_scale=7.0), Plain()])`.)

Two consequences:

1. **Blinding strength changes silently on compositing**, in the under-blinding direction for exactly the case the docstring describes — *"a POI in units where sigma is O(1) or larger should set this to a few times its expected uncertainty"*. A model that correctly declares `scale=7` gets `scale=1` the moment it is composited, and the `_warn_if_additive_blinding_is_weak` backstop only fires where `prior_sigmas` exists, so for an unconstrained POI it is silent.
2. **It breaks #161's warm start** — details on that PR.

Nothing covers it: `blind_additive_scale` appears 6 times in `test_blinding.py` and 0 times in `test_saturated_composite.py`.

Worth a deliberate policy decision rather than just copying a float through. The scale is *per-parameter units*, so one composite-wide value is arguably the wrong shape — a per-POI vector (each submodel contributing its own scale over its own POI slice, default 1.0) would compose without any "any one wins" rule and would be the natural analogue of how `prior_sigmas` is already permuted here.

### [P3] `additive_models` doesn't filter `npoi > 0`, but `overridden` does

```python
additive_models = [... for m in param_models if getattr(m, "blind_additive", False)]
overridden      = [... for m in param_models if m.npoi > 0 and not getattr(m, "blind_additive", False)]
```

A submodel with no POIs that declares `blind_additive=True` flips the whole composite to additive and is then named in the log line as the reason. Adding `m.npoi > 0` to the first comprehension makes the two consistent.

---

Everything else checks out. Happy to see this land once the composite question is settled.

lucalavezzo:
Confirmed and fixed — and thank you for proving it rather than describing it, because my first instinct was to propagate a single float and your last paragraph is the reason that's wrong.

**It's a per-POI vector now.** The scale is in each parameter's own units, so there is no correct single composite value: `max`, `first`, and "any one wins" all silently rescale somebody. Each submodel contributes its own declaration over its own POI slice, defaulting to 1.0 — the same treatment `prior_sigmas` already gets a few lines up:

```python
self.blind_additive_scale = np.concatenate([
    np.broadcast_to(np.asarray(getattr(m, "blind_additive_scale", 1.0), ...), (m.npoi,))
    for m in param_models if m.npoi > 0
])
```

The Fitter accepts a scalar or an `(npoi,)` array, which also means a **composite of composites** works: the propagated vector is itself a legal declaration. The contract docstring now says `float or array (npoi,)`.

Five tests in `test_blinding.py`, **all of which fail without the fix** — I checked by reverting just the propagation block:

```
FAILED test_composite_preserves_a_declared_scale          <- your 7.0 case
FAILED test_scale_is_per_poi_not_one_composite_value
FAILED test_composite_of_composites_keeps_the_vector
FAILED test_undeclared_submodels_get_one
FAILED test_a_poi_less_submodel_does_not_flip_the_form
```

Plus two on #170, which is where it actually bites hardest — that composite is built inside `rabbit_fit.py`, so a dropped declaration has no user-side remedy at all, and for a free POI no warning either. They cover the declared scale surviving, the bin scales keeping 1.0 over their own slice (a composite-wide scalar would have rescaled *them*), and #161's warm-start premise still holding with a non-default scale in play.

**[P3] also fixed** — `additive_models` now filters `m.npoi > 0` like `overridden` does, so a POU-only submodel neither flips the composite nor gets named as the reason. The comment says why only POI-carriers vote.

PR #161, opened 2026-09-11
saturated test: keep blinding armed, and warm-start from the main fit

lucalavezzo:
Two bugs in the `--computeSaturatedProjectionTests` path, found while using it on a blinded analysis.

## The saturated fit ran unblinded and wrote an unblinded POI

The composite fitter is built with `copy.deepcopy(fitter)` and then re-initialised with `init_fit_parms()`, which re-creates the blinding offset Variables at the composite size — and creates them at the **identity**, `offsets_poi = 1` with `offsets_theta = 0`. Nothing armed them again.

So every parameter the saturated fit reported, **including the original model's POI**, was written to `results[...]["saturated_fit"]["parms"]` in the unblinded frame. Any analysis that asked for this test on data was silently unblinded in its own output file.

The fix arms the offsets on the composite fitter before `x` is touched; `set_blinding_offsets` holds the physical point fixed while it does. The blinding state is plumbed into `save_hists` as an argument rather than read from an enclosing scope, since `blinded_fits` is not in that function's scope.

## And the composite fit cold-started from the model anchor

`xdefaultassign()` puts it at `x0default`. That is wasteful, and for a multimodal likelihood it is also wrong: if the cold saturated fit stops above the main fit's own NLL, the statistic

```
q = 2 * (NLL_main - NLL_sat)
```

comes out **negative**, which is not a deviance.

This is not hypothetical. On a 780-bin analysis fit we measured at least **five distinct positive-definite minima** whose NLL spans 14 units, each stable under a warm restart from its own postfit (parameters move by < 1e-3 σ).

The fix warm-starts the composite at the main fit's converged point with every bin scale left at its default of 1, where the composite loss **equals** the main loss by construction — so `q ≥ 0` is guaranteed, and the statistic reads as "what do these free bin scales buy from *here*". It reuses the block permutation already applied to `x0` immediately above:

```
main      [poi_o | pou_o | theta]
composite [poi_o | poi_sat | pou_o | theta]
```

`x` is the internal (blinded) coordinate on both sides and the per-name offsets are deterministic, so the entries copy verbatim with no frame conversion.

## Verification

Full suite green (12 files), lint clean. The blinding bug is generic — it does not depend on the param model — so it reproduces with a plain `Mu` model.

Independent of #159, though they touch adjacent code.


bin/rabbit_fit.py
```diff
@@ -448,6 +448,51 @@ def save_hists(args, mappings, fitter, ws, prefit=True, profile=False):
...
+                # RE-ARM BLINDING. init_fit_parms() above re-created the offset
+                # Variables at the composite size, which creates them at the
+                # IDENTITY -- offsets_poi = 1, offsets_poi_add = 0 -- and
+                # nothing armed them again. So the saturated fit ran in an
+                # UNBLINDED frame and wrote an unblinded POI into
+                # results[...]["saturated_fit"]["parms"], silently unblinding
+                # any analysis that asked for this test. Arm before touching x:
+                # set_blinding_offsets holds the physical point fixed.
+                if fitter_saturated.do_blinding:
+                    fitter_saturated.set_blinding_offsets(blind=blind)
```
davidwalter2, line 461: This half is right and I would land it — confirmed the leak reproduces with a plain `Mu` model, offsets at identity on the composite while the main fitter's is armed.

One thing on the comment above it, though: *"Arm before touching x: `set_blinding_offsets` holds the physical point fixed"* describes #159's behaviour, not today's. On current `main`, `set_blinding_offsets` only assigns the offset Variables — it holds nothing fixed, which is exactly the bug #159 is fixing. Since this PR is stated to be independent of #159, the comment is promising a property the code does not yet have.

It matters here beyond pedantry: because arming does *not* currently compensate `x`, arming after `xdefaultassign()` is what moves the saturated scales off 1 — see the warm-start comment below.

davidwalter2 (commented):
Reviewed by reproducing both halves rather than reading. **The first fix is real and important; the second is broken by the first, in the same commit.**

## The unblinding leak is genuine

Confirmed directly — after `copy.deepcopy(fitter)` + `init_fit_parms()`:

```
composite offsets_poi all identity? True     <- unblinded frame
main fitter's own offset         : 1.621     <- armed
```

So the saturated fit really does run unblinded and write the original model's POI in the clear into `results[...]["saturated_fit"]["parms"]`. Any analysis asking for this test on data was silently unblinding itself in its own output file. Good catch, and worth landing on its own.

## The warm start does not survive it

`init_blinding_values` blinds **every** POI of the composite — which now includes the saturated bin scales — and the warm start assigns only `poi_o`, `pou_o` and `theta`, never the saturated slots. So they come out as `x = 1` times their raw blinding offset. Measured on a 20-bin toy with a plain `Mu` model:

```
bin scales after the warm start (the PR claims every one is 1):
   first 6: [0.1734  0.7418  0.9050  0.0071  47.57  52.93]
   min=3.9e-08   max=1.05e+05    all == 1?  False
```

and the consequence the argument rests on:

```
main fit loss (the target)      =    276.4278
this PR (warm + blinding armed) =  37990.4261   <- claimed EQUAL by construction
main   (cold, offsets identity) =     20.1857
```

So `q = 2*(NLL_main - NLL_sat)` starts around **-75,400**. The PR sets out to guarantee `q >= 0` and instead guarantees a large negative start on exactly the blinded fits it is about — and it is a regression against today's behaviour, where the cold start at least has the scales at 1. Details inline.

Either half is fine alone. It is arming the offsets that breaks the warm start.

## The cleaner resolution, and why I would not just special-case blinding

The saturated bin scales are auxiliary per-bin scales, not a measurement — there is nothing about them to hide. They should not be POIs in the first place. As POUs they are unblinded **by construction**: `get_model_nui()` applies only the frozen-params mask, where `get_poi()` applies the squaring and the offsets.

That also fixes something broken today. `CompositeParamModel` rejects submodels that disagree on `allowNegativeParam`, and `SaturatedProjectModel` hardcodes `False`:

```
Mu.allowNegativeParam=True  Saturated.allowNegativeParam=False
composite: RAISES -> submodels with POIs disagree on allowNegativeParam
```

so `--computeSaturatedProjectionTests` is currently unusable with any `allowNegativeParam=True` model — including the additive α_s blinding in #159, which requires it. The `poi_flags` check is guarded by `if m.npoi > 0`, so a saturated model with `npoi = 0` drops out of the unification entirely.

And it simplifies this PR: the composite layout becomes `[poi_o | pou_o | pou_sat | theta]`, so the original block is contiguous at `[0 : orig.nparams]` and the warm start is one slice copy instead of three.

Filed as #162, since it is a change to `SaturatedProjectModel` rather than to this file.

**So my suggestion for this PR**: keep the re-arm, drop or defer the warm start, and let #162 make the warm start correct afterwards. Landing both together ships a `q >= 0` guarantee that does not hold.

## Tests

Neither fix has one, and `111 passed` says nothing about either — the only `saturated` hit in `tests/` is a docstring in `test_external_nll_constant.py`, and nothing exercises `save_hists` or a blinded fitter. A silent-unblinding bug is the kind that most wants a regression test; the check I used above is about fifteen lines and would have caught the warm-start breakage too.

bin/rabbit_fit.py
```diff
@@ -448,6 +448,51 @@ def save_hists(args, mappings, fitter, ws, prefit=True, profile=False):
...
+                # construction, so q >= 0 is guaranteed and the statistic reads
+                # as "what do these free bin scales buy from HERE".
+                #
+                # The permutation is the one used for x0 just above:
+                #   main      [poi_o | pou_o | theta]
+                #   composite [poi_o | poi_sat | pou_o | theta]
+                # x is the internal (blinded) coordinate on both sides and the
+                # per-name offsets are deterministic, so the entries copy
+                # verbatim without a frame conversion.
+                x_main = fitter.x.numpy()
```
davidwalter2, line 483: This does not do what the comment says once blinding is armed, and the re-arm three lines above is what arms it.

`init_blinding_values` loops over **all** `param_model.npoi`, which on the composite includes the saturated bin scales, so each gets its own `exp(N(0,5))`. The three assignments below cover `poi_o`, `pou_o` and `theta` — the saturated slots `x[orig.npoi : composite.npoi]` are never touched, so they keep `x0default` and come out of `get_poi()` as `1 * offset`:

```
bin scales after the warm start (claimed: all 1):
   first 6: [0.1734  0.7418  0.9050  0.0071  47.57  52.93]
   min=3.9e-08   max=1.05e+05
```

So the composite loss at the start is **37990.43** against the main fit's **276.43**, i.e. `q` starts near **-75,400** — the failure this block exists to prevent, now unconditional on a blinded fit. For reference, today's cold start gives **20.19**, so this is also a regression, not just a missed improvement.

Two ways to fix it. The narrow one is to leave the saturated slots unblinded — set their entries of `_blinding_values_poi` to 1 before arming (more robust than routing names through `unblind`, which depends on the generated `saturated_<channel>_<label>` names).

The better one is #162: make them POUs, which are unblinded by construction and which also removes the `allowNegativeParam` conflict that currently blocks this test entirely for `allowNegativeParam=True` models. That would let this block become a single contiguous copy of `x_main[: orig.nparams]` plus theta.

Until one of those is in, I would drop the warm start from this PR rather than ship the guarantee — the re-arm above stands on its own.

bin/rabbit_fit.py
lucalavezzo: Right on both counts, and the second one is the more useful observation.

The comment is rewritten. It now says where the property comes from (the two commits below this one in the stack, not `main`) and — more to the point — **which parameters depend on it**: not the analysis POI, but the saturated bin scales, which are POIs of the composite and therefore blinded along with everything else.

That's the connection you drew in the comment below, and it turned out to be the whole story. See there.

bin/rabbit_fit.py
lucalavezzo: Your measurement is right and the diagnosis is exact. The resolution turned out to be neither of the two you listed, though — **it's the stack order.**

What the bin scales land on is `1 * exp(N(0, 5))`. That's the *multiplicative* form: they inherit `Mu`'s. And #159 reframes `x` on arming only for the **additive** slots — it says so explicitly, "multiplicative slots are deliberately left alone". So the breakage isn't that the warm start forgot a block; it's that arming moves the physical point for every multiplicatively blinded POI, and the saturated scales are simply the first place where that becomes load bearing.

#171 extends the reframing to the multiplicative slots. With it below this commit instead of beside it, the scales stay at their physical default of 1 and the warm start delivers exactly what it claims.

Reproduced on your toy — 20 bins, plain `Mu`, blinding armed, identical script both sides, only the stack differs:

```
#159 + #161           bin scales: min=3.9e-04  max=1.1e+04   all == 1?  False
                      main fit loss     =    16970.3846
                      warm + armed      = 16570807.7402
                      q                 =    -3.31e+07

#159 + #171 + #161    bin scales: all 1
                      main fit loss     =    12000.0000
                      warm + armed      =    12000.0000
                      q                 =          0
```

(Absolute losses differ from yours because my main fitter sits at the model default rather than a converged point — it's the equality that's the claim.)

So I've restacked to **#159 → #171 → #161 → #170** rather than dropping the warm start. Which also means **#162 isn't needed to unblock this** — it stays worth doing for the reasons you gave (they aren't parameters of interest, the layout collapses, `ndf` stops being read off `npoi`), but it's no longer load bearing for this PR, and I'd rather it land on its own merits than as a bug fix.

**On tests — you were right that `111 passed` said nothing.** `tests/test_saturated_blinding.py`, six checks, in the CI matrix:

- `test_composite_reinit_leaves_the_frame_unblinded` — that `init_fit_parms` really does leave the composite at the identity. This is the one that makes the re-arm load bearing rather than defensive, so if someone later "tidies it away" this fails first.
- `test_saturated_fit_does_not_run_unblinded` and `test_original_poi_is_written_blinded` — the leak itself.
- `test_bin_scales_are_one_after_the_warm_start`, `test_warm_start_sits_on_the_main_loss_while_blinded` — **both fail on #159 + #161 and pass here.** That's your measurement, as a regression guard.
- `test_the_same_holds_unblinded` — the whole thing with blinding off, so the loss equality can't be a cancellation between two offsets.

Every check asserts the frame is genuinely armed before it asserts anything else, so none of them can pass through an accidentally unblinded fitter — which was the failure mode in the first place.

One thing your review changed that isn't in this PR: #170's `test_warm_start_sits_exactly_on_the_main_loss` had `(allowNegativeParam=False, blind_additive=False, do_blinding=True)` — plain blinded `Mu`, i.e. your exact case — **excluded from its parametrization**, because it failed. It's back in now that it passes.

lucalavezzo:
## Restacked

| | PR | base |
|---|---|---|

davidwalter2:
Reviewed at `ff3f8c8`, against `stack/blinding-frame`.

Moving #171 below this rather than beside it is the right call — that resolves the blocker I raised, and it resolves it structurally rather than by patching around it. #162 being demoted from load-bearing to merely worth-doing is the correct conclusion too.

The re-arm fixes a real silent unblinding, and the reasoning for arming *before* touching `x` is sound. Two things I checked rather than took on trust:

- **Both `save_hists` call sites.** Only the `prefit=False` one passes `blind=blinded_fits[i]`; the `prefit=True` one doesn't, but `saturated_indices` requires `not prefit`, so that branch never runs there. No gap.
- **The warm-start index arithmetic.** `[0:npoi_o]`, `[composite.npoi : composite.npoi + npou_o]`, `[composite.nparams:]` mirrors the existing `x0` permutation directly above it, entry for entry. Correct.

One inherited issue and one nit.

---

### Inherited from #159: the "copy verbatim" claim doesn't hold once a scale is declared

> `x` is the internal (blinded) coordinate on both sides and the per-name offsets are deterministic, so the entries copy verbatim without a frame conversion.

The offsets are deterministic, but they are not the *same* on both sides. `CompositeParamModel` propagates `blind_additive` and drops `blind_additive_scale` (demonstrated on #159: `7.0 -> 1.0`), so the main fitter offsets by `scale * value` and the saturated fitter by `1.0 * value` for the identical parameter name. The copied `x` then lands at a physical point off by `(scale - 1) * N(0, 5)` in the POI's own fit units.

That is squarely this PR's concern, because the whole point of the warm start is that *"the composite loss EQUALs the main loss by construction, so q >= 0 is guaranteed"*. With mismatched frames the two losses differ and the guarantee goes away — the same failure mode as the uncompensated multiplicative frame, one layer down. It only bites once some model declares a non-default scale, which is latent today and exactly what the feature is for.

Fixing the composite in #159 fixes this; no change needed here beyond possibly softening the comment's claim to name the assumption (identical per-name offsets on both fitters).

### [P3] The pre-existing comment right below the new code contradicts it

```
# The composite re-init reordered and resized the parameter
# vector (one POI per projected bin, inserted ahead of the
# original model's block), so regularizers must be re-armed ...
```

"inserted ahead of the original model's block" is the opposite of both your new diagram (`composite [poi_o | poi_sat | pou_o | theta]`) and the `x0` code above, which puts the original POIs at `[0:npoi]`. It is pre-existing on `main`, so not yours — but the new warm start sits directly on top of it and depends on the opposite reading, so a reader reconciling the two will be misled at precisely the wrong moment. Worth correcting while you are here.

lucalavezzo:
Both fixed.

**The layout comment** — corrected to "appended AFTER the original model's POIs", which matches `CompositeParamModel`'s per-block concatenation (`[poi_orig | poi_sat | pou_orig | pou_sat]`) and the diagram above it. You're right that being pre-existing doesn't help here: the new warm start sits directly on top of it and depends on the opposite reading.

**The "copy verbatim" claim** now names its assumption rather than asserting the conclusion:

> That rests on the two fitters offsetting each shared parameter name IDENTICALLY: the draw is seeded by name, and `CompositeParamModel` propagates both the blinding form and its per-POI scale […] If a submodel declaration were ever dropped in that propagation the copied `x` would land at a different PHYSICAL point and the loss equality below would quietly stop holding.

And the underlying gap is closed in #159 — per-POI vector rather than a single value, since the scale is in each parameter's units. `test_scaled_blinding_still_opens_the_bin_scales_at_one` on #170 pins exactly the failure you described: with `scale=7` declared, the bin scales still open at 1 and the composite loss still equals the main loss. It fails without the propagation fix.

Good catch — that was latent today and would have surfaced the first time somebody used the knob for what it's for.

davidwalter2:
One more, separate from the findings above — a durability point about the comments themselves, and it applies across the stack rather than just here.

Several of the new comments are written relative to the stack rather than to the tree they will live in. The clearest case is in this PR:

> `# Arm BEFORE touching x, because set_blinding_offsets holds the`
> `# physical point fixed -- for BOTH offset forms. That is a`
> `# property of the two commits below this one in the stack, not`
> `# of main: on main it only assigns the offset Variables.`

Once this merges there is no "two commits below this one in the stack", and "not of main" inverts — it becomes a property *of* main. A reader a month from now gets a statement that is both unresolvable and false, attached to the one line where getting the ordering wrong silently unblinds a fit.

The information is worth keeping; it is the framing that expires. Something like "`set_blinding_offsets` holds the physical point fixed for both offset forms, so arming must precede any assignment to `x`" says the same thing and stays true after merge.

Worth one pass over the stack for this pattern before merging — I noticed it in a few places, and `rabbit/fitter.py`'s reference to "the 2026-09-09 alpha_s bug" is a borderline case worth a look too (it explains *why*, which earns its place, but dates the reasoning to an incident a future reader cannot look up).

For what it is worth this is exactly the note @davidwalter2 gave me on #154 — strip the development history, keep the reasoning — and it improved that PR, so I am passing it on rather than inventing a standard. The comments in this stack are genuinely above average; that is precisely why it is worth making them survive the merge. By contrast "Pinned by `tests/test_saturated_blinding.py`" is the durable form and reads just as well in a year.

lucalavezzo:
Taken, and applied across all three open PRs rather than just here. You're right that the one you quoted was the worst of them — not only unresolvable after merge but *inverted*, and sitting on the line where getting the order wrong silently unblinds a fit.

It now states the property instead of its provenance:

```
# Arm BEFORE touching x: set_blinding_offsets holds the
# PHYSICAL point fixed for both offset forms, so arming has to
# precede any assignment to x.
#
# The warm start below depends on that, and not via the
# analysis POI -- via the SATURATED bin scales, which are POIs
# of the composite and so are blinded along with everything
# else. Reframed, they stay at their physical default of 1;
# unreframed they land at 1 * exp(N(0, 5)) and the composite
# opens millions of units above the main fit. Pinned by
# tests/test_saturated_blinding.py.
```

I swept the whole stack for the pattern. Five more, all fixed, one commit per PR:

| PR | was | now |
|---|---|---|
| #171 | `fitter.py`: "That is the 2026-09-09 alpha_s bug" | what actually goes wrong — slow start for a signal strength, evaluation error for a restricted-domain POI |
| #171 | `test_blinding_multiplicative.py`: same reference | same |
| #171 | `test_blinding.py`: "This assertion **used to read** …" | the distinction it pins: the multiplicative Jacobian is intended and unchanged, the start point was never part of that bargain |
| #170 | "covered now that the reframing **below this in the stack** handles that form" | the bin scales are multiplicatively blinded POIs, so they open at 1 only because arming reframes that form |
| #170 | `"""The blocker: this construction **used to raise**."""` | which composition it is and which guard governs it |

On the `2026-09-09` ones specifically — you called it borderline and I think you were being generous. The sentence was carrying two things, the *reasoning* ("a restricted-domain POI gets handed a value it cannot evaluate") and the *citation*. Only the first was load bearing, and it reads better without the second.

One I deliberately left: `# whichever offsets are currently armed`. "Currently" there is runtime state at the moment of the call, not development history, so it stays true.

And one that isn't mine — `tests/test_preconditioner_spectral.py:206`, "which is the whole argument of this PR", already on `main` from #156. Same pattern, but out of scope for this stack; happy to fix it in a one-line follow-up if you want it.

No code changed in any of these commits.

lucalavezzo:

|---|--:|--:|
| #171 | 3 | 3 |
| #161 | 7 | 4 |
| #170 | 11 | 4 |

davidwalter2:

PR #163, opened 2026-09-12
Add hodor AI code review workflow (Kimi K3 via aigw endpoint)

bendavid:
## What

Adds `.github/workflows/hodor-review.yml`: agentic PR review with [hodor](https://github.com/mr-karan/hodor) @ commit `8a97b507` (v0.7.7), driving Kimi K3 (`Kimi-K3-FLATIRON`) on the CERN AI gateway (`https://aigw.cern.ch/v1`, OpenAI-compatible), posting the review as a PR review at `--reasoning-effort max`.

## Where it runs

`aigw.cern.ch` is firewalled to the CERN network — TCP from GitHub-hosted runners is blackholed (verified: ~136 s dropped-SYN connect timeouts). The job therefore runs on the repo's CERN-based **self-hosted runners** (`[self-hosted, linux, x64]`, same as `main.yml`), with self-sufficient tooling (sha256-pinned static `gh`/`rg`/`fd` fallback, no sudo assumptions).

## Trigger model

`pull_request_target`, deliberately: plain `pull_request` never receives secrets for fork PRs — almost all of this repo's PRs are forks — so the endpoint key would be missing exactly where review is needed. The workflow, patch, and `models.json` are always taken from the **base branch**, and who may run is governed by the repository's existing fork-contributor approval gate (`first_time_contributors`: approval required until a contributor's first merged commit, then automatic), the same gate CI inherits. Drafts are skipped; `workflow_dispatch` with a PR number remains for manual runs. The workflow starts auto-firing only once merged (pull_request_target resolves workflows from the base branch).

## Endpoint config: `.github/hodor/models.json`

hodor/pi have no custom-endpoint support out of the box (stock `LLM_BASE_URL` is a stale pre-rewrite artifact). `custom-models.patch` is two hunks: honor `HODOR_MODELS_JSON` (a pi-format models file) and accept its provider prefixes in `--model`. All endpoint specifics — base URL, 1M context, 131072 max tokens, compat quirks (`thinkingFormat: openai`, no `store`/`developer` role), and the thinking-level remap (pi defaults to `medium`, which the endpoint's chat template rejects; only `low`/`high`/`max` accepted) — live in the checked-in models.json, not in the patch.

## Secrets

`HODOR_LLM_API_KEY` as an **organization** secret shared with the repos running this workflow. Exposure model, stated precisely: this job runs *with* secrets where fork `pull_request` runs of `main.yml` never receive any, and the repo's approval gate (`first_time_contributors`) becomes automatic after a contributor's first merged commit — so the auto-running set is org members plus every fork author with ≥1 merged PR. That set is limited to established collaborators at time of writing (incl. all four recent fork contributors) and is accepted deliberately; worst case on exfiltration is aigw quota abuse (the job's `GITHUB_TOKEN` is task-scoped). The org secret's visibility must be limited to repos with at-least-as-strict fork-PR gating. Upgrade path if the trusted set widens: snapshot `.agents/skills` from the base tree and point hodor's skillPaths at the copy (hunk-sized hodor change, would go upstream). Optional repo variable `HODOR_LLM_MODEL` overrides the model id.

## Verification

- Patch applies cleanly to pristine `8a97b507` (`git apply --check`), `tsc --noEmit` clean.
- Live rehearsal on an upstream-branch PR: full run on the self-hosted runner → review posted as a PR review (19 turns, 23 tool calls, 2m30s).
- models.json path smoke-tested end-to-end locally (`--model cern/Kimi-K3-FLATIRON`) against the real endpoint.
- Failure-driven fixes baked in: `fetch-depth: 0` (diff pre-fetch), runner switch, PATH export inside the tool step.

### DIFF
--- a/.github/workflows/hodor-review.yml
+++ b/.github/workflows/hodor-review.yml
@@ -86,6 +86,7 @@ jobs:
           BIN="$RUNNER_TEMP/bin"
           mkdir -p "$BIN"
           echo "$BIN" >> "$GITHUB_PATH"
+          export PATH="$BIN:$PATH"
           fetch() { # url sha256 -> extracts tarball into $RUNNER_TEMP
             tmp="$(mktemp)"
             curl -fsSL "$1" -o "$tmp"
--- a/.github/workflows/hodor-review.yml
+++ b/.github/workflows/hodor-review.yml
@@ -58,6 +58,11 @@ jobs:
     steps:
       - name: Check out this repository (patch source)
         uses: actions/checkout@v4
+        with:
+          # hodor reuses this checkout as its review workspace and diffs
+          # against origin/<base>; depth 1 leaves no origin/main ref and
+          # silently degrades reviews to command mode.
+          fetch-depth: 0
 
       - name: Preflight configuration
         run: |
--- a/.github/workflows/hodor-review.yml
+++ b/.github/workflows/hodor-review.yml
@@ -1,12 +1,18 @@
 name: Hodor AI review
 
 # Automated PR code review with hodor (https://github.com/mr-karan/hodor),
-# driven by a self-hosted OpenAI-compatible endpoint (CERN AI gateway).
+# driven by the CERN AI gateway (aigw.cern.ch) serving Kimi K3.
+#
+# aigw.cern.ch is firewalled to the CERN network, so this job MUST run on
+# the repository's CERN-based self-hosted runners; GitHub-hosted runners
+# cannot reach the endpoint (verified: TCP connect blackholed).
 #
 # hodor has no native custom-endpoint support: its OpenAI-compatible fallback
-# model is hardwired to openrouter.ai. We therefore clone hodor at a pinned
-# tag, apply .github/hodor/llm-base-url.patch (honors LLM_BASE_URL and pins
-# the client to plain OpenAI chat-completions semantics), and build it here.
+# model is hardwired to openrouter.ai (its documented LLM_BASE_URL is a stale
+# pre-rewrite artifact). We therefore clone hodor at a pinned tag, apply
+# .github/hodor/llm-base-url.patch (honors LLM_BASE_URL, pins plain OpenAI
+# chat-completions semantics, and remaps reasoning efforts medium/xhigh that
+# the endpoint's chat template rejects), and build it here.
 #
 # Required configuration (organization or repository level):
 #   secret   HODOR_LLM_API_KEY   API key for the endpoint. An *organization*
@@ -42,7 +48,8 @@ jobs:
       github.event_name == 'workflow_dispatch' ||
       (github.event.pull_request.head.repo.full_name == github.repository &&
        !github.event.pull_request.draft)
-    runs-on: ubuntu-latest
+    # CERN-network self-hosted runners only (see header comment).
+    runs-on: [self-hosted, linux, x64]
     timeout-minutes: 30
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/hodor-review.yml
```diff
@@ -0,0 +1,144 @@
...
+  group: hodor-review-${{ github.event.pull_request.number || inputs.pr_number }}
+  cancel-in-progress: true
+
+jobs:
+  review:
+    # Fork PRs on the pull_request trigger get no secrets; skip them.
+    # Skip drafts; reviews start when the PR is marked ready.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      (github.event.pull_request.head.repo.full_name == github.repository &&
```
davidwalter2: This is the line that skips 98% of the repo's PRs — 58 of the last 59 came from forks, and this PR is one of them, so hodor has reported `skipped` on every run of its own branch while CI succeeded.

Replacing the `pull_request` trigger with `pull_request_target` and dropping this clause gets the behaviour I think is wanted: the job runs for fork PRs, receives secrets (which `pull_request` never does, whatever the author's permissions), and the repository's existing approval gate decides who can trigger it — the same gate that already governs `main.yml` on these forks. The draft check is worth keeping.

Please keep the comment explaining *why* the trigger is `pull_request_target`, since the difference is not obvious and the naive fix is to delete the guard and leave `pull_request`, which fails at the preflight instead.

.github/workflows/hodor-review.yml
```diff
@@ -0,0 +1,144 @@
...
+        with:
+          path: ~/.bun/install/cache
+          key: hodor-${{ env.HODOR_VERSION }}-${{ hashFiles('.github/hodor/llm-base-url.patch') }}
+          restore-keys: hodor-
+
+      - name: Build patched hodor
+        run: |
+          set -e
+          HODOR_SRC="$RUNNER_TEMP/hodor-src"
+          git clone --depth 1 --branch "$HODOR_VERSION" https://github.com/mr-karan/hodor.git "$HODOR_SRC"
```
davidwalter2: Two things on this build step.

**Pin the commit, not the tag.** `--branch "$HODOR_VERSION"` resolves whatever `v0.7.7` points at today; a retag silently changes what is built and what the patch is applied to. The cache key hashes the patch and the version string but not the resolved commit, so a moved tag would also be served a stale cache. `git rev-parse HEAD` verified against a recorded SHA, or cloning the SHA directly, makes it reproducible.

**`--model "openrouter/$HODOR_MODEL"` (line 141) reads like a mistake.** It works only because the patch redirects the openrouter provider at `LLM_BASE_URL`, so the prefix is now a provider-selector rather than a statement about where the request goes. One line of comment at that flag would save the next person the trip through the patch to work out that CERN's gateway is not being asked to impersonate OpenRouter.

davidwalter2 (commented):
Useful thing to have, and the patch work behind it is clearly the hard part — tracking down that the documented `LLM_BASE_URL` is a stale artifact, and that the endpoint's chat template rejects the default thinking level, is not fun. My concern is not the plumbing but who it ends up reviewing.

## As written it would skip almost every PR this repo gets

Last 59 PRs by head repo:

| head repo | PRs |
|---|---|
| `davidwalter2/rabbit` | 34 |
| `lucalavezzo/rabbit` | 16 |
| `bendavid/rabbit` | 3 |
| `havynanc` / `cippy` / `mseidel42` / `pmlugato` / `reimersa` | 1 each |
| **`WMass/rabbit`** | **1** |

**58 of 59 came from forks**, so the `head.repo.full_name == github.repository` guard skips 98% of them. This PR demonstrates it on itself — from `bendavid/rabbit`, so every run so far reads:

```
Hodor AI review  event=pull_request  head=bendavid/rabbit  skipped
CI               event=pull_request  head=bendavid/rabbit  success
```

The workflow cannot review the PR that introduces it, while CI reviews it fine. `workflow_dispatch` does work — hodor runs `gh pr checkout` (`workspace.ts:204`), which fetches fork PRs — but it is manual, per-PR, and never re-fires on a push, so the incremental-review behaviour the description sells is unreachable in practice.

## It should follow CI's gating, not invent its own

`main.yml` already runs on these fork PRs, on the same five self-hosted runners, and trusted-contributor gating is handled at the repository level rather than in YAML. Hodor should inherit exactly that, so the same people who get CI get a review.

The obstacle is that a plain `pull_request` trigger will not do it: GitHub withholds secrets from fork PRs regardless of the author's permissions, so dropping the guard would make the job run and then die at the preflight on a missing `HODOR_LLM_API_KEY`. `pull_request_target` runs in the base-repo context, does receive secrets, and leaves the existing approval gate in charge of who triggers it — which is the behaviour asked for here.

## What that means for the key, and one thing the patch should do

Under `pull_request_target` the fork's tree still ends up in the workspace, because hodor has to check the head out to review it. Two properties of hodor make that worth handling explicitly rather than leaving implicit:

- the agent's tool allowlist is `["read", "bash", "grep", "find", "ls", "submit_review"]` (`agent.ts:603-611`) — `bash`, with no approval gating;
- `skillPaths = [join(workspacePath, ".agents", "skills")]` (`agent.ts:526`), so hodor loads **skills from the workspace**, i.e. from the checked-out PR. A PR can add `.agents/skills/*` and have it read as agent instructions — not injection buried in a diff comment, but a documented extension point fed untrusted input.

Since the workflow already carries a patch, the cheap fix is to force `skillPaths` empty in it: that closes the most direct vector for nothing, and the review does not need workspace skills. And `HODOR_LLM_API_KEY` should be a repository secret that can be rotated without touching other repos, rather than the shared organization one — an agent with `bash` on a CERN-network runner is a more exposed place for an org-wide credential than CI is.

## The description no longer matches the workflow

> Model: `Kimi-K3-FLATIRON` on `https://aigw.cern.ch/v1` (OpenAI-compatible; verified reachable from GitHub-hosted runners).

against the file header:

> GitHub-hosted runners cannot reach the endpoint (verified: TCP connect blackholed).

The self-hosted move came in a later commit, so the description is stale — along with the `~1.5-2 min uncached / ~30 s warm` overhead figures and the "GitHub withholds secrets there anyway" reasoning, which is the thing that now needs a different answer. Worth updating, since the description is what the next maintainer reads.

Two smaller ones inline.

davidwalter2:
Thanks — and the `models.json` mechanism is a better answer than what I suggested. Moving the base URL, context window, token limits, compat quirks and the thinking-level remap into checked-in config shrinks the patch to provider-prefix support, and it retires my `--model "openrouter/..."` comment rather than papering over it: there is a real `cern` provider now. Commit-pinning the build addresses the other inline point.

**The org secret I'd leave as you have it.** I checked the permissions behind your argument and it holds for the traffic that matters: `davidwalter2` (admin), `lucalavezzo` (write), `bendavid` (admin) and `reimersa` (admin) account for 54 of the last 59 PRs, and all four could change this workflow or its secrets directly anyway. No gate is needed for them and none applies. Your call, stated reasoning, defensible.

**One narrow point where the parity argument does not hold.** The header says PR-workspace skills are "trusted under" the same gate as `main.yml`. But `main.yml` runs on `pull_request`, which receives **no secrets** for fork PRs, while this job runs on `pull_request_target`, which does — that difference is the whole reason the trigger had to change, so the two are not equivalent in exposure. And as your own comment notes, `first_time_contributors` stops applying after a contributor's first merged commit. The remaining read-access contributors (`havynanc`, `cippy`, `mseidel42`, `pmlugato` — 5 PRs between them) would therefore run automatically, with the org key and `bash`, on CERN-network runners, on content nobody has approved.

**A fix that keeps the feature rather than removing it.** Your reason for keeping skills is that repo review guidelines can live in `.agents/skills` — and those are base-repo content. Under `pull_request_target` the workspace *starts* as the base checkout and only becomes PR content when hodor runs `gh pr checkout`. So copying `.agents/skills` out of the base tree before that step and pointing `skillPaths` at the copy preserves the guidelines exactly as intended while removing PR control over the reviewer's own instructions. Cheaper than the patch hunk you reverted, and it does not cost you the feature.

Minor, since it will confuse someone: hodor will keep reporting `skipped` on this PR until it merges, because `pull_request_target` reads the workflow from the base branch. Not worth chasing — just not evidence that the gating is still wrong.

bendavid:
Per the latest round — org secret: agreed, staying as-is (as you note, the four heaviest contributors hold admin/write and could change the workflow or secrets directly anyway).

On the parity point: the mechanics you describe are correct, and the claim in the header was imprecise — this job runs **with** secrets where fork `pull_request` runs of `main.yml` never receive any, and `first_time_contributors` becomes automatic after a contributor's first merged commit. Both the header comment (7724ef0) and the Secrets section above now state it that way, including the difference.

On acceptance rather than hardening for the residual set: the auto-running group reduces to org members plus `havynanc`, `cippy`, `mseidel42`, `pmlugato` (and `reimersa`, already admin) — five established collaborators with 5 PRs between them. They are trusted in this context; worst case is exfiltration of an aigw-scoped LLM key (quota abuse) plus a read-oriented agent with bash on a CI runner — meaningfully narrower than the arbitrary PR `python` execution CI already performs for each of them, sans key. This is a consciously accepted exposure rather than an overlooked one, and it is now written down in the workflow header so the next maintainer sees the reasoning rather than the verdict.

The skills-snapshot fix is genuinely good — it removes PR control over the reviewer's instructions while keeping repo-authored guidelines working — but I'm declining it for now to keep the patch at two hunks; it's recorded in the header and description as the agreed upgrade path if the trusted set ever widens (and it belongs in the upstream PR, mr-karan/hodor#25, rather than a carried hunk).

And yes — `skipped` on this PR until merge is the expected base-branch resolution of `pull_request_target`; the live config was rehearsed end-to-end on #165 (closed).

PR #165, opened 2026-09-13
[workflow rehearsal 2] Hodor AI review — pull_request variant

bendavid:
**Rehearsal PR — do not merge.** Identical to #163 except the trigger is `pull_request` so that this PR can run the review workflow on its own branch pre-merge (`pull_request_target` resolves workflows from the base branch, which has no workflow yet). Same-repo branch, so secrets are in scope and the job should execute: CERN self-hosted runner → aigw → models.json (`cern` provider) → SHA-pinned hodor build → `--reasoning-effort max` review posted here.

PR #166, opened 2026-09-13
hodor workflow: no repo checkout; fork carries patch + model config

bendavid:
## Problem (observed on #154, #159, #161)

Under `pull_request_target`, `actions/checkout` provides the **base branch**; hodor's CI-detection path reuses an existing GHA checkout as-is and never fetches the PR head. Since #163 merged, review runs effectively reviewed `main`-vs-`main` (`Embedding diff in prompt (0 bytes)`; state keyed to base tip `4283ef62`; 'identical-HEAD reuse' on force pushes, so nothing new got reviewed or posted). #154's real review survived only because the agent recovered by diffing `remotes/pull/154/merge` by hand.

## Fix (workflow-only, final shape after review feedback)

The insight that simplified everything: the workspace existed only because we created it — with **no checkout of this repository at all** (hodor's README flow), hodor detects the absence, clones the repo itself, and `gh pr checkout` resolves fork heads with the job `GITHUB_TOKEN`. Review state then always keys on the real PR commit, base and fork PRs alike, by construction — the whole two-stage-checkout / trusted-copy machinery is unnecessary.

Support config moves out of this repo into the **WMass/hodor fork**, branch `cern-ci`, pinned by SHA (`8bb33632`): upstream v0.7.7 + the `HODOR_MODELS_JSON` patch (also filed upstream as mr-karan/hodor#25) + `models.cern.json` (aigw endpoint, 1M context, token limits, thinking-level remap). hodor runs from source via bun — no patch apply, no build step. `.github/hodor/` is deleted from this repo.

## Kept, deliberately

- `pull_request_target` + repo approval gating (fork PRs get secrets; `pull_request` never provides them) — workflow file always comes from base.
- sha256-pinned static `gh`/`rg`/`fd` bootstrap (runner image has no sudo/guarantees), bun via setup-bun with cache keyed on the pinned hodor commit, `--reasoning-effort max`, per-PR concurrency.
- `actions/cache@v5` (aligned with #167; its hodor checkout bump is moot after this rewrite — no checkout step remains).
- Header comment documents the trust model exactly (trusted config = pinned fork SHA; workspace = PR tree; trusted-contributor set; org-secret visibility constraint).

## Update: rehearsal + one more fix found by it (b7829d1)

Rehearsal (PR #169, closed) against a same-repo branch carrying this workflow (trigger flipped to `pull_request` there, since `pull_request_target` resolves the workflow from the DEFAULT branch — verified on #168, so a base-branch rehearsal cannot see unmerged workflows) exposed a second-order instance of the empty-diff failure: on these shared persistent runners, `GITHUB_WORKSPACE` holds a **stale checkout of this repo** from other workflows, and hodor's CI detection reuses it as the review workspace without fetching the PR head (observed: 0-byte embedded diff, state keyed to stale commit). Fixed in `b7829d1` by pointing `GITHUB_WORKSPACE` at a fresh empty dir for the hodor process, so detection declines and hodor clones the real PR head.

Post-fix run (34827118614): fresh temp workspace, clone of WMass/rabbit, embedded 13,122-byte diff (4 files), state keyed to the true head `de4ad1c`, incremental mode, 2 turns, review posted cleanly.

## Verification

- Previous evidence chain stands (see earlier comments): live rehearsal on #165 posted a full review via the aigw endpoint; the base-checkout failure mode was confirmed in the logs of runs 34790058137 / 34789726405 / 34789738658 / 34788150485.
- New shape dry-run from a bare directory against **this PR's own fork head**: hodor cloned WMass/rabbit, embedded the real 5372-byte diff, ran the incremental review path with correct `headSha` (`d3b40bb`), completed against the endpoint in 16 s.

After merge, re-review #154 (and #159/#161) via Actions → Hodor AI review → Run workflow (PR number) — their comment-side state keys on the base tip, so a fresh full review results.


bendavid:
Reconciled against #167 (davidwalter2's Node-24 bumps also touch this file): the v5 bumps are absorbed here — both checkout steps now use `actions/checkout@v5` and the cache step `actions/cache@v5` (d3b40bb). Verified content-exact: this file now equals 167's version plus the merge-ref workspace hunk, nothing else. Merge order is free; whichever lands second rebases away an already-applied hunk.

PR #167, opened 2026-09-14
CI: derive PYTHONPATH per job, serialise runs, and move off Node 20

davidwalter2:
## What

Three changes to the workflows, the first two addressing intermittent CI failures on PRs whose code was fine.

## The runners are not isolated from each other

They are five processes on one host: `/home/c/cmsmwbot/CombineTF2CI/actions-runner-{1..5}`. A successful run uses all five and shares one `/tmp/cmsmwbot/<uuid>`, so cross-job file sharing through `/tmp` works — but nothing separates one *run* from another.

**`setenv` exported one runner's workspace path to every job.** It set

```yaml
echo "PYTHONPATH=$GITHUB_WORKSPACE:$PYTHONPATH" >> $GITHUB_ENV
echo "PATH=$GITHUB_WORKSPACE/bin:$PATH" >> $GITHUB_ENV
```

and exported both as job outputs, which all twelve downstream jobs imported. `$GITHUB_WORKSPACE` is specific to the runner that ran `setenv`, so every job imported `rabbit` from **that one runner's checkout**, wherever it actually ran. Fine while that tree holds the commit under test; not fine once a second run checks out a different branch into it, at which point a job runs one commit's tests against another commit's modules.

That is what we saw. The same commit, code unchanged:

| run | failing test | error |
|---|---|---|
| 1 | `test_global_asym_impacts` | `Dimensions must be equal, but are 60 and 70` |
| 2 | `test_external_term` | `tf___compute_nll() missing 1 required positional argument: 'profile'` |
| 3 (quiet runners) | — | pass |

Both tests pass locally on the branch and on `main`, including their `__main__` script paths. And the first failure's traceback spans `actions-runner-4` **and** `actions-runner-5` inside a single job — the test coming from one workspace via `PATH`, the module from another via `PYTHONPATH`.

Each job now derives both from its own `$GITHUB_WORKSPACE`, immediately after its own checkout. `setenv` keeps exporting `RABBIT_OUTDIR`, `WEB_DIR` and `PLOT_DIR`, which are genuinely shared and genuinely want one value per run.

## Concurrency group

Overlapping runs on one ref also compete for `$RABBIT_OUTDIR`, and `copy-clean` removes it with `if: always()` — so a finishing run can delete a running one's inputs. `cancel-in-progress` also stops superseded runs occupying runners, which matters more now that the hodor review holds one for ~13 minutes per PR.

The `uuidgen` in `RABBIT_OUTDIR` is deliberately **kept**: it is what stops two runs sharing that directory in the first place. Replacing it with something stable per PR would reintroduce exactly the collision `copy-clean` makes destructive.

## Node 20 deprecation

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/cache@v4, actions/checkout@v4

Three stragglers, bumped to `@v5` (both on `node24`): `checkout` in `publish.yml` and `hodor-review.yml`, `cache` in `hodor-review.yml`. `main.yml` was already on `checkout@v5` in all thirteen other places, so this just finishes the job. No `@v4` uses remain.

## Checks

All three workflows parse; `setenv`'s outputs are down to the three shared values; no job still consumes a propagated `PYTHONPATH`; and every job that checks out sets its paths *after* doing so, verified programmatically rather than by eye.

The real test is this PR's own run, plus the absence of the intermittent failures under load.


davidwalter2:
Correct, and blocking — thanks. Fixed in `c3f7a4c`.

Confirmed it empirically: `regularization` failed in **7 s** on this branch, which is command-not-found rather than a test outcome, and it is indeed the only rabbit-running job without a checkout (`copy-clean`, the other one, runs only xrdcp/rm).

Worth adding the part that makes it more than an oversight: on `main` that job's steps are

```
['', 'lcurve scan', 'plot lcurve']
```

— no checkout at all, ever. `rabbit_fit.py` resolved purely through the `PATH` inherited from `setenv`, i.e. through **another runner's tree**. So the job was not merely unaffected by the cross-runner path leak this PR removes; it was *depending* on it. That is the same class of accidental coupling the PR is about, and it happened to be load-bearing in one place.

It now gets the same `actions/checkout@v5` plus per-job `setup python path` as the other twelve.

On how it got past me: the check I validated the previous commit with asserted that jobs which *have* a checkout also set their paths — so a job with neither sailed through the gap in the predicate. It now asserts the other direction as well: any job whose steps invoke `rabbit_fit.py` or `python` must have both, checkout first. That catches this case and would have caught it before I pushed.

bendavid:
Overlap note for `.github/workflows/hodor-review.yml`: #166 rewrites that file's checkout step into a two-stage checkout (base for trusted inputs + `refs/pull/N/merge` as the review workspace). To keep the PRs non-conflicting I've absorbed your v5 bumps there — both checkout steps `@v5`, cache `@v5`; content-verified identical to this PR apart from the workspace hunk. If #167 lands first, #166 stays clean; if #166 lands first, this PR's hodor-file hunk is already applied and drops out on rebase. Either order works; main.yml/publish.yml don't overlap at all.

bendavid:
Update to my earlier cross-reference: #166 has since been simplified much further — it now checks out nothing at all (hodor clones the PR itself, README flow) and sources its patched build from a pinned WMass/hodor fork branch. Consequence for you: `hodor-review.yml` loses its `actions/checkout` step entirely (your bump there becomes moot); the only remaining action in that file is `actions/cache@v5`, which #166 already carries at v5. If you rebase/merge after #166, the file's only delta for you should be zero.

PR #169, opened 2026-09-14
[workflow rehearsal 3b] exercise rewritten hodor workflow

bendavid:
**Rehearsal PR — do not merge.** Like #165: the trigger is flipped to `pull_request` on this branch only, because `pull_request_target` resolves the workflow file from the **default branch** (verified empirically on #168 — a PR base branch carrying the new workflow still ran main's old one), while `pull_request` resolves from the merge commit and therefore runs this branch's rewritten workflow. Everything else — WMass/hoder fork pin, no-checkout clone flow, model config, postings — is the exact #166 content. Review subject: comment-only +2-line change to `setup.sh` (e8cae96).


bendavid:
Rehearsal succeeded on the final shape: hodor created a fresh temp workspace (the GITHUB_WORKSPACE neutralization from b7829d1 worked — no stale-runner-tree reuse), cloned WMass/rabbit itself, embedded the real 13,122-byte diff (4 files), keyed state to the true head de4ad1c, ran the incremental path in 2 turns, and posted 'patch is correct / 0 findings'. The journey also exposed and fixed one real defect class inside this PR: stale GITHUB_WORKSPACE reuse on the shared persistent runners. Amended understanding recorded: pull_request_target resolves the workflow from the DEFAULT branch (verified #168). Closing; canonical PR is #166.

Issue #162, opened 2026-09-11
Saturated projection scales should be POUs, not POIs


davidwalter2:
`SaturatedProjectModel` declares its per-bin scales as **POIs** (`npoi = sum(nbins_channel)`, `npou = 0`). They are not parameters of interest — they are auxiliary free scales that absorb per-bin deviations so the saturated likelihood can be compared against the main fit. Declaring them as POUs instead fixes three things at once.

### 1. They stop being blinded, by construction

`Fitter.init_blinding_values` loops over `range(param_model.npoi)` and blinds every entry, so on the composite the saturated scales each pick up their own `exp(N(0, 5))`. `get_model_nui()` — the POU block — applies only the frozen-params mask, no offsets.

This is the concrete failure in #161: arming blinding on the composite fitter (which that PR correctly does, to stop the saturated fit writing an unblinded POI) leaves the bin scales at `1 * offset` rather than 1. Measured on a 20-bin toy with a plain `Mu` model:

```
bin scales: min=3.9e-08  max=1.05e+05     (claimed: all 1)
composite loss at the start = 37990.43
main fit loss               =   276.43
```

so `q = 2*(NLL_main - NLL_sat)` starts around -75,400. Special-casing the blinding would work, but the scales simply are not POIs, and the POU block already has exactly the semantics wanted.

### 2. It unblocks the test for `allowNegativeParam=True` models

`CompositeParamModel` refuses submodels that disagree on `allowNegativeParam`, and `SaturatedProjectModel` hardcodes `False`:

```
Mu.allowNegativeParam=True  Saturated.allowNegativeParam=False
composite: RAISES -> submodels with POIs disagree on allowNegativeParam
```

So `--computeSaturatedProjectionTests` is currently **unusable** with any `allowNegativeParam=True` model — including the additively-blinded α_s POI in #159, which requires it. The check is guarded by `if m.npoi > 0`, so a saturated model with `npoi = 0` drops out of the unification entirely and the conflict disappears.

### 3. It simplifies the composite layout

```
today:     [poi_o | poi_sat | pou_o | theta]
as POUs:   [poi_o | pou_o | pou_sat | theta]
```

The original model's block becomes contiguous at `[0 : orig.nparams]`, so the parameter copies in `save_hists` (both the existing `x0` permutation and the warm start proposed in #161) collapse to one slice plus theta.

### What has to move with it

[...]

PR #170, opened 2026-09-14
saturated test: let the saturated model own its positivity transform (unblocks --computeSaturatedProjectionTests for allowNegativeParam=True models)

lucalavezzo:
> ### ⚠️ Stacked PR — depends on #159 and #161
>
> This branch is `main` + **#159** + **#161** + **one commit of mine**. Until those two
> merge, GitHub will show all three commits and the diff will look much bigger than the
> change under review. The commits map exactly:
>
> | commit here | is |
> |---|---|
> | `cb3ed09` | **#159** (`blinding: additive offsets for POIs a model declares physical`) — identical tree change to its head `cf3e026` |
> | `5a6f76c` | **#161** (`saturated test: keep blinding armed, and warm-start from the main fit`) — byte-identical patch to its head `5366520` |
> | `2c99ea7` | **the change to review** |
>
> **Review only `2c99ea7`** — [that commit alone](https://github.com/lucalavezzo/rabbit/commit/2c99ea7).
>
> Why not target #161's branch instead, for a smaller diff? A PR into `WMass/rabbit` can
> only pick a base that exists in `WMass/rabbit`, and both #159 and #161 live on the fork.
> And the branch cannot simply drop them: **the new tests need #159** (they exercise
> `blind_additive` and `set_blinding_offsets`), so a branch without it would have red CI,
> and #161 is the commit this one directly follows — 47 lines in the same function. #159
> and #161 are siblings off `main`, independent of each other, so no single existing
> branch is a superset of both.

---

## The problem

`--computeSaturatedProjectionTests` builds
`CompositeParamModel([fitter.param_model, SaturatedProjectModel(...)])`
(`bin/rabbit_fit.py:410`). Both submodels carry POIs and they disagree on
`allowNegativeParam` — `SaturatedProjectModel` defaults to `False`, and any model whose
POI is a physical parameter fed into a calculation declares `True` — so
`CompositeParamModel.__init__` raises (`rabbit/param_models/param_model.py:163-170`).
The projected saturated test is therefore **unreachable** for those analyses.

This is not specific to any one downstream model. **`Mu --allowNegativeParam` +
`--computeSaturatedProjectionTests` raises too**, for no physics reason: `Mu` takes the
flag straight from `--allowNegativeParam` (`rabbit/parsing.py:565`).

The root cause is that `allowNegativeParam=False` makes the **Fitter** apply a squaring
transform to the *whole* POI block, to keep model-frame values positive. For a composite
whose submodels disagree, that is unsatisfiable — one needs the transform, the other must
not have it.

## The fix

Give `SaturatedProjectModel` the contract `AxisNormModel` already documents
(`param_model.py:611-615`): **`allowNegativeParam=True` means "the Fitter passes my slice
through raw, I square it myself in `compute()`"**. `bin/rabbit_fit.py` then builds it with
the analysis model's own flag, so the two submodels agree by construction and the
composite's guard is never reached.

The bin scales stay positive **either way** — that is the whole point. They multiply
yields, so a negative one sends the expected yield negative and the Poisson `log()` to
NaN. `allowNegativeParam` here selects *who* enforces positivity, not *whether* it is
enforced. The stored coordinate is `sqrt(scale)` in both branches.

`rabbit/fitter.py` is **not touched**.

## Why this design and not a per-parameter mask in the Fitter

The obvious alternative is to make the Fitter's transform elementwise: a boolean mask over
the POI block, `poi = where(mask, x, x**2)`. That looks more general — but **it does not
work here**.

`Fitter.get_poi()` applies the transform at `fitter.py:780` and the blinding offsets at
`fitter.py:788`, i.e. it **squares before it blinds**. A composite inherits
`blind_additive` from any submodel that declares it (`param_model.py:228-229`), so the
saturated POIs get an additive `N(0, 5)` offset as well. Under a per-parameter mask, a
squared saturated slot would then be

```
poi = x**2 + offset          # negative for any negative draw  ->  NaN likelihood
```

whereas self-squaring happens **after** the offset is applied:

```
poi = (x + offset)**2        # >= 0 for any offset
```

On the real 780-bin card this is not a corner case: the 39 saturated additive offsets run
from **0.074 to 13.407** in magnitude.

Making the mask approach safe therefore needs a **second** per-parameter mask for the
blinding *form*, plus per-parameter versions of the `blind_additive` guard
(`fitter.py:272`) and the priors-on-POIs guard (`fitter.py:351`) — five coupled edits in
the Fitter, each an opportunity to repeat the `CompositeParamModel` POU-layout bug fixed
in #143, which squared and blinded a submodel's POUs.

## Nothing moves for existing analyses

- `allowNegativeParam=False` (every current single-model analysis, and `Mu` by default):
  `_square_internally` is `False`, the Fitter squares the whole POI block exactly as
  before, `xparamdefault` is `sqrt(default)` as before, and `is_linear` evaluates to the
  same value (`nparams == 0 or False` → `nparams == 0`). **Byte-identical arithmetic.**
- The **reported frame is unchanged in both branches.** `parms` and its covariance are
  written as `fitter.x` (`bin/rabbit_fit.py:778`, and `:539` for the saturated block), and
  the stored coordinate is `sqrt(scale)` either way — so no `saturated_*` entry in any
  output file changes meaning.
- The composite `[POIs | POUs]` layout is untouched: `SaturatedProjectModel` has
  `npou = 0`, so the composite stays `[poi_analysis | poi_saturated | pou_analysis]`,
  which is what #161's warm-start permutation already assumes. Pinned by two tests.
- The guard against a genuinely unrepresentable mix **stays**: a POI model that cannot
  self-enforce (`Mu` with `allowNegativeParam=False`) composited with a permissive one
  still raises, rather than silently losing the positivity of its signal strengths. The
  message now names the offending submodels and the resolution.

## It is measurably stronger than the workaround it replaces

Downstream we had been bypassing the check by simply handing `SaturatedProjectModel` the
parent's flag, with no transform anywhere. Same inputs, same coordinates
`x = (-1.5, 2.0, -0.4, 0.0)`, the two side by side:

| | bypass (flags aligned, no transform) | this PR |
|---|---|---|
| minimum bin scale | **−1.545** → negative yield, NaN Poisson | **0.0** |

The physical space of the scales is `[0, ∞)` in both branches of this PR, exactly as it is
for the Fitter's own transform.

## Tests

New `tests/test_saturated_composite.py`, in the repo's convention (`python tests/<file>`),
added to the CI unit-test matrix. 19 checks:

- **frame equivalence** — `compute(x)` with self-squaring equals `compute(x**2)` with the
  Fitter's squaring, and `xparamdefault` is `sqrt(default)` in both branches (checked with
  a non-unit default via `expectSignal`, since `sqrt(1) == 1` would hide a mix-up);
- **positivity** — a negative stored coordinate gives a non-negative scale, and under
  additive blinding the scales survive a ±20 excursion in the coordinates with finite
  positive yields and a finite NLL;
- **layout** — `params[:npoi]` is exactly the POIs of both submodels and the analysis
  model's POU is in the POU block; flipping the sign of the analysis POI changes the
  result while flipping the sign of a saturated coordinate does not, i.e. the transform
  really is applied per submodel;
- **warm start** — `reduced_nll()` of the composite at the warm start equals the main
  fitter's to 1e-9 for `(allowNegative, blind_additive, do_blinding)` in
  `{(T,T,T), (T,F,F), (F,F,F)}`, and a short `minimize()` never ends above it (so the
  deviance cannot come out negative);
- **guards** — `Mu(False)` + permissive still raises; the legacy all-`False` composite is
  the old arithmetic; `is_linear` is `False` in both branches;
- **the `Mu --allowNegativeParam` case** that used to raise for no reason.

Full suite: **14 / 14 test files pass, 0 fail** (every `tests/test_*.py`, with
`test_restart`, `test_preconditioner` and `test_snapshot` run as `python -m tests.<name>`
because of their relative imports). Container `isort` / `black` / `flake8` clean on the
changed files and repo-wide; the repo's own pre-commit hook (autoflake, isort, black,
pylint 8.00/10) ran on the commit.

## End to end on a real analysis

780-bin `ptll × yll` Z card, a fully differentiable SCETlib param model (86 composite
parameters: 47 model + 39 bin scales), 3673 card nuisances,
`-m Project ch0 ptll --computeSaturatedProjectionTests -t -1`:

```
Saturated chi2:  ndof: 733   2*deltaNLL: 0.0   p-value: 100.0%     <- whole-fit
[paramPriors] applying Gaussian priors to 46/86 ParamModel params  <- the COMPOSITE
Saturated chi2:  ndof: 39    2*deltaNLL: 0.0   p-value: 100.0%     <- PROJECTED, ptll
```

`ndf = 39` is one free scale per `ptll` bin, as intended. `2*deltaNLL = 0.0` is the
*required* Asimov answer, not a coincidence: the data is the prediction, so 39 free bin
scales have nothing to buy. Read back from the output file, all 39 scales are
`1.00000000` with `max|r-1| = 0`, which is the property that makes
`q = 2*(NLL_main - NLL_sat)` a deviance at all (`q >= 0`, and `q = 0` iff the projected
marginal is described exactly). Any non-zero value would have meant the warm start or the
sqrt/square frame was wrong.

Asimov cannot cover the *blinded* leg (`bin/rabbit_fit.py:943` blinds only `f == 0` or an
observed-mode toy), so that was verified separately on the same card and the same 39-bin
mapping, without a minimisation: the 39 additive offsets are real (up to |13.4|), the
physical bin scales are still exactly 1 at the warm start (`8.9e-16`), the warm-start loss
equals the main loss, and a ±20 excursion leaves every scale `>= 0` with finite positive
yields.

## Known, pre-existing, not addressed here

With **multiplicative** POI blinding, `set_blinding_offsets` deliberately does not
compensate `x` (`fitter.py:718-722`), so the saturated bin scales open at `exp(offset)`
rather than at 1 and the warm start is not on the main fit's loss — meaning the projected
deviance can come out negative for a blinded analysis that does not declare
`blind_additive`. That is true on `main` today for `Mu` + saturated and is unchanged by
this PR. It is being fixed separately, because it changes the start point of every blinded
multiplicative fit and deserves its own review.


lucalavezzo:
## Restacked

| | PR | base |
|---|---|---|

lucalavezzo:
One change from the restack, worth flagging since it's a test that was quietly narrowed:

`test_warm_start_sits_exactly_on_the_main_loss` had `(allowNegativeParam=False, blind_additive=False, do_blinding=True)` — plain `Mu`, blinded multiplicatively, i.e. the **default** configuration — excluded from its parametrization. It was excluded because it failed: the multiplicative frame was uncompensated, so the saturated bin scales opened at `1 * exp(N(0, 5))` instead of 1. That is the same failure @davidwalter2 measured on #161.

With #171 now below this in the stack it passes, so the default configuration is back in the test that exists to guarantee `q >= 0`. **20 passed.**

davidwalter2:
Reviewed at `648ca54`, against `stack/saturated-warmstart`. **No findings** — this one looks ready.

"WHO enforces positivity, not whether it is enforced" is a good way to frame it, and moving the transform into the model is the right fix given the Fitter applies one transform to the whole POI block. I checked the things that could plausibly have gone wrong rather than just reading the rationale:

- **`is_linear`** — old `nparams == 0 or allowNegativeParam`, new `nparams == 0`. Identical for the default `False` branch, and correctly becomes non-linear in the `True` branch now that `compute()` squares. The comment claiming identity for the default is accurate.
- **`set_param_default(expectSignal, allowNegativeParam=False)`** — passing the literal rather than forwarding is right: stored `sqrt(default)` in both branches, so `False` (Fitter squares) and `True` (model squares) open at the same physical point.
- **The trailing `tf.ones([1])`** — appended *after* `tf.square(iscale)`, so the unused-bins scale stays a physical 1 in the self-squaring branch, matching the Fitter-squares branch where it is likewise untouched. Easy thing to get wrong by squaring the concatenated vector instead; you didn't.
- **Composition with additive blinding** — the docstring's claim holds. With `allowNegativeParam=True` the Fitter passes the slice raw, so `get_poi()` is `x + off_add` and `compute()` squares that, giving `(x + off_add)**2 >= 0` for any offset. And `xdefaultassign` subtracting the offset keeps `x + off_add = sqrt(1)`, so the bin scales still open at exactly 1 with blinding armed — which is what #161's warm start needs.
- **The error message's claim about `AxisNormModel`** — verified, it does `tf.square(ipoi)` with `allowNegativeParam = True`, so pointing users at it as the precedent is accurate. Naming the offending submodels in the exception is a nice touch.

The only caveat is inherited, not yours: `test_saturated_composite.py` never exercises `blind_additive_scale`, which is where the composite gap I flagged on #159 would have shown up. Worth a case here once that is settled, since this PR is what makes the additively-blinded composite reachable in the first place.

lucalavezzo:
Thanks for the detailed pass — and for checking the trailing `tf.ones([1])` and the `set_param_default` literal specifically, which were the two places I'd have been most worried about getting it wrong.

Took your caveat: `test_saturated_composite.py` now exercises `blind_additive_scale`. Two cases, both of which **fail without the composite propagation fix** (now in #159, as a per-POI vector):

- `test_declared_scale_survives_the_saturated_composite` — the declared scale survives, and the saturated bin scales, which declare nothing, keep `1.0` over their own slice. A composite-wide scalar would have rescaled them too, which is the concrete reason a single value is the wrong shape here.
- `test_scaled_blinding_still_opens_the_bin_scales_at_one` — with `scale=7` in play the bin scales still open at exactly 1 and the composite loss still equals the main loss, i.e. #161's warm-start premise survives a non-default scale.

This PR is where that gap bites hardest, as you say: the composite is built inside `rabbit_fit.py`, so a dropped declaration has no user-side remedy, and for a free POI the weak-blinding warning can't report it either.

22 passed here; **223 passed** on the full suite at the top of the stack.

---

On the bot's other P2 here (`tests/test_native_minimizer.py` not in the CI matrix and not runnable as `python tests/<file>`): **false positive against this stack** — the file is already on `main` and no diff in any of these four PRs touches it. The gap itself looks real, so probably worth its own issue rather than being smuggled in here.

lucalavezzo:
Rebased onto the updated stack (`main` now contains #159 and #172). Also did @davidwalter2's comment-durability pass here — two comments were written relative to the branch:

- the re-enabled `(False, False, True)` parametrize case explained itself as *"covered now that the reframing below this in the stack handles that form too"* → now states the property: the bin scales are multiplicatively blinded POIs, so they open at 1 only because arming reframes that form;
- `"""The blocker: this construction used to raise."""` → now says which composition it is and which guard governs it.

Comment-only, no code change. 22 passed here; **234 passed** on the full suite at the top of the stack.

Note this PR still targets `stack/saturated-warmstart`, so it is **not getting CI** — `main.yml` triggers on `pull_request: branches: [ main ]`. It'll get its first run when #161 lands and this retargets to `main`.

lucalavezzo:

|---|--:|--:|
| #171 | 3 | 3 |
| #161 | 7 | 4 |
| #170 | 11 | 4 |

davidwalter2:

PR #171, opened 2026-09-14
blinding: compensate the frame for multiplicative POI offsets and for blinded NOIs

lucalavezzo:
> ### ⚠️ Stacked PR — depends on #159
>
> This branch is `main` + **#159** + **two commits of mine**. Until #159 merges, GitHub
> will show all three and the diff will look bigger than the change under review. The
> commits map exactly:
>
> | commit here | is |
> |---|---|
> | `cb3ed09` | **#159** (`blinding: additive offsets for POIs a model declares physical`) — identical tree change to its head `cf3e026` |
> | `a50b00e` | **leg 1 to review** — multiplicative POI offsets |
> | `8ad8081` | **leg 2 to review** — blinded nuisances of interest |
>
> Based on #159 rather than `main` because `set_blinding_offsets` and `xdefaultassign` are
> exactly the two functions #159 rewrites: basing on `main` would guarantee a conflict in
> both, and the frame-shift machinery this generalises is #159's.

---

## What this is

**The multiplicative and NOI counterpart of the additive start-point fix #159 made.** Same
class of bug, in the two paths #159 did not cover.

`self.x` is initialised to `x0default`, i.e. in the **unblinded** frame. Blinding is a
change of variables, so arming the offsets moves the **physical** point unless `x` is
compensated. #159 compensated the additive POI slots and said so explicitly, leaving the
other two paths alone to avoid changing results:

> *So shift `x` by `off_old - off_new` on the additive slots, which keeps `x + off`
> invariant. […] **Multiplicative slots are deliberately left alone** — their physical
> start is preserved only when the declared default is 0, and changing that would alter
> results for every analysis using them.*

The two uncompensated paths were:

| path | what the fit opened at instead of the declared default |
|---|---|
| **multiplicative POI** (the default form; `Mu`) | `default · exp(N(0, 5))` |
| **theta / nuisance of interest** (`--poiAsNoi`, i.e. the mass) | `theta0default + N(0, 5)` |

#159's own docstring names this class of failure as the 2026-09-09 α_s bug — and it was
the *multiplicative* form that caused it, by handing SCETlib a value outside its domain.

## Leg 1 — multiplicative POI offsets (`a50b00e`)

**The compensation is not the offset ratio in general.** `get_poi` is affine in the model
frame (`fitter.py:775-790`):

```
poi = T(x) * mul + add        T = identity, or square when allowNegativeParam is False
```

so the coordinate reproducing the same `poi` under new offsets is
`T⁻¹((T(x)·mul_old + add_old − add_new) / mul_new)`:

| branch | frame-preserving update |
|---|---|
| `allowNegativeParam=True` | `x → x·(mul_old/mul_new) + (add_old−add_new)/mul_new` |
| `allowNegativeParam=False` — **the default, and what `Mu` uses** | `x → x·**sqrt**(mul_old/mul_new)` |

`add` is identically zero on the squared branch (`blind_additive` raises unless
`allowNegativeParam` is True, `fitter.py:272`), which is what lets the square root be taken
of the multiplicative ratio alone. Using the un-rooted ratio there — the obvious guess —
would leave the start off by `sqrt(offset)`, still orders of magnitude; a test reads the
factor off the coordinate and asserts it is `1/sqrt(mul)` and **not** `1/mul`.

## Leg 2 — blinded nuisances of interest (`8ad8081`)

**Derived from `get_theta`, which is where it has to come from** — the POI leg needs a
square root precisely because `get_poi` has a transform in front of it, and `get_theta`
does not:

```
theta_physical = theta_stored + add       no transform, and no multiplicative
                                          offset exists for nuisances at all
```

so holding `theta_physical` fixed is the plain additive shift

```
theta → theta + (add_old − add_new)
```

the same form the additive POI slots already use. **No square root, no ratio.** A test
reads the update off the coordinate and asserts it is a difference.

**The constraint term is unaffected, and that is checked rather than assumed.**
`_compute_lc` penalises `get_x() − self.x0`: both sides are in the model frame, and `x0`
is **not** touched here (a test pins that, as #159's own `test_x0_untouched_by_arming`
does for the POI block). Holding `theta_physical` fixed therefore leaves the penalty's
value *and* its minimum exactly where they were.
`test_constrained_minimum_is_not_moved` passes **before** this commit as well as after,
for a constrained and for a free NOI — which is the evidence that the bug only ever moved
the start point, never the answer.

**`xdefaultassign`'s guard becomes `do_blinding` alone**, dropping `and npoi`: an analysis
whose parameter of interest is a NOI can have `npoi = 0` and still need its theta block
reframed — that is exactly the `--poiAsNoi` shape, and the old guard skipped it entirely.

## Both legs: `xdefaultassign` too, not only `set_blinding_offsets`

The driver calls `defaultassign()` once per (pseudo)data set (`bin/rabbit_fit.py:1033`)
and from the second set onwards the offsets are **still armed** from the previous one
(`:1070`); the toy-throw path calls `xdefaultassign()` armed for the same reason
(`fitter.py:979`). Without the reframing there, the physical start of every set after the
first depends on whether the call happens to run armed, and `defaultassign()`'s trailing
disarm no longer lands on the declared default.

Both reframes go through one `_reframe_blinded_x`, which applies `_poi_reframe` to the POI
block and `_theta_reframe` to the nuisance block and deliberately leaves the ParamModel's
own `npou` nuisances alone — they are never blinded. Every path reduces to the identity
when the offsets do not change, so **arming twice is a no-op** on both `x` and `get_x()`;
pinned by tests on both legs, since that is the property #159's additive path already
relies on.

### How big is the error being removed

**POI leg** — the offset is `exp(N(0, 5))` per POI name, so for a squared-storage POI the
physical start is off by that factor. Over 400 POI names:

| | log₁₀(physical start / declared start) |
|---|---|
| percentiles [0, 2.5, 16, 50, 84, 97.5, 100] | −5.56, −4.33, −2.39, −0.41, +2.08, +5.05, +7.44 |
| off by more than 10× | **65.5 %** of POIs |
| off by more than 100× | **39.2 %** |
| off by more than 1000× | **18.8 %** |
| extremes in the sample | **×2.8 × 10⁷** and **×2.8 × 10⁻⁶** |

**NOI leg** — the offset is `N(0, 5)` in the nuisance's own units, which for a constrained
nuisance are prior widths. Over 400 blinded constrained NOIs:

| | \|physical start − declared start\|, in prior widths |
|---|---|
| median / 84 % / 97.5 % / max | **3.70 σ** / 7.62 σ / 11.55 σ / 17.93 σ |
| more than 1 σ off | **85.2 %** of NOIs |
| more than 2 σ off | **71.2 %** |
| more than 5 σ off | **35.0 %** |

Each also drags in `0.5·offset²` of **prefit constraint penalty** — a mean of **14.3 per
NOI**, i.e. pure noise in the prefit NLL of every blinded analysis.

### A converged fit does not move

`get_poi`, `get_theta`, the offsets and `x0` are all untouched — the diff only changes code
that *assigns* `x`. So the likelihood as a function of `x` is identical and only the
minimiser's starting point moves. Measured:

| | before | after |
|---|---|---|
| POI: physical POI at the start (declared 1.0) | 1.407 | **1.000** |
| POI: 2·NLL at the start | 879.44 | **0** |
| POI: physical POI postfit | 1 | **1** |
| NOI: physical θ at the start (declared 0) | **+4.95 σ** | **0.000000** |
| NOI: 2·NLL at the start | 28.46 | **0** |
| NOI: physical θ postfit | 0.000000 | **0.000000** |

The postfits agree; what changes is that the fit no longer has to climb back to where it
was told to start.

**What can move, honestly:** a fit that does *not* reach its minimum — one that stops on
`maxiter`, stalls, or lands in a different basin of a multimodal likelihood. For those the
result changes, in the new behaviour's favour, but it does change. Anything frozen or
published from a non-converged blinded fit is worth re-checking.

### Is the old behaviour load-bearing anywhere?

No, as far as I can establish:

1. **Tests** — exactly one assertion depended on it: `test_blinding.py`'s
   `test_multiplicative_path_unchanged` ended with `assert np.isclose(f.x[0], START)` under
   the comment *"x is NOT frame-shifted for a multiplicative model"*. Renamed to
   `test_multiplicative_path_keeps_its_arithmetic_but_is_reframed` and rewritten to assert
   what is actually invariant: the `get_poi = x · offset` arithmetic is unchanged, `x` *is*
   reframed, the physical start is preserved. Nothing depended on the theta behaviour.
2. **Call sites** — `set_blinding_offsets` is called at `fitter.py:885` (the disarm inside
   `defaultassign`) and `bin/rabbit_fit.py:1070`; none reads both frames off one `x`, which
   is the only way the uncompensated behaviour could be relied on.
3. **`x0` is untouched**, still pinned by `test_x0_untouched_by_arming` and now by a NOI
   equivalent.
4. **Snapshots and `--externalPostfit` are cross-version safe.** A stored `x` means exactly
   what it meant before, because `get_poi`/`get_theta` are unchanged; and `load_fitresult`
   runs inside `fit()` (`bin/rabbit_fit.py:565`), i.e. *after* the arming, and is not
   reframed afterwards. Reloading an old fitresult lands on the same physical point — which
   is also the clean way to reproduce an existing blinded fit bit-for-bit, without needing
   a behaviour flag.
5. **The documented multiplicative property is unchanged.** `d poi/dx = mul`, so the
   reported σ, the POI row of the covariance and the impacts are still divided by the random
   factor (only the *relative* uncertainty survives). This touches the start point, not that
   Jacobian; the `test_blinding.py` module docstring now says so explicitly.

## Default, not opt-in

Deliberately unconditional. The current behaviour is not a choice anyone made — a model
declares a start value and the fitter silently moves it — it cannot change a converged
result, and an opt-in flag would leave the bug as the default in the one area where a wrong
default is least visible: you cannot see a blinded start point in the output. It would also
re-split the invariant, leaving `set_blinding_offsets` frame-preserving for additive offsets
but only conditionally so for the other two, which is the asymmetry that produced this in
the first place. Reproducibility of an existing blinded fit is better served by
`--externalPostfit` on its fitresult (point 4 above) than by a behaviour flag.

## Tests

Two new files, in the repo's convention (`python tests/<file>`), both added to the CI
unit-test matrix. Run against this branch and against plain #159 with nothing else changed:

| | before | after |
|---|---|---|
| `tests/test_blinding_multiplicative.py` | **7 failed**, 7 passed | **14 passed** |
| `tests/test_blinding_noi.py` | **8 failed**, 10 passed | **18 passed** |

(The checks that already passed before are the idempotence, invertibility, still-blinded
and `--unblind` ones: those properties held trivially for code that did nothing. They are
in the files to stop a future change from breaking them.)

Between them the two files cover, for both legs: start-point invariance, idempotence,
invertibility, `defaultassign`- and `xdefaultassign`-while-armed, loss equality armed vs
disarmed, that the coordinate is still blinded, `--unblind`, and the algebra itself
(square-root vs ratio for the squared POI; difference vs ratio for theta). The NOI file
additionally covers `npoi = 0`, `x0` untouched, an ordinary non-NOI nuisance untouched, the
constraint penalty, and the constrained and free minimum.

Full suite: **15 / 15 test files pass, 0 fail** (every `tests/test_*.py`, with
`test_restart`, `test_preconditioner` and `test_snapshot` run as `python -m tests.<name>`
because of their relative imports). Container `isort` / `black` / `flake8` clean on the
changed files and repo-wide; the repo's own pre-commit hook (autoflake, isort, black,
pylint 8.40/10) ran on both commits.


lucalavezzo:
## Restacked

| | PR | base |
|---|---|---|

lucalavezzo:
Re the bot's P2 — correct, and fixed one level down.

`tests/test_blinding.py` was never in the `unit-tests` matrix at all, which is also why the eight tests it added in #159 were erroring under `pytest` without anything noticing. It's registered in **#159** now, alongside the fixture conversion, and this PR keeps its own two entries. All four blinding suites are in the matrix on this branch:

```
- test_blinding.py                 (#159)
- test_blinding_multiplicative.py  (here)
- test_blinding_noi.py             (here)
- test_saturated_blinding.py       (#161)
```

43 passed across the three blinding suites on this branch.

lucalavezzo:
Rebased onto the updated `stack/blinding-additive`; no changes of its own this round.

Both bot findings here are addressed elsewhere:

- **`blind_additive_scale` dropped by `CompositeParamModel`** — real, and fixed in **#159** as a per-POI vector rather than a single propagated value (the scale is in each parameter's own units, so `max`/`first`/"any one wins" would each silently rescale somebody). Five tests there, two more on #170.
- **`test_native_minimizer.py` not in the CI matrix** — **false positive against this stack.** That file is already on `main` and no diff in any of these four PRs touches it; the bot appears to be attributing inherited content to the delta. The gap it describes looks real on `main` though, so worth a separate issue.

**223 passed** on the full suite at the top of the stack.

davidwalter2:
Reviewed at `bb3cb10`, against `main` (`stack/blinding-additive` is now an ancestor of `main` since #159 landed, so the diff is identical either way). **No findings.**

This was the one piece of the stack nobody had read, and #161's correctness rests on it, so I checked the algebra rather than the prose.

**`_poi_reframe`.** Solving `T(x_new)*mul_new + add_new == T(x_old)*mul_old + add_old` gives `x_new = T⁻¹((T(x_old)*mul_old + add_old − add_new)/mul_new)`, and both branches are that:

- `allowNegativeParam=True`, `T = id` → `x*(mul_old/mul_new) + (add_old−add_new)/mul_new`. Matches.
- `allowNegativeParam=False`, `T = ²`, and `add ≡ 0` because `blind_additive` raises unless `allowNegativeParam` → collapses to `x*sqrt(mul_old/mul_new)`. Matches.

Worth noting the squared branch stays correct for a negative stored `x`: `(x·√r)² = x²·r` regardless of sign, so the physical value is right and the sign branch is preserved. Using `abs` would have been wrong in a subtler way.

**`_theta_reframe`.** `get_theta() = theta + add`, no transform and no multiplicative offset, so the plain shift is right.

**The two load-bearing docstring claims, both verified in code rather than taken on trust:**

- *"`_compute_lc` penalises `get_x() − self.x0`, i.e. the MODEL frame"* — confirmed, `lc = cw * 0.5 * tf.square(self.get_x() - self.x0)`. So holding `get_x()` fixed leaves both the value and the minimum of the constraint exactly where they were.
- *"the `npou` block is never blinded"* — confirmed, `get_model_nui()` applies only `frozen_params_mask`, no offset. Leaving that block untouched is correct.

**`xdefaultassign` gated on `do_blinding` alone rather than `npoi`** is a real catch — `--poiAsNoi` gives `npoi = 0` with a blinded theta block, which the old gate skipped entirely. And the comment correction is an improvement in its own right: `x0default` is the stored coordinate at the identity offsets, not a physical point, which is what makes the identity→armed reframe the right operation. `defaultassign()`'s trailing `set_blinding_offsets(False)` does reframe back, as the comment says.

**I mutation-tested your test suite** against the three errors the docstrings warn about, and it holds up:

| mutation | result |
|---|---|
| drop the `sqrt` in the squared branch | **4 failed** |
| `_theta_reframe` → no-op | **8 failed** |
| revert `xdefaultassign` to the `npoi` gate | **14 failed** |

The `sqrt` one is the case I'd have expected to slip through, since it is invisible at `mul ≈ 1` — it doesn't. All three new test files are in the CI matrix too, which given #173 is worth saying out loud.

Ready from my side. Retargeting to `main` is a no-op for the diff and is what will finally give this CI — note it has had *none* so far, since `main.yml` triggers on `pull_request: branches: [main]` and this targets a `stack/*` branch.

lucalavezzo:
Thank you — the mutation table is the part I'll steal. Dropping the `sqrt` was exactly the mutation I'd have bet slipped through, since at `mul ≈ 1` it's invisible; good to see 4 tests catch it rather than 0.

**Retargeted to `main`**, and you were right about the consequence: this PR had received **no CI at all** until now. `main.yml` is `pull_request: branches: [ main ]`, so every PR in the stack pointing at a `stack/*` base was silently untested — the checks section was simply empty rather than failing, which is the worst way for that to present. First run is going now.

That's a real cost of the stacked-base layout I set up this morning, and it's worth being explicit that it isn't fixed, only deferred: **#161 and #170 still have no CI**, because they still target `stack/*`. The plan is to drain one at a time — as each lands, the next retargets to `main` and gets its first real run before merging. So nothing merges untested, but nothing gets early warning either.

If you'd rather have it properly, adding `stack/**` to the trigger's branch list is one line. I haven't done it, because it's a change to shared CI config on a PR you've just called ready, and it belongs with #173 more than with the blinding work. Your call.

One correction to the comments since your review: I did a durability pass over this PR too, per your note on #161 — three comments here were written relative to the branch (two dating the reasoning to "the 2026-09-09 alpha_s bug", one explaining itself as "this assertion used to read …"). Comment-only, no code change, `16 passed` / `48` across the three blinding suites.

Full suite at the top of the stack: **234 passed**, plus `test_snapshot.py::test_sigterm_writes_a_snapshot_before_the_process_dies`, which is flaky rather than broken — it fails on unmodified `main` in a worktree and passed in an earlier run of this same tree.

lucalavezzo:

|---|--:|--:|
| #171 | 3 | 3 |
| #161 | 7 | 4 |
| #170 | 11 | 4 |

davidwalter2:

PR #172, opened 2026-09-14
ndfsat: charge a degree of freedom for free parameters, wherever they are declared

lucalavezzo:
> **Independent PR — based cleanly on `main`, no stack.** It needs neither #159 nor #161:
> the `prior_sigmas` machinery it reads and the `ndfsat` line it changes are both already on
> `main`, and the diff is one commit. (My #170 and #171 are stacked on #159; this one is
> not, deliberately — it should be mergeable on its own.)

---

## Lead with the evidence

A real 780-bin `ptll × yll` card, 47 param-model parameters of which **46 carry σ = 1
priors** (only `alphaS` is free), and **no** unconstrained card systematics. The saturated
statistic is unchanged; only its ndf moves, 733 → 779. Across five independent data fits:

| fit | sat χ² | p (ndf 733, as-is) | p (ndf 779, this PR) | linear χ² | p (ndf 780) |
|---|---|---|---|---|---|
| **ENVDATA** (the main postfit) | 754.56 | **28.3 %** | **72.9 %** | 752 | **75.8 %** |
| ENVDATAWALL | 752.24 | 30.3 % | 74.8 % | 748 | 78.9 % |
| ENVDATAMIRROR | 775.52 | 13.4 % | 52.8 % | 773 | 56.4 % |
| ENVDATAWARM | 749.50 | 32.8 % | 77.0 % | 747 | 79.7 % |
| ENVDATALAT | 763.22 | 21.3 % | 65.0 % | 760 | 68.9 % |

**Under this change the saturated and the linear statistic agree; rabbit's present rule
puts them ~45 percentage points apart.** They *should* agree here: the model is near-linear
on this card, and the two χ² **values** already agree to 2–4 out of ~770 in every row. It
was only the ndf conventions that disagreed. That coincidence is the strongest single
argument that the current formula is wrong rather than merely conventional.

It also makes eras comparable again. Cards where these parameters lived in the datacard
reported `ndfsat ≈ 772`; moving 41 constrained parameters into a param model cost 41
degrees of freedom for no change in physics, which is what made old and new p-values look
incomparable.

## The bug

```python
ndfsat = tf.size(nobs) - param_model.nparams - indata.nsystnoconstraint   # :750
```

Every **model** parameter is charged, but only the **unconstrained** card systematics. So
the formula distinguishes parameters by **where they are declared**, which is not a
statistical property. A `pdfEig23` model parameter with a σ = 1 Gaussian prior is
statistically identical to a `pdf23…` card nuisance with a σ = 1 Gaussian prior.

**Why zero is the right charge for a constrained parameter:** it adds one parameter *and*
one pseudo-measurement, so it costs **net zero** degrees of freedom. That is already
exactly why rabbit charges constrained card nuisances nothing. Constrained model parameters
are the same object and must be charged the same way.

## The fix

```python
ndfsat = nbins - fitter.nfreeparms
```

`nfreeparms` counts `cw == 0` over the whole parameter vector. `Fitter.cw` is the single
place that records which parameters are constrained across `[ParamModel params | systs]` —
declared `prior_sigmas` entries that are finite and `> 0` folded in by `init_fit_parms()`
on one side, `indata.constraintweights` on the other — and it is **the same vector
`_compute_lc` penalises**, so the count cannot drift from what the likelihood actually
does. Nothing is hardcoded and nothing is re-derived in the driver.

A companion `nfreeparms_breakdown` feeds one new log line so the decomposition is visible:

```
Saturated chi2:
    ndof: 779
      = 780 bins - 1 free of 47 ParamModel params - 0 free of 3673 systematics
```

## Backward compatibility — verified, and exact

**This is an exact no-op for any analysis that does not put priors on model parameters.**
Not approximately: if the model declares no `prior_sigmas`, `param_cw_np` stays all zeros,
so `nfreeparms` is identically `param_model.nparams + indata.nsystnoconstraint` — the
previous expression, term for term.

Verified end to end through the driver, not just by reading the code. A 20-bin card with 7
model parameters, 2 unconstrained and 2 constrained card systematics, run on both trees:

| priors the model declares | reported `ndof` before | after |
|---|---|---|
| 0 | 11 | **11** ← exact no-op |
| 1 | 11 | **12** |
| 6 | 11 | **17** |

Before, the reported ndof is 11 whatever the model declares. After, each declared prior
buys back exactly one degree of freedom and the prior-free case is untouched. That is what
makes this safe for mW and every other existing analysis.

## The linear χ² path: deliberately out of scope

`_chi2` uses `ndf = tf.size(res) - mapping.ndf_reduction`, i.e. `nbins` less any dof the
*mapping* removes, and subtracts nothing for fitted parameters. Consistency between the two
statistics is half the argument here, so this needs an explicit answer rather than silence.

**I left it alone, and I think that is right, but it is a judgement.** The reason is that it
is not the same statistic: `_chi2` computes `res^T (C_pred + C_data)^{-1} res` with the
**postfit** prediction covariance, so the fit's freedom is already absorbed — at least
partly — into `C_pred` rather than into the ndf. "Subtract the free parameters" there would
risk double-counting and needs its own derivation, which I have not done.

Two things make this a safe split rather than an inconsistency swept under the rug:

- on the card above it would be a **1-dof** change (780 → 779, p 75.8 % → 75.0 %), because
  only `alphaS` is free — so it does not touch the agreement that motivates this PR;
- the agreement in the table is *between* the fixed saturated ndf (779) and the linear ndf
  as it stands (780). If the linear path also needed −47, the two would disagree again
  after this change.

Worth its own issue/PR for analyses with many genuinely free model parameters (an `ABCD`
model, or `Mu` with many signals), where the difference is not one dof.

**Also noticed, also out of scope:** `impacts_parms` computes
`nstat = npoi + npou + indata.nsystnoconstraint` (`fitter.py:1150`) — the same
"declared-where" asymmetry, defining the stat-only floating block. It changes a different
quantity (every impact table) and deserves its own argument.

## A possible future refinement, not done here

A *strongly data-constrained* nuisance does eat some fraction of a degree of freedom. The
measured effective count via `1 − σ_post²/σ_prior²` is **35.7** on this card, giving
ndf 744.3 and p 38.9 % for ENVDATA. That is arguably more correct, but it is a research
choice and it makes the goodness-of-fit statistic depend on the postfit covariance — so the
statistic would stop being computable without a Hessian, and would change meaning between a
`--noHessian` run and a full one. The rule in this PR is the defensible, conventional one,
it is the one rabbit already applies to card nuisances, and it is what matches the linear
χ². Mentioning the alternative so it is on the record, not proposing it.

## Tests

New `tests/test_ndfsat.py`, in the repo's convention (`python tests/<file>`), added to the
CI unit-test matrix. **9 failed before → 9 passed after** (on `main` the checks fail on the
absent `Fitter.nfreeparms`; the behavioural before/after is the end-to-end driver table
above, which needs no new API). The checks:

- each declared prior buys back exactly one dof, for 0, 1 and all model nuisances priored;
- a constrained model parameter and a constrained card systematic cost the **same**, tested
  as a delta against the fully-free model;
- the logged breakdown sums to the number the ndf uses, and its systematics half is
  `nsystnoconstraint` by construction;
- the count agrees with what `_compute_lc` actually penalises — probed parameter by
  parameter by displacing each one and watching whether the penalty moves;
- the exact no-op for a prior-free model, with and without free card systematics;
- a guard that the no-op check is not vacuous (the delta is exactly the number of priors).

Full suite: **13 / 13 test files pass, 0 fail** (every `tests/test_*.py` on `main` plus the
new one; `test_restart`, `test_preconditioner` and `test_snapshot` run as
`python -m tests.<name>` because of their relative imports). Container `isort` / `black` /
`flake8` clean on the changed files and repo-wide; the repo's own pre-commit hook
(autoflake, isort, black, pylint 8.38/10) ran on the commit.


davidwalter2:
Reviewed at `aaae14a`, against `main`.

The statistics are right and the design choice is the good one. A Gaussian-constrained parameter adds a parameter and a pseudo-measurement and nets zero, so charging by *where a parameter is declared* rather than by whether it is constrained was indeed not a statistical property. Deriving it from `cw` — the same vector `_compute_lc` penalises — is what makes it impossible to drift, and that reasoning belongs in the docstring where you put it.

I verified the breakdown's claim rather than assuming it: `cw = tf.concat([param_cw_np, indata.constraintweights])`, so `nfreeparms_breakdown`'s second entry is `indata.nsystnoconstraint` exactly, by construction. The added log line is a real usability win — an ndf that changed silently would be hard to trace otherwise.

One finding.

---

### [P2] Frozen parameters are still charged a degree of freedom

`cw` records *constraints*. Frozen-ness lives in a separate `frozen_params_mask` Variable, with `floating_indices` derived from it. So a parameter that is both unconstrained and frozen has `cw == 0` and is counted by `nfreeparms` as free — but a frozen parameter is fixed, and costs no degree of freedom at all.

Effect: `ndfsat` too small, so `chi2.sf(chi2_val, ndfsat)` returns a p-value that is too pessimistic. Any run using `--freezeParameters` on an unconstrained parameter is affected, by exactly the number of such parameters.

This is not a regression — `param_model.nparams - nsystnoconstraint` ignored frozen parameters too. I am raising it because this is the PR that makes `nfreeparms` *the* answer to "what costs a degree of freedom", and its own docstring opens with:

> Number of fit parameters that cost a degree of freedom.

Frozen ones don't, so the abstraction stops one step short of the thing it names. Since `frozen_params_mask` is already maintained, it is:

```python
free = (self.cw.numpy() == 0.0) & ~self.frozen_params_mask.numpy()
```

with the same `&` applied in `nfreeparms_breakdown`.

`test_ndfsat.py` has no frozen case — `freezeParameters=[]` in the fixture is only the default. Worth one, alongside `test_constrained_model_params_cost_no_dof`: freezing a free parameter should raise `ndfsat` by one, and freezing an already-constrained one should change nothing.

One note for whoever merges: this changes the reported p-value for any analysis whose model declares priors. That is the point, and your no-op argument covers models without priors exactly, but it is a user-visible results change and worth calling out in the merge commit.

lucalavezzo:
You're right, and the docstring is the reason it's worth fixing here rather than filing it — "Number of fit parameters that cost a degree of freedom" is a claim, and a frozen parameter doesn't.

Fixed as the intersection rather than a subtraction, pulled into a helper so the count and its logged breakdown can't drift:

```python
def _free_parms_mask(self):
    return (self.cw.numpy() == 0.0) & ~self.frozen_params_mask.numpy()
```

The intersection is the point: freezing an **already constrained** parameter must change nothing, because it was costing nothing to begin with. A plain `nfreeparms - n_frozen` would double-count it and push `ndfsat` too large — the opposite error.

Three tests, and I'll be precise about which of them actually guards what, since only one fails before the change:

- `test_freezing_a_free_param_buys_back_a_dof` — **fails before, passes after.** The regression guard.
- `test_freezing_an_already_constrained_param_changes_nothing` — passes either way by construction. It's there to catch the subtraction mistake above, not this bug.
- `test_breakdown_also_excludes_frozen` — likewise, internal consistency between the count and the logged split.

Also took your merge note into the commit message, extended: this changes the reported saturated p-value for any analysis whose model declares priors, and now additionally for any run freezing unconstrained parameters.

---

Separately, the hodor bot raised a P2 on #170/#171 about `tests/test_native_minimizer.py` not being in the CI matrix and not being runnable under `python tests/<file>`. **That one is a false positive against this stack** — the file is already on `main` (`git ls-tree origin/main` finds it) and no diff in any of these four PRs touches it. The underlying gap looks real though, so it may be worth its own issue.

Issue #160, opened 2026-09-11
Move blinding out of fitter.py into its own module


davidwalter2:
Blinding is currently implemented inline in `fitter.py`, and with #159 adding an additive form it is getting subtle enough to want its own home — next to `BinByBinStat` and `Preconditioner`, which are already collaborators rather than inline Fitter code.

This is not primarily about file length. Blinding touches 54 lines of a 3094-line `fitter.py`, i.e. 1.7%. The argument is where those lines sit and what ties them together.

### Current shape

| lines | what |
|---|---|
| 45-46, 73, 166 | imports, one constructor flag |
| 215-257 | offset `tf.Variable` setup, inside `init_fit_parms` |
| **568-687** | **`init_blinding_values` + `set_blinding_offsets` — 123 cohesive lines** |
| 715-716, 797-798 | two-line touch points in `get_theta` / `get_poi` |

One self-contained block (seed derivation, the deterministic draw, the offset vectors, arm/disarm) plus four scattered touch points. The invariant that ties them together — **`self.x` is the internal, blinded frame; `get_poi()` / `get_theta()` are the physical values the model and the likelihood see** — is currently stated only in comments distributed across four methods.

The seam is clean. Outside `fitter.py` the only coupling is `--unblind` / `--blindingGroup` in `parsing.py` and a single `ifitter.set_blinding_offsets(...)` in `bin/rabbit_fit.py`. Nothing outside the Fitter touches `_blinding_offsets_*` or `_blinding_values_*`.

### Why it is worth doing

**Testability.** #159 had to build a toy `ParamModel`, a tensor writer and a full `Fitter` just to assert that an offset cancels, because the logic is only reachable through a constructed fit. A standalone object would be unit-testable directly — including the question of whether an offset actually *hides* the value for a given parameter scale, which today has no natural home (see the `blind_additive` scale discussion on #159).

**The invariant is getting harder.** #159 adds a change-of-frame compensation with an idempotence requirement, spread across `set_blinding_offsets`, `xdefaultassign` and `get_poi`. That wants one owner and one place to state the contract.

**Consistency.** `BinByBinStat` (984 lines) and `Preconditioner` (510) are both collaborators constructed once by the Fitter, owning their own TF Variables, with a flag mirroring an option that callers branch on. Blinding is the same shape and the odd one out.

### Proposed interface

`rabbit/blinding.py`, a `Blinding` class on the `BinByBinStat` pattern:

[...]

Issue #173, opened 2026-09-14
CI runs 15 of the ~152 unit tests: the matrix lists files by hand, so unlisted ones never run


davidwalter2:
## The problem

`.github/workflows/main.yml` has no `pytest` job. The `unit-tests` job is a hand-maintained matrix of filenames, each invoked as `python tests/${{ matrix.test }}`:

```yaml
matrix:
  test:
    - test_bbstat.py
    - test_external_term.py
    - test_global_asym_impacts.py
    - test_multi_systematic.py
    - test_sparse_fit.py
```

A test file that is not in that list runs nowhere. The default is **don't run**, and nothing reports the omission — a PR that adds tests goes green either way.

## How much is actually dead

Measured on the current `main` (13 test files, 5 registered):

| never run in CI | tests |
|---|--:|
| `test_native_minimizer.py` | 41 |
| `test_preconditioner.py` | 38 |
| `test_restart.py` | 23 |
| `test_preconditioner_spectral.py` | 13 |
| `test_snapshot.py` | 11 |
| `test_external_nll_constant.py` | 5 |
| `test_auxiliary.py` | 4 |
| `test_param_model_initial_params.py` | 2 |
| **total** | **137** |
| *registered and running* | *15* |

So roughly **90% of the unit tests never execute in CI**, including the preconditioner tests from #156 and the native-minimizer tests from #153.

I found this because I did the same thing: my #154 added `tests/test_sharded_fit.py` (~800 lines, 35 tests) with no matrix entry, which would have merged and never run once. Fixed there in b54bc65. This is not about any one PR — it is that the mechanism makes the mistake invisible.

## Suggested fix

Add a catch-all job alongside the existing matrix — keep the matrix for parallelism and per-file isolation, and add one job that collects everything, so a new file is covered from the moment it lands:

```yaml
  all-unit-tests:
    runs-on: [self-hosted, linux, x64]
    needs: setenv
    steps:
      - uses: actions/checkout@v5
      - name: setup python path
        run: |
          echo "PYTHONPATH=$GITHUB_WORKSPACE:$GITHUB_WORKSPACE/tests:$PYTHONPATH" >> $GITHUB_ENV
          echo "PATH=$GITHUB_WORKSPACE/bin:$PATH" >> $GITHUB_ENV
      - run: python -m pytest tests/ -q
```

### One thing to fix first

[...]

PR #174, opened 2026-09-14
blinding: never materialise the offset, and never print it

davidwalter2:
Two ways the blinding offset was reachable by accident, and one consequence of closing the second.

Context: rabbit's offset is `sha256(parameter_name)` with no salt, so it is computable by anyone with the code. Blinding here protects against *accidental* observation and unconscious bias, not against deliberate recovery. That makes the accident barrier the whole value, and both issues below lower it.

## 1. The weak-smearing warning handed over the offset

`_warn_if_additive_blinding_is_weak` logged, at `WARNING` level:

```
... <POI> (|offset|=XXXX, X.X sigma). Raise param_model.blind_additive_scale ...
```

`|offset|` to four significant figures, plus `offset/sigma`. Either field alone recovers the offset — the declared scale and the prefit sigma determine each other — and with the reported value that is the unblinded result. At `WARNING` level it lands in every log, CI transcript and pasted ticket. The warning existed to say blinding *may* be ineffective, and printing the number made it ineffective.

It also judged the **sample** rather than the configuration, so a perfectly sound setup whose draw happened to land near zero was reported, and an unsound one whose draw landed far out was not.

Now `_warn_if_additive_blinding_scale_is_weak` tests the **width of the draw** — `BLINDING_DRAW_STD * blind_additive_scale` — against the prefit sigma, names the affected parameters and the knob to turn, and prints **no numbers**. `test_the_weak_blinding_warning_does_not_leak_the_secret` asserts the message contains neither the offset nor the sigma in any of three float formats, while still naming the parameter and `blind_additive_scale`.

## 2. Frame compensation materialised the secret

`set_blinding_offsets` compensated `x` so the physical point held still. That is what writes the offset down: `x` becomes `x0default` mapped through the offsets, and `x0default` is **public** — it is the default the model declares. One subtraction recovers the offset from any prefit coordinate that gets written, logged, or inspected in a debugger.

The compensation is removed. `x` is left alone, so an armed fit opens at the model default plus the offset. `test_offset_is_not_recoverable_from_the_prefit_coordinate` states this as the attack: `x0default - x` must be identically zero while the offset is non-zero.

**What that costs, measured rather than assumed.** The minimiser reaches the same physical minimum from the shifted start, but only to its own convergence tolerance, and the two forms are not equal: additive agrees to `< 1e-6` relative, multiplicative to `~6e-5`, because `exp(N(0, 5))` starts exponentially further away. Both are pinned with those measurements in the test, so a regression that degrades them is visible.

The armed run also carries `0.5 * cw * offset**2` of extra prefit constraint penalty per blinded NOI — ~12.5 on average. It does not move the minimum (`test_constrained_minimum_is_not_moved`), and the test that used to assert the penalty was zero now measures it instead, so the cost is documented rather than hidden.

## 3. Where it cannot start, it now refuses

The real limitation, found while testing rather than reasoned about. With no compensation, a smearing wide enough to actually hide a POI can take it outside its evaluable range:

| `blind_additive_scale` | physical start | min yield | NLL |
|---|---|---|---|
| narrow (the default) | inside the model's range | positive | finite |
| wide enough to hide the POI | outside it | **negative** | **NaN** |

Negative yields, `log` of a negative number, NaN — and the minimiser has no finite point to descend from. It does not converge slowly; it cannot start. This is the 2026-09-09 alpha_s/SCETlib class of failure. The sting is that `blind_additive_scale` is the knob raised to make blinding effective, so the configurations that hide best are the likeliest to break.

`_check_blinded_start_is_evaluable` now runs once per armed fit and raises with an actionable message instead of producing NaN partway through. **A POI whose required smearing takes it out of its own domain cannot be blinded this way** — a genuine limitation, reported as one rather than worked around. The error message names the knobs and, like the warning, prints no numbers.

## 4. The saturated bin scales, caught by the above

They are blinded only because they share a POI block with the analysis model, and uncompensated they no longer opened at the 1.0 the warm start requires — 7 saturated tests failed. They are the saturated test's own machinery, not a result, so `SaturatedProjectModel` now declares `blind_exempt = True` and `CompositeParamModel` carries the names through the POI-block permutation. **This closes the substance of #162.**

## Verification

243 tests pass, `black`/`isort` clean, no new `flake8` findings (22 `E501` before and after, all pre-existing). Mutation-checked: reintroducing the compensation fails 3 tests; making the warning print the offset again fires the leak guard with the exact leaked string. `tests/test_blinding_start_point.py` is registered in the CI matrix and runs under `python tests/<file>` as the matrix invokes it.

Two tests whose subject no longer exists were removed rather than weakened (`test_squared_branch_uses_the_square_root_of_the_ratio`, `test_shift_is_additive_not_scaled`), and the "same loss at the same coordinate" tests were dropped in favour of the stronger "same minimum" test, which is what actually matters now.

Ref #160 (blinding extraction) — deliberately *not* included here; see the comment below.

https://claude.ai/code/session_01YVHjy5WbP2A7gdBHE3MQyN


davidwalter2:
### On #160 (moving blinding into its own module)

The plan was to do that here, and I've deliberately held it back. Explaining rather than just doing it differently, since it was asked for.

When we started, this was two contained fixes to `fitter.py`. It grew two more parts that weren't foreseen: the **fail-fast refusal** (removing the compensation makes some blinded fits un-startable, which had to be handled rather than shipped), and the **saturated exemption** (7 tests in the saturated path were built on the compensation). That is four behavioural changes in one commit already.

Layering a pure extraction on top would mean a diff in which everything has moved *and* behaviour has changed, and a reviewer cannot separate the two. Every test that pokes `f._blinding_offsets_poi_add` — and there are many across four files — would churn for reasons unrelated to any of the fixes above. `CLAUDE.md` also asks for one subject per PR, which this already stretches.

So my suggestion is #160 immediately after this, as a pure move with no behaviour change, which is both quick to review and mechanically checkable — the test suite should pass untouched except for attribute renames.

One thing worth recording: **this PR makes #160's flagged design problem go away on its own.** The issue noted that `set_blinding_offsets` mutating `self.x` was wrong for a collaborator, and proposed returning the shift for the Fitter to apply. With the compensation removed there is no shift and no mutation — `set_blinding_offsets` only assigns its own Variables — so the seam is already clean and the `Blinding` class can be extracted without that complication. The one method that genuinely needs the Fitter is `_check_blinded_start_is_evaluable`, since it evaluates the likelihood; that should stay on the Fitter and be called after arming, which keeps the collaborator free of Fitter internals exactly as the issue wanted.

Happy to do it either way — say so and I'll add it here as a second commit instead.

davidwalter2:
Both confirmed and fixed in 44453cb. The P2 in particular is a good catch — it is the failure mode this branch is least able to afford.

**[P2] Shadowed test.** Exactly as described: two definitions of `test_disarming_returns_to_the_starting_coordinate`, at lines 218 and 235. My rewrite landed *above* the original instead of replacing it, so Python kept the second and the rewrite was never collected. Worse than a dead test — the body that *did* run was the old one, still asserting on `assert_allclose` tolerance and still documenting *"this also pins that the reframing is invertible"*, i.e. the machinery this PR deletes. So the file read as though it covered the new contract while actually exercising the superseded one.

Kept the rewrite, dropped the original. The collected count is unchanged at 11, which is the point: the same number of tests, but now the correct one.

Root cause was mine and worth naming: I rewrote these tests with a script that located `def test_<name>` and replaced the block, and on this one the block boundary detection stopped short, so it inserted rather than replaced. A duplicate definition is invisible to pytest, to `black`, to `isort` and to `flake8` — nothing in the toolchain flags it.

**[P3] Dead scaffolding.** Right, and the reason is as you guessed: I copied the harness wholesale out of `test_blinding.py` when splitting the file out, including two model classes this file never instantiates. `PouOnlyAdditiveModel`'s docstring advertising the composite voting rule is the actively misleading part — a reader goes looking for a test that isn't there. Both deleted; `ParamModel` is still needed by `ToyModel`.

79 blinding and saturated tests pass, `black`/`isort`/`flake8` clean.

One follow-up this suggests, which I would rather not smuggle in here: a suite-wide guard against duplicate top-level test names would have caught the P2 immediately, and it sits naturally with #173, where the theme is already tests that silently do not run. Happy to add it there.

davidwalter2:
All three fixed in 1516f34. The P2 is a regression I introduced and it is a good catch — reachable, and CI was green with it in place.

**[P2] POI-less composite crash.** Confirmed exactly: composing two POI-less models raises `ValueError: need at least one array to concatenate`, and all four ABCD-family models have `npoi = 0`, so a multi-region fit built from `--paramModel` crashed in the constructor where it worked before. Cause is as you diagnosed — I un-nested the scale propagation out of `if additive_models:` when deleting the form vote, and lost the guarantee that at least one submodel had POIs. Guarded on `if self.npoi:`, which is also the honest condition: there is no POI block to scale.

Worth stating plainly: **CI passed on 1207cdd with this crash present**, because nothing in the matrix composes two POI-less models. Added the test that does, and mutation-checked it — restoring the unguarded form reproduces the `ValueError`.

**[P3] Stale form-voting paragraph.** Right, and it was worse than stale: it visually headed the `blind_exempt` name-propagation block, so it described deleted machinery *and* mislabelled the code beneath it. Deleted with its separator.

**[P3] `blind_exempt` documentation.** Also right, and this is the same mis-slice as the P2: I deleted the `blind_additive` contract entry and took the `blind_exempt` entry with it, because I had added the latter directly beneath the former. The attribute is live — `SaturatedProjectModel` sets it, the Fitter honours it, the composite propagates it, and *both* refusal messages in `_check_blinded_start_is_evaluable` tell the reader to declare it. Restored, including the `blind_exempt_params` name form and why it is names rather than indices. Also fixed the `blind_additive_scale` entry, which still explained itself by contrast with `exp(N(0, 5))`.

228 tests pass, `black`/`isort` clean, `E501` unchanged at 22.

A pattern worth naming, since it produced two of these three: the form removal was done with a script that locates a definition and replaces the block to the next top-level statement, and its boundary detection has now mis-fired six times in this PR — a shadowed test, a stale decorator, a deleted-along-with-its-neighbour `tf.Variable`, a swallowed test body, an over-broad rename, and these two doc entries. Every one was caught, but by tests or by this review rather than by the tool. I should have stopped using it for structural edits much earlier.

davidwalter2:
Fixed in 6082386. Good catch, and the annoying part is that the fix is a rule I had already written two commits earlier and failed to apply here.

`warn_if_blinding_is_weak` deliberately reads blinded-ness off the offsets, precisely so an `--unblind`ed POI is not reported as badly hidden — there is a test for it. `_warn_if_squared_storage_leaks_through_the_covariance` looked only at `allowNegativeParam` and `npoi`, so it fired on the parametrisation alone. The result is worst in the most ordinary run there is: unblinding a finished result keeps `do_blinding` on and keeps the default squared storage, so it warned that an expected sigma recovers the offset when every drawn value was zero, and told the reader to change parametrisation to fix a leak that could not occur.

Now guarded on `np.any(self._blinding_values_poi_add[:npoi])`, which is exactly the set `--unblind` and `blind_exempt` zero out, as you suggested — they are already populated at the call site. Test added for the all-unblinded case, and it fails with the guard removed.

231 tests pass, `black`/`isort` clean.

Two things still outstanding on this branch, both waiting on @davidwalter2 rather than on review:

- whether the sqrt(poi) covariance leak should be **fatal** rather than a warning — I argued for a warning because the squaring is rabbit's default and data fits blind by default, so refusing would make every existing analysis unrunnable, but it is a one-line change either way;
- the **seed rotation**, which was deliberately lifted out of this PR to keep it to one subject and still needs its own.

PR #178, opened 2026-09-16
blinding: move it out of fitter.py into its own module

### LINKED ISSUE #160: Move blinding out of fitter.py into its own module

davidwalter2:
Closes #160.

Blinding is now `rabbit/blinding.py`: a `Blinding` collaborator on the
`BinByBinStat` pattern, built once per parameter layout by `init_fit_parms`,
owning the deterministic draw, the offset Variables, arming, the hot-path
application and the two warnings about configurations that do not actually
hide anything. `enabled` mirrors the Fitter's `do_blinding`, and a disabled
instance owns no Variables and applies nothing — so `get_poi()` / `get_theta()`
lost their `if self.do_blinding` branches rather than gaining an indirection.

Mostly code motion. The four places it is not:

- **`match_regexp_params` moves to `common.py`.** Both the Fitter (freezing,
  systematic selection) and blinding resolve user expressions to parameter
  names, and blinding cannot import the Fitter it is a collaborator of.
- **The shard evaluators get a `BlindingView`**, which shares
  `apply_poi`/`apply_theta` with `Blinding` instead of having the offsets set
  directly on the shard object. The formula is written once for both, so a
  sharded fit can differ from a single-device one only in which tensors are
  pinned. `_BLINDING_OFFSET_ATTRS` now names attributes on the collaborator,
  and the threading guard additionally requires `BlindingView` to have a slot
  for each.
- **`warn_if_blinding_is_weak` becomes `Blinding.warn_if_weak`** — it reads the
  offsets and the model's declared scales and nothing else. Since the driver
  now calls it on the collaborator,
  `test_every_driver_called_fitter_method_is_classified_for_sharding` follows
  `fitter.blinding.*` too; without that, moving a method off the Fitter would
  drop it from the enumeration instead of forcing a decision about it, which is
  the failure mode #154 added that test for. Checked by deleting the
  `_SHARDED_SAFE` entry and watching it fail.
- **`set_blinding_offsets` stays on the Fitter.** Arming moves the physical
  point the fit opens at, and whether the model can be evaluated there is a
  question about the likelihood, which the Fitter owns; it arms the
  collaborator and then runs that check.

The issue also proposed having arming return the shift rather than mutating
`self.x`. #174 removed the frame compensation entirely, so there is no shift to
return — `arm()` does not touch `x` at all, and carries the reason.

### The point of it

`tests/test_blinding_module.py`: the draw, the group and exemption policy,
arming, application and the weak-smearing verdict, tested against stand-in
`indata` / `param_model` objects — no tensor file, no likelihood, no fit. 16
tests in 8 s, against ~100 s for the fit-based blinding files. Two branches
that had no coverage at all are now covered: an unblinded POI is not reported
as weakly blinded, and a variance left NaN by `--noHessian` means "no verdict"
rather than "weak". The fitter-level invariance tests are unchanged apart from
the renamed state they read.

### Checks

Full suite 293 passed (277 before, +16 new). Lint clean under CI's own
invocations (black, isort, the F-select flake8).

https://claude.ai/code/session_018LeeRn8UD8JY8pL3hyQzb2


lucalavezzo (commented):
Read the whole thing as an extraction and it holds up: I diffed each moved method against `f77f10e:rabbit/fitter.py` and `init_blinding_values` → `Blinding._init_values`, `_warn_if_squared_storage_leaks_through_the_covariance` → `_warn_if_squared_storage_leaks`, `warn_if_blinding_is_weak` → `warn_if_weak` and `match_regexp_params` → `common.py` are byte-identical modulo renames, docstrings and one black reflow. The non-motion parts all look like equivalences: the `do_blinding` guards → `blinding.enabled`, the `get_poi`/`get_theta` branches → `apply_poi`/`apply_theta` (identity when disabled), disarm via `zeros_like(values_*)` rather than `zeros(npoi)`, and `gview.blinding = self.blinding` in place of the conditional `setattr` loop.

Also checked, all clean: no stale references to the old names in `.py`, `.md` or CI config; no import cycle (`blinding` → `common` only); every consumer of the borrowed `get_poi`/`get_theta`/`get_x` has a `blinding` attribute, with `BlindingView` built per-shard inside the device loop and `_pin`'s `aux[1..]` order matching `_read_aux`; nothing reassigns `fitter.param_model` after `init_fit_parms`, so the captured model can't go stale, and `__deepcopy__` carries `blinding` and rebuilds the shards. 125 tests pass here, and `rabbit_fit.py` runs end-to-end with blinding armed plus `--computeSaturatedProjectionTests --saveHists`, which exercises the renamed driver call site and the saturated re-arm path.

Two comments inline. Only the first one I'd want resolved before merge, and only because the new test locks the behaviour in.

rabbit/blinding.py
```diff
@@ -0,0 +1,439 @@
...
+            unblind_names = [
+                p.decode() if isinstance(p, bytes) else str(p)
+                for p in unblind_parameters
+            ]
+            logger.info(f"Unblinding {len(unblind_names)} parameters: {unblind_names}")
+
+        # check if dataset is an integer (i.e. if it is real data or not) and use this to choose the random seed
+        is_dataobs_int = np.sum(
+            np.equal(self.indata.data_obs, np.floor(self.indata.data_obs))
+        )
```
lucalavezzo, line 303: `np.sum` here means the discriminator is true if **any** bin is integral, not all of them. So a non-integer card that happens to contain a single exactly-`0.0` bin (an empty bin, not rare in sparse regions) is seeded `_data` and draws the *same offsets as the real-data card*. An MC-closure run against known truth then gives `off = blinded − truth`, i.e. it unblinds the data fit.

I know this is verbatim from `fitter.py:647` and predates the PR, but it is re-landed as new lines here and `test_the_draw_separates_real_data_from_a_toy` now asserts the property — and that test only covers all-integer vs all-non-integer, so it passes without ever probing the mixed case. `np.all(...)` is a one-word fix while the lines are being touched; otherwise I'd at least not cement the current behaviour with a test that can't see the hole.

Separately, the quantity is read off `indata.data_obs` (the card) rather than the dataset actually being fitted, so an armed `--pseudoData` fit (`ifit == 0`, blinded per `rabbit_fit.py:1052`) shares the data fit's frame regardless — which is narrower than what that test's docstring claims.

rabbit/blinding.py
```diff
@@ -0,0 +1,439 @@
...
+        # reported sigma itself a channel for the offset, since
+        # sigma_true / sigma_reported WAS the offset.
+        self.offsets_poi_add = tf.Variable(
+            tf.zeros([self.param_model.npoi], dtype=self.indata.dtype),
+            trainable=False,
+            name="offset_poi_add",
+        )
+        self._init_values(unblind, blinding_group)
+
+    def arm(self, blind=True):
```
lucalavezzo, line 153: `arm()` is public but skips the domain check: every production path goes through `Fitter.set_blinding_offsets`, which arms and *then* runs `_check_blinded_start_is_evaluable` (the docstring below says as much). `fitter.blinding.arm(True)` is now reachable and arms a wide offset with no check, so a POI pushed outside its evaluable range gives NaN yields and an opaque minimiser failure instead of the named `RuntimeError`. The new test module already calls `arm(True)` directly.

Minor, but `_arm` — or folding the check in as a callback on `arm` — would keep the one reachable arming path the checked one.

rabbit/blinding.py
davidwalter2: Good catch, fixed in df8735f: `np.sum` → `np.all`, so a single exactly-0.0 bin no longer puts an MC/Asimov card in the real-data frame. I added `test_an_empty_bin_does_not_make_an_mc_card_look_like_data` for the mixed case (it fails against the old `np.sum`), and renamed the existing test to `..._separates_a_real_data_card_from_an_mc_card`.

On `--pseudoData`: agreed, the discriminator reads the card, not the dataset being fitted. I narrowed the test docstring and added a comment at the draw to say so. Fixing it properly means the frame would have to depend on the fitted dataset, which is drawn once at construction, before the dataset is chosen. That is a behaviour change rather than an extraction, so I would do it in a follow-up PR instead of this one.

rabbit/blinding.py
davidwalter2: Done in df8735f: renamed to `Blinding._arm`, with a note in its docstring to arm through `Fitter.set_blinding_offsets`. That leaves the checked path as the only public one. I preferred this to an arming callback because `__deepcopy__` carries `blinding` over, and a callback bound to the original Fitter would then check the wrong likelihood. The unit tests call `_arm` directly, which is fine because they have no likelihood to check.

PR #181, opened 2026-09-17
Give the saturated fit its own snapshot path

lucalavezzo:
## The bug

`save_hists()` builds the saturated fitter with `copy.deepcopy(fitter)`, so it inherits
`--snapshotFile`. Its own `minimize()` then writes periodic snapshots — onto the main
fit's file. From a real run's log:

```
218  Wrote parameter snapshot (converged)   <- the main fit, 3720 params
298  Wrote parameter snapshot (periodic)    <- the SATURATED fit, 3759 params
318  Wrote parameter snapshot (periodic)
351  Wrote parameter snapshot (periodic)
```

Two consequences, and the second is the one that cost real work:

1. **It destroys the main fit's `reason="converged"` snapshot** — the one artefact that
   makes a fit whose postfit died recoverable. Seen on a fit whose fitresults file was
   `meta`-only after a postfit crash: the converged snapshot was the only record of the
   minimum, and by the time anyone looked it had been replaced by the saturated fit's
   iterate.
2. **What replaces it has the composite layout** (the analysis parameters plus one bin
   scale per projection bin), so `--externalPostfit` cannot load it back into the main
   fit at all — while `Snapshotter._write` logs `resume with --externalPostfit <path>`
   on every write, inviting precisely that.

## The fix

Retarget `fitter_saturated.snapshot_file` to a sibling path immediately after the
deepcopy, **per mapping**, since each projection runs its own saturated fit.

A path rather than `None`, deliberately: the saturated fit can be the dominant cost of a
postfit — one measured at 87 % of postfit wall time, and one here ran 90+ minutes — so it
is worth keeping its own crash protection.

`snapshot.sibling_snapshot_path()` derives it: parent stem + a sanitised tag, same
directory and suffix, `None` passing through.

```
/ceph/x/snapshot_fitresults_MYFIT.hdf5 + "saturated_Project ch0 ptll"
  -> /ceph/x/snapshot_fitresults_MYFIT_saturated_Project_ch0_ptll.hdf5
```

The tag needs sanitising because the caller builds it from `mapping.key`, which carries
spaces.

## Tests

`tests/test_saturated_snapshot_path.py`:

- the helper: `None` passes through, the path differs from the parent, keeps the
  directory and suffix, the tag is sanitised, and distinct mappings get distinct paths;
- **`test_driver_retargets_the_saturated_fitters_snapshot`** — the regression test. It is
  a **source-level** check on `save_hists()`, for the same reason
  `test_saturated_blinding.py` uses one: the behaviour lives in a script's function, and
  the three lines at fault have no seam a unit test can reach without reimplementing the
  postfit. Verified to fail with the driver hunk reverted and pass with it.

`tests/test_saturated_snapshot_path.py` + `test_saturated_blinding.py` +
`test_saturated_composite.py`: **35 passed**. `isort` / `flake8` / `black` clean with the
CI's own flags.

## Relation to #180

Independent, though in the same block. #180 fixes a crash *earlier* in `save_hists`
(regularizers unarmed when the blinded-start check computes the loss); on this branch
alone a blinded, regularized saturated test still dies before it reaches any of this.
Both branch off `main` so each can be reviewed and merged on its own.


lucalavezzo:
Measured evidence from four real runs today, and it turns up a **worse manifestation than the PR description covers**.

Parameter counts on the snapshot files (3720 = the analysis layout, 3759 = analysis + 39 saturated bin scales for a 39-bin `ptll` projection):

| file | n | attrs | |
|---|---|---|---|
| `snapshot_CCKRYLOV.hdf5` | **3759** | `periodic`, iter 241 | clobbered |
| `snapshot_CCKRYLOVWARM.hdf5` | **3759** | **`reason='converged'`** | clobbered, **and indistinguishable from a real one** |
| `snapshot_fitresults_CCWALLCOLDR.hdf5` | **3759** | `periodic`, iter 145 | clobbered |
| `snapshot_fitresults_CCWALLWARM.hdf5` | 3720 | `converged` | clean — only because that run crashed *before* the saturated `minimize()` ran |

Three of four. The one clean file is clean by accident.

**The `reason='converged'` case is the part worth emphasising.** The saturated fitter writes its own `converged` snapshot when its minimization ends, so the file ends up claiming `converged` while holding the composite vector. Nothing in the metadata separates it from a genuine main-fit converged snapshot — only the parameter count does. So the natural defensive rule ("only resume from `reason=converged`") does not protect a user today, and `Snapshotter._write` is still advising `resume with --externalPostfit <path>` on every one of those writes.

Concretely this cost us the ability to re-seed from two converged minima that had been reached at 1720 and 31 iterations respectively. In one case the run's fitresults file was `meta`-only after an unrelated postfit crash (#180), so the converged snapshot had been the only record of the minimum — which is the scenario the snapshot exists for.

No change to the patch; posting because it strengthens the case and because the metadata-lie variant is not obvious from reading the code.

